PR #4, opened 2021-10-06
CM-68 : Google Sheets and .NET Core with C#

muhammedsalimp:
Code example for CM-68 : Google Sheets and .NET Core with C#


CodeMazeBlog:
A few things to change:

- we need tests to ensure the project works as intended
- the solution should be moved from "csharp-advanced-topics" section to "dotnet-client-libraries" section
- post, put and delete actions in ItemsController.cs should return a status code result (201 no content for POST and 204 no content for PUT and DELETE)

asif-anwar:
retest this please

PR #45, opened 2021-11-29
Action and fucn delegates by Ali

AlyHaneef:
Action and Func delegates with examples .


AlyHaneef:
Action and Func Delegate by Ali

PR #50, opened 2021-11-30
Delegates implemented in .NET Core by Martin

martinchege324:
The previous project was done in .NET 4.7.2 instead of .NET Core Version 5.


csharp-advanced-topics/Delegates/DelegateTests/DelegateTests.csproj
```diff
@@ -0,0 +1,20 @@
+<Project Sdk="Microsoft.NET.Sdk">
+
+  <PropertyGroup>
+    <TargetFramework>net5.0</TargetFramework>
+
+    <IsPackable>false</IsPackable>
+  </PropertyGroup>
```
CodeMazeBlog, line 7: .NET 5 is good, but .NET 6 is better. You don't need to change it now, but if we go to the next phase .NET 6 is mandatory for all the articles in order to keep them fresh for longer.

csharp-advanced-topics/Delegates/Delegates/Program.cs
```diff
@@ -0,0 +1,29 @@
...
+        public static int CalcProduct(int firstNum, int secondNum)
+        {
+            return firstNum * secondNum;
+        }
+        //define a void function that prints the product of two numbers
+        public static void PrintProduct(int result)
+        {
+            Console.WriteLine("The product is {0}", result);
+        }
+        static void Main(string[] args)
```
CodeMazeBlog, line 17: Newline needed.

csharp-advanced-topics/Delegates/DelegateTests/DelegateUnitTests.cs
```diff
@@ -0,0 +1,18 @@
+using Microsoft.VisualStudio.TestTools.UnitTesting;
+using Delegates;
+namespace DelegateTests
+{
+    [TestClass]
+    public class DelegateUnitTests
+    {
+        [TestMethod]
+        public void TestCalcProduct()
```
CodeMazeBlog, line 9: Check out the guidelines for proper test naming conventions.

csharp-advanced-topics/Delegates/DelegateTests/DelegateUnitTests.cs
```diff
@@ -0,0 +1,18 @@
...
+        {
+            int first = 20;
+            int second = 20;
+            int expected = 400;
+            int actual = Program.CalcProduct(first, second);
+            Assert.AreEqual(expected, actual);
+        }
+    }
+}
```
CodeMazeBlog, line 18: Some tests seem to be missing. Why haven't you created two more tests to test the Action and Func respectively? I know tests don't make too much sense in this article, but we need to see how you deal with them since every article should have at least some tests. Tests are important for us since that way we can make sure source code works properly by running dotnet test in Jenkins. We also ensure that the results won't change when we bump the framework versions in the future.

CodeMazeBlog (commented):
Okay overall, I would love to see one or two more tests and there are some minor issues that need to be addressed.

PR #75, opened 2021-12-10
"Action and Func delegates in C#" article code examples

salisayedi:
"Action and Func delegates in C#" article code examples added in csharp-advanced-topics section

### DIFF
--- a/csharp-advanced-topics/ActionAndFuncDelegatesInCsharp/Tests/ActionAndFuncExampleUnitTest.cs
+++ b/csharp-advanced-topics/ActionAndFuncDelegatesInCsharp/Tests/ActionAndFuncExampleUnitTest.cs
@@ -1,0 +1,40 @@
+using System.Collections.Generic;
+using ActionAndFuncDelegatesInCsharp;
+using Microsoft.VisualStudio.TestTools.UnitTesting;
+
+namespace Tests
+{
+    [TestClass]
+	public class ActionAndFuncExampleUnitTest
+	{
+		[TestMethod]
+		public void WhenShouldNotAskHowAreYou_ThenSayHi()
+		{
+			var jack = new Audience
+			{
+				Name = "Jack",
+				LikesChatting = true
+			};
+			var output = new List<string>();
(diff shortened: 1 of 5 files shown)


csharp-advanced-topics/ActionAndFuncDelegatesInCsharp/Tests/ActionAndFuncExampleUnitTest.cs
```diff
@@ -0,0 +1,40 @@
...
+using Microsoft.VisualStudio.TestTools.UnitTesting;
+
+namespace Tests
+{
+    [TestClass]
+	public class ActionAndFuncExampleUnitTest
+	{
+		[TestMethod]
+		public void WhenShouldNotAskHowAreYou_ThenSayHi()
+		{
```
CodeMazeBlog, line 12: Something strange happening with formatting here.

csharp-advanced-topics/ActionAndFuncDelegatesInCsharp/Tests/ActionAndFuncExampleUnitTest.cs
```diff
@@ -0,0 +1,40 @@
...
+			var jack = new Audience
+			{
+				Name = "Jack",
+				LikesChatting = true
+			};
+			var output = new List<string>();
+			ActionAndFuncExample.SayHi(jack, (message, audience) => output.Add(message),
+				_ => true);
+
+			Assert.IsTrue(output.Count == 2 && output[0] == "Hi!" && output[1] == "How are you?");
```
CodeMazeBlog, line 37: These are some nice tests. Proceeding forward, you can think about using tests as code snippets in the articles. This will help you write less code, since you don't need to write the main application, and help the readers understand what's expected from the output without having Console.WriteLine throughout the article and printing the results separately. Hope that makes sense. We can talk about it some more later. This is just an idea.

CodeMazeBlog (commented):
Good source code overall. Left some suggestions for the next stage.

PR #77, opened 2021-12-11
Add Action Func article

guiljs:
I added a new Solution for article about Action and Func in C#. This solution is in csharp-advanced-topics section.

It contains a test project, a library, a console application and a web application to demonstrate action and func features in C#.


csharp-advanced-topics/ActionFuncInCsharp/ActionFuncConsole/Program.cs
```diff
@@ -0,0 +1,52 @@
...
+//{
+//    Console.WriteLine($"DoSomethingWithAction found that {numberFound} is divided by 5");
+//}
+
+//int DoSomethingWithFunc(int numberFound)
+//{
+//    Console.WriteLine($"DoSomethingWithFunc : {numberFound}");
+//    return numberFound;
+//}
+
```
CodeMazeBlog, line 52: We don't leave the commented-out code in the source.

CodeMazeBlog (commented):
I've left most of the review in the document. It applies to the code. While the source code is just fine, it's too complicated for this topic.

PR #88, opened 2021-12-14
CM-110: “As” and “is” Keywords in C#

sdayube:
Issue: https://content.code-maze.com/projects/CM/issues/CM-110
Draft: https://drafts.code-maze.com/?p=61578&preview=true


csharp-basic-topics/AsAndIsOperatorsInCSharp/AsAndIsOperatorsInCSharpTests/AsAndIsOperatorsTests.cs
```diff
@@ -0,0 +1,99 @@
...
+        }
+
+        public string[] PrintedOutputToArray()
+        {
+            var printedString = output.ToString();
+            return printedString.Split(new[] { Environment.NewLine }, StringSplitOptions.RemoveEmptyEntries);
+        }
+
+        [Fact]
+        public void EqualityVsTypeComparison_ShouldPrintThatComparisonsAreValid()
```
CodeMazeBlog, line 30: A small change to test names. We have a convention we're trying to use, so check the guidelines, please. I think this is clear enough as it is, but let's try to use the same approach as everyone else for consistency.

CodeMazeBlog (commented):
Excellent source code. A minor change is needed in regards to test names. Other than that, everything else seems fine.

csharp-basic-topics/AsAndIsOperatorsInCSharp/AsAndIsOperatorsInCSharpTests/AsAndIsOperatorsTests.cs
sdayube: No problem, already changed it. Thanks for the feedback!

PR #129, opened 2021-12-23
CM-2 : Tuple in C#

king-Alex-d-great:


CodeMazeBlog:
You should use a separate branch for each article. You can see the commits from both the previous article and this one. My suggestion is to create a separate branch for the article, and then create a PR from that branch to our main. That way we get the clean commit history. You don't have to do it for this article since you've already created a PR and there are no conflicts, but please do it in the future.

csharp-intermediate-topics/TuplesInCsharp/Tests/TuplesTest.cs
```diff
@@ -0,0 +1,45 @@
...
+            var tupleOne = Program.MethodA(num);
+
+            Assert.IsType<Tuple<bool, string, int[]>>(tupleOne);
+
+            Assert.False(tupleOne.Item1);
+            Assert.Contains("is odd", tupleOne.Item2);
+        }
+
+        [Fact]
+        public void MethodB_WhenExecuted_ReturnsAString()
```
CodeMazeBlog, line 37: Please check our test naming convention in the guidelines and rename the tests.

csharp-intermediate-topics/TuplesInCsharp/TuplesInCsharp/Program.cs
```diff
@@ -0,0 +1,84 @@
...
+//CREATING TUPLES USING THE TUPLE.CREATE() METHOD
+var example1 = Tuple.Create<string>("code-maze"); //With the specific type indicated
+//var example1 = Tuple.Create("code-maze"); //Same code without type parameter
+
+var example2 = Tuple.Create("code-maze", 24);
+
+var example3 = Tuple.Create(true, 3, "code-maze", 23M, "blockchain");
+
+//Tuple with eight members
+var example4 = Tuple.Create(true, "between", "zero", "and", "one" , false, "bezao", 24);
```
CodeMazeBlog, line 22: Same thing as in the code snippets in the article. Variable names should be more descriptive. I know I'm guilty of doing this myself, but we should strive to do better.

CodeMazeBlog:
> You should use a separate branch for each article. You can see the commits from both the previous article and this one. My suggestion is to create a separate branch for the article, and then create a PR from that branch to our main. That way we get the clean commit history. You don't have to do it for this article since you've already created a PR and there are no conflicts, but please do it in the future.

EDIT: I've seen this happening a few times, even if people create new branches, they carry over old commits. If that happens, the best thing to do is to delete your fork, create a new one, and then create a branch from a fresh fork.

PR #131, opened 2021-12-23
added action and func delegates in C#

git-sarita:
Added action and func delegates test project as well. Please review it.


git-sarita:
closed

CodeMazeBlog:
Hey, @git-sarita there are a lot of issues with this PR.
First, it seems that you've worked in an existing folder of our repository DelegatesInCsharp. You need to create your own folder for your solution, for example, SaritaDelegatesInCsharp or something unique.
On top of that you've worked in .NET Framework 4.6, and we're expecting .NET 6 as the target framework.
There are also many binary files pushed which I don't know how you managed to do since there's a .gitignore file in the root folder of the repository.

Please fix these issues before we can proceed to the review.

PR #132, opened 2021-12-23
Action and Func Delegates Demo

ullahahsan1983:
Demo project for evaluation article about Action and Func delegates


CodeMazeBlog:
Tests are missing. Please add some tests in accordance with the evaluation requirements.

ullahahsan1983:
Test project is added. Please review

csharp-basic-topics/ActionAndFuncDelegates/ActionAndFuncDelegates/Messenger.cs
```diff
@@ -0,0 +1,15 @@
+﻿namespace CodeMaze
```
CodeMazeBlog, line 1: Strange namespace name. Have you changed it on purpose?

csharp-basic-topics/ActionAndFuncDelegates/ActionAndFuncDelegates/Program.cs
```diff
@@ -0,0 +1,122 @@
+using System;
+using System.Collections.Generic;
+using System.Diagnostics;
+using System.Linq;
+
+namespace CodeMaze
+{
+    public class Program
+    {
+        public static void Main(string[] args)
```
CodeMazeBlog, line 10: You can use file scoped namespaces, global using and remove the main method since it's .NET 6. It's up to you of course, but just to remind you that we support that.

csharp-basic-topics/ActionAndFuncDelegates/ActionAndFuncDelegatesTests/ActionAndFuncTests.cs
```diff
@@ -0,0 +1,89 @@
+using System;
+using System.IO;
+using CodeMaze;
+using Xunit;
+
+namespace ActionAndFuncDelegatesTests
```
CodeMazeBlog, line 6: The tests are okay. In the future you should maybe take into account the main source code since we want to make sure everything works. When you write new examples for the tests, we still don't know if the source code from the main project is working or not. 
The alternative would be writing tests in such a way that they can be used as article code snippets. That's something we suggest for those that are good with writing tests.
Your choice, but we definitely need to be able to make sure the entire source code is working as expected without having to run each one ourselves.

csharp-basic-topics/ActionAndFuncDelegates/ActionAndFuncDelegates/Messenger.cs
ullahahsan1983: Actually, I wanted to give it a name that is somewhat related to the ownership rather than the default generated one.

csharp-basic-topics/ActionAndFuncDelegates/ActionAndFuncDelegates/Program.cs
ullahahsan1983: Well, I was aware of this flexibility introduced in .NET6, but then thought that I might need to keep this project backward compatible. Now, I agree with you. Since we are setting target framework as .NET6, there is no point in backward compatibility. I will change this shortly.

csharp-basic-topics/ActionAndFuncDelegates/ActionAndFuncDelegatesTests/ActionAndFuncTests.cs
ullahahsan1983: Hmm, your alternative idea makes sense. I will need to revise my article anyway, so this is the right time I improve the project as you suggested.

csharp-basic-topics/ActionAndFuncDelegates/ActionAndFuncDelegates/Messenger.cs
ullahahsan1983: I gave it a second thought. You're right. I should not violate standard naming conventions for an arbitrary purpose. I will change that.

ullahahsan1983:
I pushed a new commit with the refinements you suggested. I am not so sure about the tests though. As this topic is more syntax-oriented rather than output-oriented, I found it difficult to test in a standard GWT way. However, I have added coverage of all the methods used in the source project anyway. I hope this meets the requirement. Please review

csharp-basic-topics/ActionAndFuncDelegates/ActionAndFuncDelegates/Messenger.cs
CodeMazeBlog: Yeah, it's better to stick to the default ones for the articles. Anyways, thanks for explaining it.

csharp-basic-topics/ActionAndFuncDelegates/ActionAndFuncDelegates/Program.cs
CodeMazeBlog: You have a point on backward compatibility, but it's easy to change it nowadays with dotnet format. It's really up to you, I just wanted to tell you that you can do it like this if you want.

csharp-basic-topics/ActionAndFuncDelegates/ActionAndFuncDelegatesTests/ActionAndFuncTests.cs
CodeMazeBlog: Thanks, everything looks fine. I've put the idea out there although it's still your evaluation article. Your code is properly organized and formatted and that's what matters the most. You should use whatever you're most comfortable with.

CodeMazeBlog:
Everything looks alright now. Thanks for clarifying it. Love the reflection 👍 Not sure I've seen anyone do that :)

PR #173, opened 2022-01-01
Add Action and Func Delegates - satyap48

LaSaeta:
Pull request for Action and Func Delegates in C#

### DIFF
--- a/csharp-advanced-topics/FuncAndActionDelegates/Tests/Tests.cs
+++ b/csharp-advanced-topics/FuncAndActionDelegates/Tests/Tests.cs
@@ -1,0 +1,49 @@
+﻿using Microsoft.VisualStudio.TestTools.UnitTesting;
+using FuncAndActionDelegates;
+using System;
+using System.Collections.Generic;
+using System.Linq;
+using System.Text;
+using System.Threading.Tasks;
+
+namespace FuncAndActionDelegates.Tests
+{
+    [TestClass()]
+    public class Tests
+    {
+        [TestMethod()]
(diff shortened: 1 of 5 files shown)


csharp-advanced-topics/FuncAndActionDelegates/Tests/Tests.cs
```diff
@@ -0,0 +1,49 @@
...
+using System.Linq;
+using System.Text;
+using System.Threading.Tasks;
+
+namespace FuncAndActionDelegates.Tests
+{
+    [TestClass()]
+    public class Tests
+    {
+        [TestMethod()]
```
CodeMazeBlog, line 14: You don't need () with TestClass and TestMethod

csharp-advanced-topics/FuncAndActionDelegates/Tests/Tests.cs
```diff
@@ -0,0 +1,49 @@
...
+        public void GivenFuncDelegate_WhenMethodReferenced_DelegateInvokationListNotEmpty()
+        {
+            Func<int, int, int> funcDel = Program.AddNumbers;
+            var invokationList = funcDel.GetInvocationList();
+            Assert.IsNotNull(invokationList);
+            Assert.AreEqual(1, invokationList.Length);
```
CodeMazeBlog, line 20: Try to introduce a newline between the Arrange/Act/Assert parts.

CodeMazeBlog (commented):
The code looks good overall. If we proceed further, try to create branches for each article you write and then create a PR from these branches. That way we can keep the commit history clean.

PR #201, opened 2022-01-10
luca.beretta ActionFuncDelegatesTests

berets76:
Added Tests project

### DIFF
--- a/csharp-advanced-topics/DelegatesInCsharp/ActionFuncTests/OtherTests.cs
+++ b/csharp-advanced-topics/DelegatesInCsharp/ActionFuncTests/OtherTests.cs
@@ -1,0 +1,60 @@
+﻿using Microsoft.VisualStudio.TestTools.UnitTesting;
+using System;
+using System.Linq;
+
+namespace ActionFunc.Tests
+{
+    [TestClass()]
+    public class OtherTests
+    {
+        [TestMethod()]
+        public void AnonymousFuncTest()
+        {
+            try
+            {
+                int a = 3;
+                int b = 9;
+                string expected = "27";
+                Func<int, int, string> myAnonimousFunc = (a, b) => { return $"{a * b}"; };
+                var result = myAnonimousFunc(a, b);
+                Assert.AreEqual(expected, result);
+            }
+            catch
+            { Assert.IsTrue(false); }
+        }
+
+        [TestMethod()]
+        public void AnonymousActionTest()
+        {
+            try
+            {
+                Action<string> myAnonymousAction = x => { Console.WriteLine($"Hello anonymous {x}"); };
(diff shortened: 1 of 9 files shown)


csharp-advanced-topics/DelegatesInCsharp/ActionFuncTests/OtherTests.cs
```diff
@@ -0,0 +1,60 @@
...
+    [TestClass()]
+    public class OtherTests
+    {
+        [TestMethod()]
+        public void AnonymousFuncTest()
+        {
+            try
+            {
+                int a = 3;
+                int b = 9;
```
CodeMazeBlog, line 16: We use var to define variables unless we must use the type explicitly or it's better to use it to make the examples clearer for the reader. In short, please use vars unless you need to use the type explicitely.

csharp-advanced-topics/DelegatesInCsharp/ActionFuncTests/ShapeTests.cs
```diff
@@ -0,0 +1,41 @@
+﻿using Microsoft.VisualStudio.TestTools.UnitTesting;
+using System;
+
+namespace ActionFunc.Tests
+{
+    [TestClass()]
+    public class ShapeTests
+    {
+        [TestMethod()]
+        public void PrintTest()
```
CodeMazeBlog, line 10: Same review as the other test class. vars, (), try-catch, use newlines between Arrange/Act/Assert.

csharp-advanced-topics/DelegatesInCsharp/ActionFuncTests/OtherTests.cs
```diff
@@ -0,0 +1,60 @@
...
+                Func<int, int, string> myAnonimousFunc = (a, b) => { return $"{a * b}"; };
+                var result = myAnonimousFunc(a, b);
+                Assert.AreEqual(expected, result);
+            }
+            catch
+            { Assert.IsTrue(false); }
+        }
+
+        [TestMethod()]
+        public void AnonymousActionTest()
```
CodeMazeBlog, line 27: Test method names are not in accordance with the naming conventions we've mentioned in the guidelines. Please change them.

csharp-advanced-topics/DelegatesInCsharp/DelegatesInCsharp.sln
```diff
@@ -1,11 +1,15 @@
 ﻿
```
CodeMazeBlog, line 1: Not sure what happened here, but you seem to have used one of our already existing projects. You need to create your own solution and not use the existing ones. Each solution in our repository is one article that's published. If you do it like this you can break the existing code.

PR #234, opened 2022-01-15
Added Func and Action Delegates

Sarvesh64:


Sarvesh64:
Do I need to add more commits?

PR #242, opened 2022-01-16
Action and Func delegates in C# - console041@gmail.com

hovermind:
Practical example projects with tests for Action and Func


hovermind:
!

CodeMazeBlog (commented):
The source code is good, but the example is too complicated and out of the scope for this particular article. I've talked more about it in the article review document.

PR #244, opened 2022-01-16
CM-22: String Split in C#

geo-bourno:


strings-csharp/StringSplit/Tests/StringSplitUnitTest.cs
```diff
@@ -0,0 +1,188 @@
...
+            {
+                Console.WriteLine($"High Score { i + 1 }: { highScoresArray[i] }");
+            }
+
+            //// Output:
+            ////High Score 1: 15562
+            ////High Score 2: 12988
+            ////High Score 3: 9876
+            ////High Score 4: 6584
+            ////High Score 5: 6292
```
CodeMazeBlog, line 29: You don't need this in the test. Apply to other examples.

strings-csharp/StringSplit/Tests/StringSplitUnitTest.cs
```diff
@@ -0,0 +1,188 @@
...
+
+            Assert.IsTrue(highScoresArray.Length == 5);
+            Assert.IsTrue(highScoresArray[0] == "15562");
+            Assert.IsTrue(highScoresArray[1] == "12988");
+            Assert.IsTrue(highScoresArray[2] == "9876");
+            Assert.IsTrue(highScoresArray[3] == "6584");
+            Assert.IsTrue(highScoresArray[4] == "6292");
```
CodeMazeBlog, line 36: This is excellent, but you need to use Assert.AreEqual instead of IsTrue. Apply to other examples.

PR #266, opened 2022-01-24
working action and func demo with tests

RenegadeMaster:
PR for test article


RenegadeMaster:
Removed local .gitignore as requested

PR #327, opened 2022-02-13
CM-117: Debugging Basics in C#

lasaetarubia:


CodeMazeBlog (commented):
The code is alright, there's not much to it, but let's create a new section called "visual-studio" instead of putting this into the "csharp-basic-topics". If the build fails, I'll change it after the merge, but let's try that out.

PR #391, opened 2022-03-24
CM-464: Quicksort Algorithm in C#

martinchege324:


CodeMazeBlog:
retest this please

csharp-algorithms/QuickSort/QuickSortTests/QuickSortUnitTests.cs
```diff
@@ -0,0 +1,46 @@
...
+        [TestMethod]
+        public void GivenUnsortedArray_CheckSortedNotNull()
+        {
+            var sortFunction = new QuickSortMethods();
+            var array = QuickSortMethods.CreateRandomArray(200);
+
+            var sortedArray = sortFunction.SortArray(array, 0, array.Length - 1, string.Empty);
+
+            Assert.IsNotNull(sortedArray);
+        }
```
CodeMazeBlog, line 44: GivenUnsortedArray_WhenArrayIsSorted_ThenResultingArrayIsNotNull
For example ^^
Remember, the "when" part is the action and it's mandatory, that's what you test.
And you sometimes forget the "then" in the "then" word. (the second and the third test)

csharp-algorithms/QuickSort/QuickSortTests/QuickSortUnitTests.cs
```diff
@@ -0,0 +1,46 @@
+using Microsoft.VisualStudio.TestTools.UnitTesting;
+using QuickSort;
+using System;
+using System.Diagnostics;
+
+namespace QuickSortTests
+{
+    [TestClass]
+    public class QuickSortUnitTests
```
CodeMazeBlog, line 9: This is a great oportunity to test for StackOverflowException you've mentioned in the article. You can even mention that they can check it out in the code if curious.

PR #402, opened 2022-03-28
CM-408: Sort List<T> by Property in the Object in .NET

lasaetarubia:


CodeMazeBlog (commented):
You need to do some refactoring as per the article review. 
Also, please move the solution to the other section that I've made. It's called "collections-lists". I plan to do some reorganizing.

PR #444, opened 2022-04-24
CM-504 : Introduction to Google Charts With Angular and ASP.NET Core Web API

lasaetarubia:


CodeMazeBlog:
Hey @lasaetarubia, try moving the .sln file to the third level. As it stands Jenkins cannot find it because it's on the fourth level.

So, instead of /aspnetcore-webapi/UsingGoogleCharts/API/API.sln it should be aspnetcore-webapi/UsingGoogleCharts/API.sln. I think we can leave the client project where it is, and we'll see how to include it in the Jenkins build in the future. 

Just make sure to change the relative paths to the projects in the .sln file once you move it.

CodeMazeBlog:
Hey @lasaetarubia, try moving the .sln file to the third level. As it stands Jenkins cannot find it because it's on the fourth level.

So, instead of /aspnetcore-webapi/UsingGoogleCharts/API/API.sln it should be aspnetcore-webapi/UsingGoogleCharts/API.sln. I think we can leave the client project where it is, and we'll see how to include it in the Jenkins build in the future. 

Just make sure to change the relative paths to the projects in the .sln file once you move it.

CodeMazeBlog:
Looks alright.

PR #455, opened 2022-05-01
CM-496: Check if a Number is a Power of 2 in C#

marco-luzzara:


CodeMazeBlog (commented):
Should be in "numbers-csharp" instead of csharp-basic-topics

PR #482, opened 2022-05-25
CM-507 : Vertical Slice Architecture in ASP.NET Core

lasaetarubia:


CodeMazeBlog:
Looks alright for the most part. Please apply the suggestions mentioned in the review. Also, can please you can add a few tests to check these endpoints? I think that might be enough for this article since I don't see any tests or a test project even.

PR #525, opened 2022-06-26
CM-503: Check if String Contains a String in String Array in C#

mahdiguliyev:


csharp-basic-topics/ContainsMethodInCsharp/ContainsMethodTests/Tests.cs
```diff
@@ -0,0 +1,80 @@
...
+
+        [TestMethod]
+        public void WhenGivenStringValueExistInText_ThenReturnTupleAsStarterPositionOfValueAndBooleanValue()
+        {
+            var result = find("Code-Maze", "If you want to read great articles, then let's check the Code-Maze.");
+            Assert.IsTrue(result.Item2);
+        }
+
+        static (int, bool) find(string givenString, string templateText)
+        {
```
CodeMazeBlog, line 47: The find() method shouldn't be in the test class itself. It should be in a separate class and called from the test class when needed.

CodeMazeBlog: This way you've duplicated the code, and haven't tested the method you're actually calling when running the application,

csharp-basic-topics/ContainsMethodInCsharp/ContainsMethod/Program.cs
```diff
@@ -0,0 +1,69 @@
+﻿#region LINQ Method Syntax
```
CodeMazeBlog, line 1: Please remove the regions.

csharp-basic-topics/ContainsMethodInCsharp/ContainsMethod/Program.cs
```diff
@@ -0,0 +1,69 @@
...
+
+var isExist = false;
+var countries = new string[] { "USA", "England", "Germany", "Sweeden", "Poland" };
```
CodeMazeBlog, line 4: Apply the naming suggestions from the review doc.

csharp-basic-topics/ContainsMethodInCsharp/ContainsMethod/Program.cs
```diff
@@ -0,0 +1,69 @@
...
+
+if (countries.Contains("Sweeden"))
+    isExist = true;
+else
+    isExist = false;
```
CodeMazeBlog, line 9: Ternary operator please.

csharp-basic-topics/ContainsMethodInCsharp/ContainsMethod/Program.cs
```diff
@@ -0,0 +1,69 @@
...
+var articles = new string[] { "article-1", "article-2", "article-3", "article-4", "article-5" };
+
+var isFounded = from article in articles
+                where article.Contains("article-1")
+                select article;
+
+foreach (var found in isFounded)
+{
+    Console.WriteLine("Existed article: " + found);
+}
```
CodeMazeBlog, line 26: Naming changes

csharp-basic-topics/ContainsMethodInCsharp/ContainsMethod/Program.cs
```diff
@@ -0,0 +1,69 @@
...
+
+#endregion
+
+#region Manually
+
+var result = find("Code-Maze", "If you want to read great articles, then let's check the Code-Maze.");
+Console.WriteLine("Result: " + result);
+
+static (int, bool) find(string givenString, string templateText)
+{
```
CodeMazeBlog, line 36: Needs to be in another class, and not static in order to be tested properly. Actually all you did in the article should be in a separate class to be properly tested in the tests. Please refactor this and make sure to use this approach in the future, because the tests you used here don't guarantee the application is running as expected.

CodeMazeBlog:
Also, the whole solution should be moved to "string-csharp" section of the repository.

PR #526, opened 2022-06-27
CM-432: Unsafe Code in C# (unsafe keyword)

lasaetarubia:


csharp-intermediate-topics/UnsafeCodeInCSharp/Tests/Tests.cs
```diff
@@ -0,0 +1,22 @@
+﻿using Microsoft.VisualStudio.TestTools.UnitTesting;
+
+namespace UnsafeCodeInCSharp.Tests
+{
+    [TestClass]
+    public class Tests
+    {
+        [TestMethod]
+        public void WhenPointerVariablePassedThenValueTriples()
```
PhilBroderick, line 9: Can this follow the normal test name pattern of Given_When_Then please Satya? The Given part is optional, so simply a matter of separating the When and Then parts with `_`.

csharp-intermediate-topics/UnsafeCodeInCSharp/UnsafeCodeInCSharp/Program.cs
```diff
@@ -0,0 +1,44 @@
...
+            int num = 10;
+            int* arg = &num;
+
+            Console.WriteLine("GetTriple Input: {0}", num);
+
+            GetTriple(arg);
+
+            Console.WriteLine("GetTriple Output: {0}", num);
+
+            string text = "Happy";
```
PhilBroderick, line 26: Opt to use `var` for variables where the data type is clear.

csharp-intermediate-topics/UnsafeCodeInCSharp/UnsafeCodeInCSharp/Program.cs
```diff
@@ -0,0 +1,44 @@
...
+                int a = 100;
+                int* ptr = &a;
+
+                Console.WriteLine("The value of variable: {0}", a);
+                Console.WriteLine("The value of variable using pointer: {0}", ptr->ToString());
+                Console.WriteLine("The address of variable : {0}", (int)ptr);
+            }
+
+            int num = 10;
+            int* arg = &num;
```
PhilBroderick, line 18: Could we use a clearer name for this variable instead of `arg`? Maybe `numPointer` or something along those lines?

PhilBroderick (changes requested):

PR #527, opened 2022-06-30
CM-309: HttpClient vs RestSharp - Which One to Use in .NET

mahrncic:
Creating a PR for implementation of HttpClient vs RestSharp comparison and benchmarks.


CodeMazeBlog:
Looks alright.

PR #536, opened 2022-07-04
mtwheed/mohamed.a.twheed@gmail.com Action and Func delegates in C#

MohamedTwheed:
Action and Func in C#

### DIFF
--- a/csharp-advanced-topics/Tests/Test.cs
+++ b/csharp-advanced-topics/Tests/Test.cs
@@ -12,17 +12,20 @@ namespace Tests
     {
 
         [TestMethod]
-        public void whenRateAndTimeIsSentForSingleMethod_DelegateExecutesTheReferencedMethod()
+        public void WhenRateAndTimeIsSentForSingleMethod_DelegateExecutesTheReferencedMethod()
         {
-            getInterestRate delegateTotalInterest = new getInterestRate(InterestService.getTotalInterestByHours); 
+            getInterestRate delegateTotalInterest = new getInterestRate(InterestService.getTotalInterestByHours);
+
             string result = delegateTotalInterest.Invoke(8, 10);
+
             Assert.AreEqual("Total Interest of $10 rate in 8 hours is 0.8", result);
         }
 
         [TestMethod]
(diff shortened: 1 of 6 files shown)


CodeMazeBlog:
A lot of code from the article is missing. There's no Action or Func code at all, and all I can see is an example for multicast delegate. Please add all the code that's missing, organize it into classes, and add tests for them.

Your solution is not properly organized either, the solution should be at the second level, and then you should have different folders for different projects. One for the main project and one for the test project (which you don't have).

Tests are mandatory and while we don't expect strict unit tests, we need tests that validate the code and show that it's working as expected. Please add some.

csharp-advanced-topics/Tests/Tests.csproj
```diff
@@ -0,0 +1,20 @@
+<Project Sdk="Microsoft.NET.Sdk">
+
+  <PropertyGroup>
+    <TargetFramework>net5.0</TargetFramework>
+
+    <IsPackable>false</IsPackable>
+  </PropertyGroup>
```
CodeMazeBlog, line 7: .NET 6 please

csharp-advanced-topics/Tests/Test.cs
```diff
@@ -0,0 +1,76 @@
...
+    public class Test
+    {
+
+        [TestMethod]
+        public void whenRateAndTimeIsSentForSingleMethod_DelegateExecutesTheReferencedMethod()
+        {
+            getInterestRate delegateTotalInterest = new getInterestRate(InterestService.getTotalInterestByHours); 
+            string result = delegateTotalInterest.Invoke(8, 10);
+            Assert.AreEqual("Total Interest of $10 rate in 8 hours is 0.8", result);
+        }
```
CodeMazeBlog, line 20: - Method names should be capitalized
- Please separate Arrange/Act/Assert with an empty line
- Use vars whenever possible, unless it's important to show which type it is or the compiler can't tell. (eg StringBuilder should be var)

CodeMazeBlog:
Thank you @MohamedTwheed, I've added some suggestions to improve the code. Please apply them when you can.

PR #539, opened 2022-07-06
Eval article action, func delegate in C#

JSGund:
I have pushed the updated changes for the framework


CodeMazeBlog:
Hey @JSGund , the build is failing. There might be several reasons for that but for now you can try:

- using one of the existing sections "folders" of our repository. Seems like you created a new one
- upgrading the project to the latest .NET version -  .NET 6

The build needs to pass in order for us to review the code.
Hope this helps.

PR #544, opened 2022-07-09
CM-445: How to Turn a C# object into a JSON String in .NET

dgarcia202:


CodeMazeBlog:
The code looks alright, but you've committed files to two separate sections. It seems like there's something from one of your previous topics in there, and Jenkins, smart as it is, compiled the other solution since it came first. It hasn't even checked this one at all.

Please remove the excess code and let's see if Jenkins compiles the solution.

dgarcia202:
Hey, I messed up my local branch and included some unwanted files in the PR. The issue is now solved.

CodeMazeBlog:
Looks alright now.

PR #561, opened 2022-07-23
CM-183: Blazor WebAssembly Exception Handling With Error Boundaries

PhilBroderick:


blazor-features/ErrorBoundaries/Tests/ErrorBoundariesTests.cs
```diff
@@ -0,0 +1,63 @@
...
+
+        // Act
+        var cut = RenderComponent<Index>();
+        var counter = cut.FindComponent<Counter>();
+        var counterButton = counter.Find("button");
+
+		for (var i = 0; i <= count; i++)
+		{
+            counterButton.Click();
+		}
```
CodeMazeBlog, line 33: If I had to guess this might be a mix of tabs and spaces which GitHub doesn't show properly. Try to use either one of them.

CodeMazeBlog:
Looks alright. One little issue to fix, and apply the changes from the review.

blazor-features/ErrorBoundaries/Tests/ErrorBoundariesTests.cs
PhilBroderick: Weird.. that wasn't showing up locally for me. Should be fixed now.

CodeMazeBlog: Happens more than you know.
I use the option in VS (CTRL + R , CTRL + W) that shows tabs/spaces visually:
image
It helps a lot.

CodeMazeBlog: I guess it's not that big of a deal, but we should try to keep to code consistent as much as we can.

PR #576, opened 2022-08-05
CM 574 - What is IndexOutOfRangeException in C#

anurag1302:


csharp-exceptions/IndexOutOfRangeExceptionSolution/IndexOutOfRangeExceptionProject/UtilityHelper.cs
```diff
@@ -0,0 +1,154 @@
...
+        }
+
+        public void UpperBoundMethod(bool forcefullyRaiseException)
+        {
+            if (forcefullyRaiseException)
+            {
+                for (int i = 0; i <= NumbersArray.Length; i++)
+                {
+                    Console.WriteLine(NumbersArray[i]);
+                }
```
CodeMazeBlog, line 33: Use var instead of int in for loops.

CodeMazeBlog (commented):
Small fix, looks alright otherwise.

PR #587, opened 2022-08-21
CM-571: Getting Started with Mapster in ASP.NET Core

mahrncic:


dotnet-client-libraries/HowToUseMapster/HowToUseMapster/Controllers/PeopleController.cs
```diff
@@ -0,0 +1,49 @@
...
+    [Route("api/people")]
+    [ApiController]
+    public class PeopleController: ControllerBase
+    {
+		[HttpGet("new-person")]
+		public IActionResult GetNewPerson()
+		{
+		    var person = MappingFunctions.MapPersonToNewDto();
+
+		    return Ok(person);
```
CodeMazeBlog, line 14: Indentation looks off, although it probably isn't. You've probably used the mix of tabs and spaces that make the code looks like this on GitHub. Use either tabs or spaces and we'll be alright.

mahrncic: Yes it was looking okay locally, however, I fixed it now!

PR #596, opened 2022-08-31
CM-573: What is NullReferenceException in C#?

martinchege324:
Initial draft and code

### DIFF
--- a/exception-handling/NullReferenceException/NullReferenceException/ExceptionMethods.cs
+++ b/exception-handling/NullReferenceException/NullReferenceException/ExceptionMethods.cs
@@ -1,0 +1,129 @@
+﻿namespace NullReference
+{
+    public class ExceptionMethods
+    {
+        public List<string> StudentList() 
+        {
+            List<string> studentList = null;
+
+            studentList.Add("John Doe");
+
+            return studentList;
+        }
+
+        public List<string> FixedStudentList()
+        {
+            List<string> studentList = new List<string>();
+
+            studentList.Add("John Doe");
+
+            return studentList;
+        }
+
+        public int[] RandomNumbers()
+        {
+            var rand = new Random();
+
+            int[] numbers = null;
+
+            for(int i = 0; i < numbers.Length; i++)
+                numbers[i] = rand.Next();
+
+            return numbers;
+        }
+
+        public int[] FixedRandomNumbers()
+        {
+            var rand = new Random();
+
+            var numbers = new int[50];
+
+            for (int i = 0; i < numbers.Length; i++)
+                numbers[i] = rand.Next();
+
+            return numbers;
+        }
+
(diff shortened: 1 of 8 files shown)


exception-handling/NullReferenceException/NullReferenceException/ExceptionMethods.cs
```diff
@@ -0,0 +1,129 @@
...
+            var people = personObj.AddRange(new String[] { "John", "Mary", "Jane", "Usher", "Andrew", "Grace", "Aston", "Sheila" });
+            var result = Array.Find(people, p => p.LastName == searchString);
+
+            if (result != null)
+            {
+                return result.ToString();
+            }
+            else
+            {
+                return $"{searchString}could not be found";
```
PhilBroderick, line 70: Need a space here between `{searchString}` and `could`.

exception-handling/NullReferenceException/NullReferenceException/Program.cs
```diff
@@ -0,0 +1,4 @@
+﻿Console.WriteLine("Main Program");
+
+
+
```
PhilBroderick, line 4: Some empty lines that can be removed.

PhilBroderick (changes requested):
A couple of small tweaks are noted here and in the feedback doc. But looking good 😄

exception-handling/NullReferenceException/NullReferenceException/ExceptionMethods.cs
```diff
@@ -0,0 +1,129 @@
...
+        }
+
+        public int[] RandomNumbers()
+        {
+            var rand = new Random();
+
+            int[] numbers = null;
+
+            for(int i = 0; i < numbers.Length; i++)
+                numbers[i] = rand.Next();
```
PhilBroderick, line 30: Can we standardize the use of brackets for loops? There is a couple of methods that use brackets in loops and most that don't. It doesn't bother me which we choose, but best to be consistent.

exception-handling/NullReferenceException/NullReferenceException/ExceptionMethods.cs
```diff
@@ -0,0 +1,139 @@
...
+            }
+
+            return numbers;
+        }
+
+        public string Teachers(string searchString)
+        {
+            var personObj = new Teacher();
+            
+            var people = personObj.AddRange(new String[] { "John", "Mary", "Jane", "Usher", "Andrew", "Grace", "Aston", "Sheila" });
```
PhilBroderick, line 55: One last nitpick. Could we use `string` instead of `String` in all occurrences? There is a mix in the code. I have applied the changes to the article already 😄

PhilBroderick (changes requested):

PhilBroderick (approved):

PR #628, opened 2022-09-24
CM-234: Make first letter of a string upper case in C#

martinchege324:
Initial draft

### DIFF
--- a/strings-csharp/FirstLetterToUpper/FirstLetterToUpperTests/FirstLetterToUpperUnitTests.cs
+++ b/strings-csharp/FirstLetterToUpper/FirstLetterToUpperTests/FirstLetterToUpperUnitTests.cs
@@ -1,0 +1,85 @@
+using FirstLetterToUpper;
+
+namespace FirstLetterToUpperTests
+{
+    [TestClass]
+    public class FirstLetterToUpperUnitTests
+    {
+        [TestMethod]
+        public void GivenALowerCaseString_WhenUsingSubStringTechnique_VerifyFirstCharIsUpper()
+        {
+            var upperCase = new FirstLetterToUpperMethods();
+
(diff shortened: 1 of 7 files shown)


PhilBroderick (changes requested):
Looks fine. See suggestions in feedback doc that will require some code changes.

PhilBroderick (approved):

strings-csharp/FirstLetterToUpper/FirstLetterToUpperTests/FirstLetterToUpperUnitTests.cs
```diff
@@ -5,11 +5,11 @@ namespace FirstLetterToUpperTests
     [TestClass]
     public class FirstLetterToUpperUnitTests
     {
+        FirstLetterToUpperMethods upperCase = new FirstLetterToUpperMethods();
```
PhilBroderick, line 8: Should be `var upperCase`. I've changed in the article 😃

PhilBroderick (changes requested):
Just one small change.

strings-csharp/FirstLetterToUpper/FirstLetterToUpperTests/FirstLetterToUpperUnitTests.cs
martinchege324: Hey @PhilBroderick, it didn't work because it's a global variable that's why I had to use FirstLetterToUpperMethods upperCase  instead of var

martinchege324: Please advise

PhilBroderick (approved):

PR #637, opened 2022-09-28
Added Span Count method for better performance

jtmueller:
The [SpanExtensions.Count](https://learn.microsoft.com/en-us/dotnet/api/microsoft.toolkit.highperformance.extensions.spanextensions.count?view=win-comm-toolkit-dotnet-6.1) method from the [CommunityToolkit.HighPerformance](https://www.nuget.org/packages/CommunityToolkit.HighPerformance/) nuget package is more than 3 times faster than the fastest method of counting characters in a string that this repo demonstrates.


CodeMazeBlog:
Thank you. We've also updated the article and replied to the comment.

PR #642, opened 2022-10-06
CM-509: Boosting Performance with Sealed Classes in .NET

mahrncic:


CodeMazeBlog:
Looks alright.

CodeMazeBlog:
Maybe we can move this to a new section called "dotnet-performance". I expect us to have more of these kinds of topics so we can as well create the section now. Try to create it if you can, and move the solution to the new section. Let me know if you have any problems.

mahrncic:
As agreed, I moved this solution to the "dotnet-performance" folder.

PR #687, opened 2022-10-30
CM 286: Different Ways to Use Select Tag Helper in ASP.NET Core

anurag1302:


aspnetcore-features/SelectTagHelperSolution/SelectTagHelper/Models/SelectViewModel.cs
```diff
@@ -0,0 +1,24 @@
...
+        public int? SelectedDepartment { get; set; }
+    }
+
+    public enum Department
+    {
+        IT,
+        HR,
+        Finance,
+        Admin
+    }
```
CodeMazeBlog, line 23: Extract to another file.

anurag1302: It is done.

PR #691, opened 2022-10-31
CM-178 Custom Event Arguments in Blazor WebAssembly

luiscoco:
CM-178 Custom Event Arguments in Blazor WebAssembly


luiscoco:
Some improvements

PR #695, opened 2022-11-04
CM-551: How to Use the “Using Static” Feature in C#

soundaranbu:


csharp-basic-topics/UsingStaticInCSharp/UsingStaticInCSharp.Tests/UsingStaticUnitTest.cs
```diff
@@ -0,0 +1,55 @@
+namespace UsingStaticInCSharp.Tests;
+public class UsingStaticUnitTest
+{
+    readonly StringWriter output;
+
+    public UsingStaticUnitTest()
+    {
+        output = new StringWriter();
+        Console.SetOut(output);
+    }
```
CodeMazeBlog, line 10: We cannot have Console.SetOut as a valid testing method. I understand what you were going for, but there are better ways to test this. Please use Verify method that tests if the method has been called. We have an article on that that can help you: https://code-maze.com/using-moq-to-determine-if-method-is-called/

soundaranbu: Thanks for the feedback. I've now implemented the same.

PR #699, opened 2022-11-05
CM-66/using system linq dynamic core library

aineaTheSecond:


csharp-linq/SystemDynamicLinqCoreLibrary/SystemDynamicLinqCoreLibrary/Program.cs
```diff
@@ -0,0 +1,46 @@
...
+
+var itEmployeesList = employees.Where("c => c.Department = \"IT\"").ToList();
+
+var relevantData = employees.Select("new {Name, Department}").ToDynamicList();
+
+var sortedData = employees.OrderBy("Name").ToList();
+
+var multipleSortedData = employees.OrderBy("Department desc, Name").ToList();
+
+
```
CodeMazeBlog, line 21: All the different examples should be in methods and separated form the Program class. This will help test them in the test project.

csharp-linq/SystemDynamicLinqCoreLibrary/Tests/SystemDynamicLinqCoreLibraryUnitTests.cs
```diff
@@ -0,0 +1,83 @@
...
+{
+    public class SystemDynamicLinqCoreLibraryUnitTests
+    {
+        private Employee _employee;
+
+        [SetUp]
+        public void SetUp()
+        {
+            _employee = new Employee();
+        }
```
CodeMazeBlog, line 15: While these tests are fine, you're not testing the application. We have a lot of code in the program.cs and we don't know if it works or not, without running the application. You need to test this code instead of recreating the Employee class in the test project and testing it separately.

Please create classes with methods that demonstrate different examples, and then test these instead of creating new ones.

aineaTheSecond: I have updated the application by creating an additional TestData class. The class has methods demonstrating the examples, all tested. 
Please check the latest commits.

csharp-linq/SystemDynamicLinqCoreLibrary/SystemDynamicLinqCoreLibrary/Program.cs
aineaTheSecond: I have moved all this logic to a separate class.

CodeMazeBlog: Excellent!

csharp-linq/SystemDynamicLinqCoreLibrary/Tests/SystemDynamicLinqCoreLibraryUnitTests.cs
CodeMazeBlog: Much better, thanks!

PR #707, opened 2022-11-09
CM-596: Producer/Consumer Applications With .NET Channels

PhilBroderick:
I've added a new folder for this project - let me know if there is a better place for this.


CodeMazeBlog:
A good idea about the new folder. You can also upgrade the projects to .NET 7 now that it's out. Shouldn't change anything in the article. 
The code looks alright as expected.

PR #713, opened 2022-11-14
CM-590: How to Implement LinkedList in C#

FilipiBraga:
LinkedList and CustomLinkedList implementation


PhilBroderick (changes requested):
Looks fine. There'll be some changes required that are noted in the feedback document.

PhilBroderick (approved):

collections-lists/LinkedList/LinkedList/CustomImplementation/CustomLinkedList.cs
```diff
@@ -0,0 +1,241 @@
...
+{
+    public class CustomLinkedList<T> : IEnumerable<Node<T>>
+    {
+        public Node<T>? First { get; set; }
+        public int Count { get; set; }
+
+        public void AddFirst(Node<T> nodeToAdd)
+        {
+            if (First is null)
+                First = nodeToAdd;
```
PhilBroderick, line 19: Could we add brackets for this `if` statement? The article has already been changed.

PhilBroderick (changes requested):
Just one small change to make to the code. Also if we can move this solution to the `collections-csharp` folder instead as the `collections-lists` folder is strictly related to problems around the `List` class.

collections-lists/LinkedList/LinkedList/CustomImplementation/CustomLinkedList.cs
FilipiBraga: I have moved to the correct folder and added brackets to a couple of `if` statements that were missing.

PhilBroderick (approved):

PR #727, opened 2022-11-23
CM-609:Value vs Reference Types in C#

alfredkakuli:


mahrncic (changes requested):

csharp-basic-topics/ValueVsReferenceTypes/ValueVsReferenceTypes/Program.cs
```diff
@@ -0,0 +1,24 @@
+﻿using ValueVsReferenceTypes;
+class Program
+{
+
+    public static void Main()
+    {
+        WorkingDaysCalculator workingDaysCalculator = new();
+        Car car = new();
+
+        CarModel carModel1 = new()
```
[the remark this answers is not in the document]
alfredkakuli: hello, thanks.
The idea of car2 and Model1 was to explicitly differentiate between the first and second instances before and after mutation. About the extra CarModel class, I was thinking of a typical real-world scenario where a car has a bunch of properties with persistence.
Once again thank you, I will update as per your guidelines.

csharp-basic-topics/ValueVsReferenceTypes/ValueVsReferenceTypes/Car.cs
```diff
@@ -0,0 +1,13 @@
+﻿namespace ValueVsReferenceTypes
+{
+    public class Car
+    {
+        public string ChangeCarModel( string newModel)
```
mahrncic, line 5: Remove the blank space before the `string` keyword in the method parameters space.

mahrncic (changes requested):
Good job on the first batch of changes, I added a couple of more :)

Also, please make sure to upgrade the solution/project to .NET 7

csharp-basic-topics/ValueVsReferenceTypes/ValueVsReferenceTypes/Car.cs
```diff
@@ -0,0 +1,15 @@
+﻿namespace ValueVsReferenceTypes
+{
+    public class Car
+    {
+        public Car(string modelName)=> ModelName = modelName;
```
mahrncic, line 5: Why do we initialize the value of the `ModelName` setter in the constructor?

csharp-basic-topics/ValueVsReferenceTypes/ValueVsReferenceTypes/WorkingDaysCalculator.cs
```diff
@@ -0,0 +1,19 @@
+﻿namespace ValueVsReferenceTypes
+{
+    public class WorkingDaysCalculator
+    {
+        public WorkingDaysCalculator() => DaysOfTheWeek = 7;
```
mahrncic, line 5: Again, why do we initialize the value of the `DaysOfTheWeek` setter in the constructor, and not on the setter directly?

mahrncic (changes requested):

PR #734, opened 2022-11-30
CM-563: Schedule Jobs with Quartz.NET

PhilBroderick:


dotnet-client-libraries/Quartz.NET/Tests/BackgroundJobTests.cs
```diff
@@ -0,0 +1,59 @@
...
+        _jobExecutionContextMock.SetupGet(j => j.MergedJobDataMap).Returns(jobDataMap);
+        using var stringWriter = new StringWriter();
+        Console.SetOut(stringWriter);
+
+        // Act
+        await _backgroundJob.Execute(_jobExecutionContextMock.Object);
+
+        // Assert
+        Assert.Equal(expectedOutput, stringWriter.ToString());
+    }
```
CodeMazeBlog, line 37: I'm not a fan of Console.SetOut is a valid way to test anything. I see what you did, but whenever possible we should try to avoid testing the console output. Let's try to create examples that actually do something else. Console.Writeline is good to show things in the article, but a very weak way of testing the application in tests. Can lead to potential problems and is hard to understand as well.

We can leave the code as is for this article, but try to figure out how you would do it without the SetOut method, and please ask the authors to change the code if they do it with SetOut. In 99.99% of cases, there's a better way to test the application.

PhilBroderick: I completely agree with you. I would usually avoid testing Console output directly, opting instead to abstract the output behind an interface and mock it in tests.

I made the decision to use Console for purposes of a simple example and implemented the code/tests like this to keep things simpler for the reader, so we didn't have to introduce output abstractions in the article.

PR #740, opened 2022-12-02
CM-588: ToString method in C#

aineaTheSecond:


CodeMazeBlog:
Looks alright except the change I've mentioned in the review document.

PR #747, opened 2022-12-07
Cm 604 extreme linq performance boostin net7

tecmaivan:


CodeMazeBlog:
The code looks alright, I guess there is no need for tests. There are some empty lines I would like to see removed, and we should probably move this project to the "dotnet-performance" section instead of "csharp-linq" even though it's about LINQ.

PR #754, opened 2022-12-14
CM-630: How to Add a BearerToken to an HttpClient Request

FilipiBraga:
Two applications implementation to demonstrate how to Add a bearer token to an HttpClient request


authorization-dotnet/AddBearerTokenToHttpClientRequest/ConsumerApp.Application.cs/ProgramApplication.cs
```diff
@@ -0,0 +1,141 @@
...
+            _userRepository = userRepository;
+            _loginApiRepository = loginApiRepository;
+        }
+
+        public async Task SignIn()
+        {
+            Console.WriteLine("----- Let's sign in");
+
+            Console.Write("Type the Email: ");
+            string email = Console.ReadLine() ?? "johnDoe@codemaze.com";
```
PhilBroderick, line 26: These should be `var`.

authorization-dotnet/AddBearerTokenToHttpClientRequest/ConsumerApp.Infrastructure/Repository/LoginApiRepository.cs
```diff
@@ -0,0 +1,44 @@
...
+        private readonly HttpClient _httpClient;
+
+        public LoginApiRepository(HttpClient httpClient)
+        {
+            _httpClient = httpClient;
+        }
+
+        public async Task<AccessToken> AuthenticateAsync(string email, string password)
+        {
+            var uri = new Uri("https://localhost:5001/api/auth/login");
```
PhilBroderick, line 22: Could be defined as the `BaseAddress` on the HttpClient - see feedback doc.

PhilBroderick (changes requested):
I've left some more suggestions in the feedback doc that will need to be addressed here.

Instead of having two different solutions, we should have one ideally. This is much easier for the reader to understand what is happening when all the projects are in the same solution.

PhilBroderick (approved):

PR #757, opened 2022-12-16
CM-544: How to Get an Enum Member as a String in C#

sdayube:


CodeMazeBlog:
Let's move the code to the "strings-csharp" section in the repository. It doesn't belong in the basic topics even though the topic is pretty basic.

PR #766, opened 2022-12-20
CM-592: Understanding PLINQ in .NET

marco-luzzara:


csharp-linq/UnderstandingPLINQ/Tests/Tests.csproj
```diff
@@ -0,0 +1,21 @@
+<Project Sdk="Microsoft.NET.Sdk">
+
+  <PropertyGroup>
+    <TargetFramework>net6.0</TargetFramework>
```
PhilBroderick, line 4: We should target .NET 7 now.

PhilBroderick (changes requested):
Just need to update to .NET 7 and good to go.

PhilBroderick (approved):

PR #780, opened 2022-12-29
CM-572: How to Randomize a List<T> in C#

martinchege324:
Initial draft and code

### DIFF
--- a/collections-lists/RandomizeGenericListsInCSharp/RandomizeGenericListsInCSharp/RandomizeGenericListsInCSharp.csproj
+++ b/collections-lists/RandomizeGenericListsInCSharp/RandomizeGenericListsInCSharp/RandomizeGenericListsInCSharp.csproj
@@ -2,7 +2,7 @@
 
   <PropertyGroup>
     <OutputType>Exe</OutputType>
-    <TargetFramework>net6.0</TargetFramework>
+    <TargetFramework>net7.0</TargetFramework>
     <ImplicitUsings>enable</ImplicitUsings>
     <Nullable>enable</Nullable>
   </PropertyGroup>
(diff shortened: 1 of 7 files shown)


collections-lists/RandomizeGenericListsInCSharp/RandomizeGenericListsInCSharp/RandomizeGenericListsInCSharp.csproj
```diff
@@ -0,0 +1,14 @@
+<Project Sdk="Microsoft.NET.Sdk">
+
+  <PropertyGroup>
+    <OutputType>Exe</OutputType>
+    <TargetFramework>net6.0</TargetFramework>
```
PhilBroderick, line 5: We should be targeting .NET 7 now.

collections-lists/RandomizeGenericListsInCSharp/RandomizeGenericListsTests/RandomizeGenericListsUnitTests.cs
```diff
@@ -0,0 +1,63 @@
...
+
+    [TestMethod]
+    public void GivenAnOrderedList_WhenNotEmpty_VerifyShuffleGuid()
+    {
+        var shuffledList = _randomizeListObj.GenerateRandomOrderByGuid(_orderedList);
+
+        var firstVal = shuffledList.First();
+        var lastVal = shuffledList.Last();
+        
+        Assert.IsFalse(firstVal == 0);
```
PhilBroderick, line 25: Instead of `Assert.IsTrue()`, we should be using `Assert.AreEqual()` when comparing values and not explicitly checking the value of a `bool`.

PhilBroderick (changes requested):
Looks alright. A couple of suggestions here and a few more in the feedback doc.

PhilBroderick (approved):

PR #805, opened 2023-01-10
CM-623: Inter-Service Communication in Microservice Architecture in ASP.NET Core

adhikari-chayan:
Inter-Service Communication in Microservice Architecture in ASP.NET Core


CodeMazeBlog:
Do we need all the services in the solution? Shall we keep it to the ones related to this article only?
Also, please update all the projects to .NET 7 (and maybe create a PR to upgrade the versions in the last article as well).

adhikari-chayan:
> Do we need all the services in the solution? Shall we keep it to the ones related to this article only? Also, please update all the projects to .NET 7 (and maybe create a PR to upgrade the versions in the last article as well).

Done. Will create separate PR for the last article

CodeMazeBlog:
Ok, merging this. Please do remember to fix the .NET versions for the previous article.

PR #811, opened 2023-01-18
CM-315 : How to Mock EF Core DbContext

muhammedsalimp:
CM-315 : How to Mock EF Core DbContext


dotnet-testing/EFCoreMocking/Tests/Tests/Usings.cs
```diff
@@ -0,0 +1 @@
+global using Xunit;
```
mahrncic, line 1: Let's use this '`Usings.cs` class. You have a couple of same `using` statements in different test classes. For example, `using EFCoreMocking.Data`, `using Moq`, etc.

Let's move all of these using statements that are used in different test class to this one, and save some lines :)

mahrncic (changes requested):
Good job, just a couple of small changes :)

mahrncic (approved):

mahrncic (approved):

PR #812, opened 2023-01-20
CM-423: Using multipart form-data in HttpClient

dgarcia202:


mahrncic (approved):

mahrncic (changes requested):
All works good now. However, we don't even need to have the `UsingMultipartFormDataInHttpClient` project. Since we only have the `Program` class there which is not used anywhere.

So just remove the `UsingMultipartFormDataInHttpClient` project and we're good.

mahrncic (approved):

PR #822, opened 2023-01-22
Added Function and Action guides

EmreKabali:
Added Console App about Action and Function with Test program.


EmreKabali:
Added Console App about Action and Function with Test program.

PR #834, opened 2023-01-25
CM-289:How to get current authenticated user with claims

htset:


mahrncic (changes requested):
Looks good. However, tests are missing, and we should always have tests for our articles/code when possible :)

mahrncic (approved):

PR #849, opened 2023-02-01
CM-330: How to Compare Two Json Objects Using C#

aineaTheSecond:


json-csharp/ComparingTwoJsonObjects/ComparingTwoJsonObjects/Car.cs
```diff
@@ -0,0 +1,27 @@
...
+            return false;
+        }
+
+        var car = (Car)obj;
+
+        return Name == car.Name && Model == car.Model && Make == car.Make && Price == car.Price;
+    }
+}
+
+public class Price
```
karimkod, line 23: put this class in a separate file

karimkod (approved):

PR #855, opened 2023-02-01
CM-403: Determine Whether Two Date Ranges Overlap

FilipiBraga:


PhilBroderick (approved):
Looks fine. May require some changes depending on discussion from feedback doc.

CodeMazeBlog:
Sometimes builds break because of the .gitignore file.
Please remove it from the solution folder. We have a master .gitignore at the repository level.
If you need something added to it, we can do that.

PR #861, opened 2023-02-04
CM-37: Access Specifiers In C#

lasaetarubia:


csharp-basic-topics/AccessSpecifiersInCsharp/AccessSpecifiersInCsharp/BankAccount.cs
```diff
@@ -0,0 +1,25 @@
+﻿namespace AccessSpecifiersInCsharp
+{
+    public class BankAccount
+    {
+        private int balance;
```
PhilBroderick, line 5: We should follow the VS naming conventions, so this should be `_balance`.

PhilBroderick (changes requested):

PhilBroderick (approved):
Looks fine.

csharp-basic-topics/AccessSpecifiersInCsharp/AccessSpecifiersInCsharp/Program.cs
```diff
@@ -0,0 +1,34 @@
...
+            var account = new BankAccount();
+
+            account.Deposit(100);
+            var balance = account.GetBalance();
+            
+            Console.WriteLine(balance);
+            
+            var rectangle = new Rectangle(10, 5);
+            
+            int area = rectangle.GetArea();
```
PhilBroderick, line 23: Can we change this to `var`?

PhilBroderick (changes requested):
We need to rename the project/solution file. It can be tricky but there are ways to do it.

PhilBroderick (approved):

PR #881, opened 2023-02-16
CM-238: Return multiple values to a method caller in C#

FilipiBraga:
Implementation of the Return Multiple Values to a Method Caller in C#

### DIFF
--- a/csharp-basic-topics/ReturnMultipleValues/ReturnMultipleValues/MultipleValues.cs
+++ b/csharp-basic-topics/ReturnMultipleValues/ReturnMultipleValues/MultipleValues.cs
@@ -2,7 +2,7 @@
 {
     public sealed class MultipleValues
     {
-        public string StringValue { get; set; } = string.Empty;
+        public required string StringValue { get; set; }
         public bool BoolValue { get; set; }
         public int IntValue { get; set; }
     }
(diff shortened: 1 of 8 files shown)


csharp-basic-topics/ReturnMultipleValues/ReturnMultipleValues/MultipleValues.cs
```diff
@@ -0,0 +1,9 @@
+﻿namespace ReturnMultipleValues
+{
+    public sealed class MultipleValues
+    {
+        public string StringValue { get; set; } = string.Empty;
```
PhilBroderick, line 5: MIght be able to use the `required` keyword to avoid setting a default value.

csharp-basic-topics/ReturnMultipleValues/ReturnMultipleValues/MultipleValuesReturner.cs
```diff
@@ -0,0 +1,56 @@
...
+        public static Tuple<string, bool, int> GetValuesUsingTuple()
+        {
+            var stringValue = "Some String Value";
+            var boolValue = false;
+            var intValue = int.MaxValue;
+
+            return Tuple.Create(stringValue, boolValue, intValue);
+        }
+
+        public static (string srtValue, bool boolValue, int intValue) GetValuesUsingTupleLiterals()
```
PhilBroderick, line 14: `str` is usually the abbreviation we use for `string`.

PhilBroderick (changes requested):
Looks alright. Couple of changes required.

PhilBroderick (approved):

PR #882, opened 2023-02-17
CM-441: TimeSpan in C#

martinchege324:
First draft

### DIFF
--- a/dotnet-datetime/TimeSpanInCSharp/TimeSpanInCSharpTests/TimeSpanUnitTests.cs
+++ b/dotnet-datetime/TimeSpanInCSharp/TimeSpanInCSharpTests/TimeSpanUnitTests.cs
@@ -1,0 +1,201 @@
+using System.Globalization;
+using System.Text.RegularExpressions;
+using TimeSpanInCSharp;
+
+namespace TimeSpanInCSharpTests
+{
+    [TestClass]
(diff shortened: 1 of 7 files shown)


PhilBroderick (approved):
Looks fine.

dotnet-datetime/TimeSpanInCSharp/TimeSpanInCSharpTests/TimeSpanUnitTests.cs
```diff
@@ -0,0 +1,201 @@
...
+
+            Assert.AreEqual(subtractMethod, subtractOperator);
+            Assert.AreEqual(expected, subtractOperator);
+            Assert.AreEqual(expected, subtractMethod);
+            Assert.IsInstanceOfType(subtractMethod, typeof(TimeSpan));
+            Assert.IsInstanceOfType(subtractOperator, typeof(TimeSpan));
+        }
+
+        [TestMethod]
+        public void GivenATimeSpan_WhenConvertedToString_VerifyAccurateResult()
```
PhilBroderick, line 140: We should verify the format of the returned value, so something like:

```
var expected = "04:00:00";

Assert.AreEqual(expected, actual);
```

This will need added to the article also.

PhilBroderick (changes requested):

PR #888, opened 2023-02-19
CM-323: How to read connection strings in .NET

dgarcia202:


aspnetcore-features/HowToReadConnectionStringsInDotNet/HowToReadConnectionStringsInDotNet/Controllers/ProductsController.cs
```diff
@@ -0,0 +1,27 @@
+﻿using Microsoft.AspNetCore.Mvc;
+using Microsoft.Extensions.Configuration;
+
+namespace HowToReadConnectionStringsInDotNet.Controllers
```
mahrncic, line 4: Can we use file-scoped namespaces? It looks better in my mind since we can just remove one level of indentation from our files :)

mahrncic (changes requested):
Looks good, just one simple change :)

mahrncic (approved):

PR #895, opened 2023-02-20
Cm 624 Blazor Web Assembly Custom Events

alvaromon:


blazor-features/BlazorWebAssemblyCustomEvents/BlazorWebAssemblyCustomEvents/CustomEvents/EventHandlers.cs
```diff
@@ -0,0 +1,17 @@
...
+{
+    [EventHandler("ondoubleclick", typeof(DoubleClickEventArgs), enableStopPropagation: true, enablePreventDefault: true)]
+    public static class EventHandlers
+    {
+        // purposefully empty
+    }
+
+    public class DoubleClickEventArgs : EventArgs
+    {
+        public DateTime? EventTimestamp { get; set; }
```
mahrncic, line 14: Is there a reason why the `EventTimestamp` property is nullable? I think that we always need to have the `EventTimestamp`

blazor-features/BlazorWebAssemblyCustomEvents/BlazorWebAssemblyCustomEvents/CustomEvents/EventHandlers.cs
```diff
@@ -0,0 +1,17 @@
...
+        public bool isDoubleClick { get; set; } = false;
```
mahrncic, line 15: The default value of the `bool` type properties is `false`, so there is no need to set that explicitly :)

mahrncic (changes requested):

blazor-features/BlazorWebAssemblyCustomEvents/BlazorWebAssemblyCustomEvents/CustomEvents/EventHandlers.cs
```diff
@@ -0,0 +1,17 @@
...
+    [EventHandler("ondoubleclick", typeof(DoubleClickEventArgs), enableStopPropagation: true, enablePreventDefault: true)]
+    public static class EventHandlers
+    {
+        // purposefully empty
+    }
+
+    public class DoubleClickEventArgs : EventArgs
+    {
+        public DateTime EventTimestamp { get; set; }
+        public bool isDoubleClick { get; set; }
```
mahrncic, line 15: Sorry I missed that on the first review, but properties inside classes should always follow the Pascal-Case naming convention. Based on that, the `isDoubleClick` property should start with the uppercase -> `IsDoubleClick`

mahrncic (changes requested):
One minor change left :)

mahrncic (approved):

mahrncic (approved):

PR #899, opened 2023-02-22
CM-233 : What does "yield break;" do in C#?

muhammedsalimp:


csharp-basic-topics/YieldBreakInCSharp/Tests/UtilityTests.cs
```diff
@@ -0,0 +1,27 @@
+using YieldBreakInCSharp;
+
+namespace Tests
+{
```
mahrncic, line 4: Please use file-scoped namespaces. Just to avoid having one indentation lever more :)

csharp-basic-topics/YieldBreakInCSharp/YieldBreakInCSharp/Utility.cs
```diff
@@ -0,0 +1,41 @@
+﻿namespace YieldBreakInCSharp
+{
```
mahrncic, line 2: Please use file-scoped namespaces

mahrncic (changes requested):

mahrncic (approved):

PR #908, opened 2023-02-25
Cm 25  partial classes in c#

sternet:
Created example Console project with unit tests for [CM-25] (https://content.code-maze.com/browse/CM-25)
Partial Classes In C# task


sternet:
Aborting because of some trouble with gitignore

PR #911, opened 2023-02-26
CM-486: Using Trie Class for Efficient Text Pattern Searching in C#

lasaetarubia:


PhilBroderick (changes requested):
Looks fine. Might need to re-run benchmarks however.

PhilBroderick (approved):

PR #933, opened 2023-03-02
CM-453: How to Serialize a list to JSON in C#

cjaynjoku:


json-csharp/HowToSerializeAListToJsonInCSharp/HowToSerializeAListToJsonInCSharp/SerializeMethodsBenchmark.cs
```diff
@@ -0,0 +1,33 @@
...
+        public void UseTheJsonSerializerClass()
+            => _serializeWithNewtonsoftJson.UseTheJsonSerializerClass();
+
+        [Benchmark]
+        public void UseTheSerializeMethod()
+            => _serializeWithSystemTextJson.UseTheSerializeMethod();
+
+        [Benchmark]
+        public void UseTheSerializeToUtf8BytesMethod()
+            => _serializeWithSystemTextJson.UseTheSerializeToUtf8BytesMethod();
```
CodeMazeBlog, line 31: As mentioned in the article we need to lose the "UseThe..." prefix. On top of that, we could maybe add the type of the library instead, in the benchmark method names.
For example:
UseTheSerializeObjectMethod => NewtonSoftJsonSerializeObject
UseTheSerializeMethod => SystemTextJsonSerialize

My feeling is that this would make them much more readable both in the code and once the table is generated. You should be able to distinguish between the libraries and methods immediately.

json-csharp/HowToSerializeAListToJsonInCSharp/Tests/NewtonsoftJsonMethodsTests.cs
```diff
@@ -0,0 +1,73 @@
...
+    ""country"": ""England"",
+    ""numberOfPlayers"": ""25""
+  },
+  {
+    ""name"": ""Sunderland"",
+    ""yearFounded"": 1879,
+    ""country"": ""England"",
+    ""numberOfPlayers"": ""30""
+  }
+]";
```
CodeMazeBlog, line 55: Would "[raw string literals](https://learn.microsoft.com/en-us/dotnet/csharp/language-reference/builtin-types/reference-types#string-literals)" from C# 11 help here so you don't have to indent the text all the way to the left?

Play around a bit and let me know. This might help you professionally as well, and I'm just curious :)

CodeMazeBlog (changes requested):
Looks alright, but please rename these few methods for better readability and output.

CodeMazeBlog:
Looks alright now.

PR #941, opened 2023-03-07
Action and Func Delegates

shahidpro1:


shahidpro1:
I've added the Test project with the unit tests.

PR #977, opened 2023-03-21
CM-359: Datetime Calendars in C#

sdayube:


dotnet-datetime/DateTimeCalendar/DateTimeCalendar/CalendarTests.cs
```diff
@@ -0,0 +1,67 @@
+﻿using System;
```
karimkod, line 1: This is an unused import. Let's remove it

dotnet-datetime/DateTimeCalendar/DateTimeCalendar/CalendarTests.cs
```diff
@@ -0,0 +1,67 @@
...
+using System.Globalization;
+
+namespace DateTimeCalendar
+{
+    public class CalendarTests
```
karimkod, line 6: Let's rename this to be CalenderUnitTest as well as the filename CalendarUnitTest.cs

karimkod (changes requested):
Apart from the small change requested, the folder structure is not in line with our convention. We should have a project even if it's empty. called DateTimeCalendar.csproj and another project named Tests.csproj or DateTimeCalendarTests.csproj that contains the test
Please update the namespaces in all the files accordingly

PR #1006, opened 2023-03-31
bug fix performance test

DanHarltey:
Hi, I read your blog post https://code-maze.com/csharp-how-to-delete-elements-from-an-array/ and thought that the DeleteWithArraySegment results did not look correct. There is a bug in the performance test that means when it is getting tested all the matching elements have been removed from the array. So it does nothing,

Fixed this, added Buffer.BlockCopy as it should be quicker. Add DeleteWithList as I can’t have Linq winning :-)


CodeMazeBlog:
Thank you for the valuable feedback @DanHarltey! Sometimes mistakes like this slip through.

We'll talk to the author and let him investigate the issue first. After that, we'll decide if we're going to use this PR (we assume you don't mind since you've created a PR in the first place) or create another one to fix the problem.

PR #1014, opened 2023-04-03
Kohalloran1215: Request Body Read

ohallaatme:
Code to support article for how to read request body in ASP.NET Core API Controller


aspnetcore-webapi/RequestBodyRead/RequestBodyRead/Program.cs
```diff
@@ -0,0 +1,47 @@
+﻿
+using System.Diagnostics.Contracts;
+
+namespace RequestBodyRead;
```
PhilBroderick, line 4: This looks a bit odd having the namespace between the using statements. We should move this to the bottom of the block.

aspnetcore-webapi/RequestBodyRead/RequestBodyRead/Program.cs
```diff
@@ -0,0 +1,47 @@
...
+
+using Microsoft.AspNetCore.HttpOverrides;
+using Microsoft.AspNetCore.Mvc;
+using Microsoft.Extensions.DependencyInjection.Extensions;
```
PhilBroderick, line 8: We have some unused usings here. Visual Studio is usually pretty good at telling you what is being used and what isn't: 
image

aspnetcore-webapi/RequestBodyRead/RequestBodyRead/Program.cs
```diff
@@ -0,0 +1,47 @@
+﻿
```
PhilBroderick, line 1: Also some empty space here.

aspnetcore-webapi/RequestBodyRead/RequestBodyRead/Program.cs
```diff
@@ -0,0 +1,47 @@
...
+        {
+            app.UseSwagger();
+            app.UseSwaggerUI();
+        }
+
+        app.UseHttpsRedirection();
+
+        app.UseAuthorization();
+
+
```
PhilBroderick, line 41: Likewise with empty space.

aspnetcore-webapi/RequestBodyRead/RequestBodyRead/Shared/DataTransferObjects/UserForCreationDto.cs
```diff
@@ -0,0 +1,13 @@
+﻿using System;
+namespace RequestBodyRead.Shared.DataTransferObjects
+{
+	public record UserForCreationAndUpdateDto
+	{
+        public string? FirstName { get; set; }
+
+        public string? LastName { get; set; }
+
+        public string? Email { get; set; }
```
PhilBroderick, line 10: Instead of defining these as auto-properties, is it possible to make use of the record constructor? So something like:
```csharp
public record UserForCreationAndUpdateDto(string? FirstName = null, string? LastName = null, string? Email = null);
```

aspnetcore-webapi/RequestBodyRead/Tests/UserControllerTests.cs
```diff
@@ -0,0 +1,139 @@
...
+            var url = "/api/users/";
+            var response = await client.GetAsync(url);
+            var content = await response.Content.ReadAsStringAsync();
+            List<User>? users = JsonConvert.DeserializeObject<List<User>>(content);
+
+            var guids = users.Select(x => x.Id).ToList();
+            var guidsExpected = DataBaseMockup.Users.Select(x => x.Id).ToList();
+
+            foreach (var guid in guidsExpected)
+                Assert.True(guids.Contains(guid));
```
PhilBroderick, line 39: Instead of `Assert.True`, we should use `Assert.Contains`. Moreover, we can skip the `foreach` loop by writing our assertion like so:

```csharp
Assert.Contains(guidsExpected, guids.Contains);
```

aspnetcore-webapi/RequestBodyRead/Tests/UserControllerTests.cs
```diff
@@ -0,0 +1,139 @@
...
+            var resEmails = result.Select(x => x.Email).ToList();
+            var resFirstNames = result.Select(x => x.FirstName).ToList();
+            var resLastNames = result.Select(x => x.LastName).ToList();
+
+            Assert.True(resEmails.Contains("james.ryan@somealias.com") &&
+                resEmails.Contains("sophie.brown@somealias.com"));
+            Assert.True(resFirstNames.Contains("James") &&
+                resFirstNames.Contains("Sophie"));
+            Assert.True(resLastNames.Contains("Ryan") &&
+                resLastNames.Contains("Brown"));
```
PhilBroderick, line 104: Let's add a space between each `Assert.True` to make it clearer what we are asserting.

aspnetcore-webapi/RequestBodyRead/RequestBodyRead/Controllers/UsersController.cs
```diff
@@ -0,0 +1,78 @@
...
+    {
+        List<UserForCreationAndUpdateDto>? users;
+
+        if (buffer)
+        {
+            users = await _bufferService.ReadUserModelAsync(this.Request.BodyReader);
+        }
+        else
+        {
+            string body = await new StreamReader(this.Request.Body).ReadToEndAsync();
```
PhilBroderick, line 70: Should be `var`

PhilBroderick (changes requested):
Some work for this one is required.

When opening the solution in Visual Studio, and even looking at the code in GitHub, it looks like we've mixed up tabs and spaces. We don't have a preference either way, but we need to select one and stick to it throughout the solution, so this will need addressed.

PR #1031, opened 2023-04-12
CM 479 - How to Get the IP Address of the Remote Host in ASP.NET Core Web API

amuslic:


aspnetcore-webapi/RemoteHostIpAddress/RemoteHostIpAddress/Controllers/WeatherForecastController.cs
```diff
@@ -0,0 +1,80 @@
...
+            return remoteIpAddress;
+        }
+
+        public IPAddress? GetRemoteHostIpAddressUsingXRealIp()
+        {
+            IPAddress? remoteIpAddress = null;
+
+            if (HttpContext.Request.Headers.TryGetValue("X-Real-IP", out var xRealIp) &&
+                IPAddress.TryParse(xRealIp, out var realIp) &&
+                (realIp.AddressFamily == AddressFamily.InterNetwork || realIp.AddressFamily == AddressFamily.InterNetworkV6))
```
karimkod, line 72: This conditin is a bit too complex, I suggest we use variables before the if and use those variables in the iff instead, e.g : 
`var xRealIpExists = HttpContext.Request.Headers.TryGetValue("X-Real-IP", out var xRealIp);
IPAddress.TryParse(xRealIp, out var realIp); 
var isValidIP =  (realIp.AddressFamily == AddressFamily.InterNetwork || realIp.AddressFamily == AddressFamily.InterNetworkV6);`

This way the code is more readable.

aspnetcore-webapi/RemoteHostIpAddress/RemoteHostIpAddress/Controllers/WeatherForecastController.cs
```diff
@@ -0,0 +1,80 @@
...
+
+        public IPAddress? GetRemoteHostIpAddressUsingXForwardedFor()
+        {
+            IPAddress? remoteIpAddress = null;
+            var forwardedFor = HttpContext.Request.Headers["X-Forwarded-For"].FirstOrDefault();
+
+            if (!string.IsNullOrEmpty(forwardedFor))
+            {
+                var ips = forwardedFor.Split(new char[] { ',' }, StringSplitOptions.RemoveEmptyEntries)
+                                      .Select(s => s.Trim());
```
karimkod, line 50: Can't we use this version of the Split https://learn.microsoft.com/en-us/dotnet/api/system.string.split?view=net-7.0#system-string-split(system-char-system-stringsplitoptions)
Since it take one characater, without needing the table : `new char[] { ',' }`

aspnetcore-webapi/RemoteHostIpAddress/RemoteHostIpAddressTests/RemoteHostIpAddressTests.cs
```diff
@@ -0,0 +1,122 @@
...
+
+            // Act
+            var result = _controller.GetRemoteHostIpAddressUsingXForwardedFor();
+
+            // Assert
+            Assert.Equal(IPAddress.Parse("192.168.1.1"), result);
+        }
+
+        [Fact]
+        public void WhenGetRemoteHostIpAddressUsingXForwardedFor_WithInvalidIp_ThenReturnsNull()
```
karimkod, line 60: The name of the method doesn't conform to our guidelines, the underscore should only be used to separate between Given, When and Then. Here it's used to separate the With. This method should be named:  
`WhenGetRemoteHostIpAddressUsingXForwardedForWithInvalidIp_ThenReturnsNull` instead.

aspnetcore-webapi/RemoteHostIpAddress/RemoteHostIpAddressTests/RemoteHostIpAddressTests.cs
```diff
@@ -0,0 +1,122 @@
...
+
+            // Act
+            var result = _controller.GetRemoteHostIpAddressUsingXForwardedFor();
+
+            // Assert
+            Assert.Null(result);
+        }
+
+        [Fact]
+        public void WhenGetRemoteHostIpAddressUsingXRealIp_WithInvalidIp_ThenReturnsValidIpAddress()
```
karimkod, line 81: Same as the underscore remark.

aspnetcore-webapi/RemoteHostIpAddress/RemoteHostIpAddressTests/RemoteHostIpAddressTests.cs
```diff
@@ -0,0 +1,122 @@
...
+
+            // Act
+            var result = _controller.GetRemoteHostIpAddressUsingXRealIp();
+
+            // Assert
+            Assert.Equal(IPAddress.Parse("192.168.1.1"), result);
+        }
+
+        [Fact]
+        public void WhenGetRemoteHostIpAddressUsingXRealIp_WithInvalidIp_ThenReturnsNull()
```
karimkod, line 102: Same, the usage of underscore.

karimkod (changes requested):

aspnetcore-webapi/RemoteHostIpAddress/RemoteHostIpAddressTests/RemoteHostIpAddressTests.cs
amuslic: sure, thanks for the reminder :)

aspnetcore-webapi/RemoteHostIpAddress/RemoteHostIpAddress/Controllers/WeatherForecastController.cs
amuslic: sure, thanks :)

aspnetcore-webapi/RemoteHostIpAddress/RemoteHostIpAddress/Controllers/WeatherForecastController.cs
amuslic: sounds good, thanks :)

aspnetcore-webapi/RemoteHostIpAddress/RemoteHostIpAddress/RemoteHostService.cs
```diff
@@ -0,0 +1,62 @@
...
+
+            if (!string.IsNullOrEmpty(forwardedFor))
+            {
+                var ips = forwardedFor.Split(',', StringSplitOptions.RemoveEmptyEntries)
+                                      .Select(s => s.Trim());
+
+                foreach (var ip in ips)
+                {
+                    if (IPAddress.TryParse(ip, out var address) &&
+                        (address.AddressFamily == AddressFamily.InterNetwork || address.AddressFamily == AddressFamily.InterNetworkV6))
```
karimkod, line 26: this is taking a bit of space and can't be viewed in the article without scrolling, can we switch to using 
`address.AddressFamily is AddressFamily.InterNetwork or AddressFamily.InterNetworkV6` and if it still doesn't fit into the screen without screen maybe we can split into two lines.

aspnetcore-webapi/RemoteHostIpAddress/RemoteHostIpAddress/RemoteHostService.cs
```diff
@@ -0,0 +1,62 @@
...
+            var xRealIpExists = httpContext.Request.Headers.TryGetValue("X-Real-IP", out var xRealIp);
+
+            if(xRealIpExists)
+            {
+                if (!IPAddress.TryParse(xRealIp, out IPAddress? address))
+                {
+                    return remoteIpAddress;
+                }
+
+                var isValidIP = (address.AddressFamily == AddressFamily.InterNetwork || address.AddressFamily == AddressFamily.InterNetworkV6);
```
karimkod, line 49: Same as above 
`address.AddressFamily is AddressFamily.InterNetwork or AddressFamily.InterNetworkV6`

aspnetcore-webapi/RemoteHostIpAddress/RemoteHostIpAddress/RemoteHostService.cs
```diff
@@ -0,0 +1,62 @@
+﻿using System.Net.Sockets;
+using System.Net;
+
+namespace RemoteHostIpAddress
+{
+    public class RemoteHostService : IRemoteHostService
+    {
+        public IPAddress GetRemoteHostIpAddressUsingRemoteIpAddress(HttpContext httpContext)
```
karimkod, line 8: If this can return null, we should consider changing the return type to `IPAddress ?`

karimkod (changes requested):

aspnetcore-webapi/RemoteHostIpAddress/RemoteHostIpAddressTests/IRemoteHostServiceTests.cs
```diff
@@ -0,0 +1,110 @@
+using Microsoft.AspNetCore.Http;
+using Moq;
+using RemoteHostIpAddress;
+using System.Net;
+using Xunit;
+
+namespace RemoteHostIpAddressTests
+{
+    public class IRemoteHostServiceTests
```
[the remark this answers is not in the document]
karimkod: Hello @amuslic, could you please double check this? unit tests should end with unit test `RemoteHostServiceUnitTest` I'm not sure this was applied. Thanks!

karimkod:
Hello @amuslic, 
For the sake of a better practice, it's not adviced to return nulls from methods. And as you know the methods created for getting the IP address all return nulls if we can't retrieve it. Do you think there is a nother way to do so, without returning null. Maybe return an empty Ip Address. I don't know if that's possible so let me know what you think. 
If you think that there is no other simple way around not retunning null, then we can keep it. 
Let me know what you think?

PR #1045, opened 2023-04-18
CM-331 URL encode and decode

asfd1234:


csharp-basic-topics/UrlEncodingAndDecoding/Tests/UrlEncodingDecodingTest.cs
```diff
@@ -0,0 +1,79 @@
+using System.Net;
+using System.Web;
+
+namespace Tests
+{
+    [TestClass]
+    public class UrlEncodingDecodingTest
```
karimkod, line 7: Can we change this to be UrlEncodingDecodingUnitTest, alongside the file name. 
Thanks!

karimkod:
Hey @asfd1234 @mcconachiej  can we please move the code to "aspnetcore-features" instead of "csharp-basic-topics'. Thanks!

PR #1050, opened 2023-04-19
CM-661 - Value vs Reference Types in C# - In Depth Guide

OldPen16:


csharp-basic-topics/ValueAndReferenceTypes/Tests/Test.cs
```diff
@@ -0,0 +1,77 @@
+using Microsoft.VisualStudio.TestTools.UnitTesting;
+using ValueAndReferenceTypes;
+namespace Tests
+{
+    [TestClass]
+    public class Test
```
karimkod, line 6: This class should have a meaningful name ending with "UnitTest" instead of just "Test"

csharp-basic-topics/ValueAndReferenceTypes/Tests/Test.cs
```diff
@@ -0,0 +1,77 @@
...
+        {
+            var pass = new PassByValueAndReferenceSample();
+            int counter = 100;
+            pass.PassByReference(ref counter);
+
+            Assert.AreEqual(101, counter);
+        }
+
+        [TestMethod]
+        public void When_BoxingUnBoxing()
```
karimkod, line 59: All test should have the When_Then sections and the underscore should be used to separate those two sections not like it's used here.

csharp-basic-topics/ValueAndReferenceTypes/Tests/Test.cs
```diff
@@ -0,0 +1,77 @@
...
+        {
+            var conv = new BoxingAndUnboxing();
+            int input = 500;
+            var output = conv.BoxingUnBoxing(input);
+
+            Assert.AreEqual(output, input);
+        }
+
+        [TestMethod]
+        public void When_StringAreReferenceType()
```
karimkod, line 69: All test should have the When_Then sections and the underscore should be used to separate those two sections not like it's used here.

csharp-basic-topics/ValueAndReferenceTypes/ValueAndReferenceTypes/ValueTypeSample.cs
```diff
@@ -0,0 +1,18 @@
+﻿namespace ValueAndReferenceTypes
+{
+    public struct ValueTypeRectangle
+    {
+        public int breadth;
+        public int length;
+        public int area;
```
karimkod, line 7: These should be public properties instead of public fields like this : 

```C#
        public int Breadth { get; set;}
        public int Length { get;  set;}
```

karimkod, line 7: This area is not used. can be removed.

csharp-basic-topics/ValueAndReferenceTypes/ValueAndReferenceTypes/ValueTypeSample.cs
```diff
@@ -0,0 +1,18 @@
...
+        public int Area()
```
karimkod, line 8: add a line before the method to separate it from properties.

csharp-basic-topics/ValueAndReferenceTypes/ValueAndReferenceTypes/ReferenceTypeSample.cs
```diff
@@ -0,0 +1,26 @@
...
+        private int _breadth = 20;
+        public int length { 
+            get { return _length; } 
+            set { _length = value; }
+        }
+        public int breadth
+        {
+            get { return _breadth; }
+            set { _breadth = value; }
+        }
```
karimkod, line 15: These properties are not useful since they just return the backed field. hey should be removed.

csharp-basic-topics/ValueAndReferenceTypes/ValueAndReferenceTypes/ReferenceTypeSample.cs
```diff
@@ -0,0 +1,26 @@
+﻿namespace ValueAndReferenceTypes
+{
+    public class ReferenceTypeRectangle
+    {
+        private int _length = 50;
+        private int _breadth = 20;
```
karimkod, line 6: These should be public properties instead of public fields like this :

```
        public int Breadth { get; set;}
        public int Length { get;  set;}
```

csharp-basic-topics/ValueAndReferenceTypes/ValueAndReferenceTypes/Program.cs
```diff
@@ -0,0 +1,10 @@
+﻿namespace ValueAndReferenceTypes
+{
+    public class ValueAndReferenceTypes
+    {
+        public static void Main(string[] args)
+        {
+
+        }
+    }
+}
```
karimkod, line 10: This can be safely deleted and just add a placeholder Console.WriteLine().

karimkod (changes requested):

karimkod (changes requested):

PR #1128, opened 2023-05-26
Fix RemoveDuplicateElementsBenchmark, because the results were wrong

jakubsuchybio:
- Upgrade to net7
- Switch from comparing speed on strings to comparing on ints
- Rewrite the 200 element array with 3 many duplicate strings to parametrized setup with N=100,10000 and Duplicates calculated by percentages 0,25,50,100

Because previous benchmark results were scewed towards O(n^2) algo as faster than O(n) algo, which is obviously bad result. Also benchmarking on comparing strings in the array might scew the results even further

Also wrote a comment to your article which is definitely wrong too, but can't edit that.
https://code-maze.com/csharp-array-remove-duplicates/


jakubsuchybio:
Ehm... Why didn't it get merged?

PR #1132, opened 2023-05-27
CM-90: Hashtable in C#

cjaynjoku:


collections-csharp/HashtableInCSharp/HashtableInCSharp/HashtableDemo.cs
```diff
@@ -0,0 +1,88 @@
+﻿using System.Collections;
+
+namespace HashtableInCSharp
+{
+    public class HashtableDemo
```
jshergal, line 5: There is no instance data in this class, the whole thing can be made static, including each of the methods

collections-csharp/HashtableInCSharp/HashtableInCSharp/HashtableDemo.cs
```diff
@@ -0,0 +1,88 @@
...
+    {
+        public Hashtable CreateHashTable()
```
jshergal, line 7: This method currently is not being used anywhere, only returns one of the three hash tables and is not covered by any tests. As I have no access to the article, I don't know if this is code that is being shown in the article as ways in which a HashTable can be initialized, but if so, then I would suggest breaking this into three methods, one for each HashTable, and then adding tests around each one.

collections-csharp/HashtableInCSharp/HashtableInCSharp/HashtableDemo.cs
```diff
@@ -0,0 +1,88 @@
...
+                { 2, new User() { FirstName = "Lone", LastName = "Costa"}},
+                { 3, new User() { FirstName = "Alberto", LastName = "Daci"} },
+            };
+            
+            var thirdHashTable = new Hashtable(dictionary);
+
+            return secondHashTable;
+        }
+
+        public Hashtable AddToHashTable(Hashtable hashTable)
```
jshergal, line 25: This, in my opinion, would be more intuitive as an extension method. I would also suggest renaming it to "AddSampleDataToHashTable"

collections-csharp/HashtableInCSharp/Tests/HashtableInCSharpTest.cs
```diff
@@ -0,0 +1,77 @@
...
+    {
+        private readonly HashtableDemo _hashtableDemo;
+        private readonly Hashtable _users;
+        public HashtableInCSharpTest()
+        {
+            _hashtableDemo = new HashtableDemo();
+            _users = _hashtableDemo.AddToHashTable(new Hashtable());
+        }
+        [Fact]
+        public void GivenAnEmptyHashtable_WhenAddingToAHashtable_ThenReturnsFilledEmptyHashTable()
```
jshergal, line 16: This test isn't actually adding anything to a HashTable, it should create an empty HashTable and then call the "AddToHashTable" method and finally verify that something was added.

collections-csharp/HashtableInCSharp/HashtableInCSharp/HashtableDemo.cs
```diff
@@ -0,0 +1,88 @@
...
+            {
+                User firstUser = (User)hashTable[1];
+
+                return firstUser;
+            }
+
+            return default;
+        }
+
+        public void RetrieveAllElementsFromHashTable(Hashtable hashTable)
```
jshergal, line 49: I would rename this to "PrintAllElements" as it is not retrieving them. Since these HashTable's are holding User items, either we should rename this method OR we could change it to return an IList<User> and internally populate a list that we then return.

jshergal:
One other thing that would be nice is benchmarks showing why Dictionary<TKey, TValue> is preferred over HashTable since .Net Framework 2.0. See this note: https://github.com/dotnet/platform-compat/blob/master/docs/DE0006.md

collections-csharp/HashtableInCSharp/Tests/HashtableInCSharpTest.cs
```diff
@@ -5,71 +5,120 @@ namespace Tests
...
-        public void GivenAnEmptyHashtable_WhenUpdatingElementInHashtable_ThenReturnsUpdatedHashTable()
+        public void GivenEmptyHashtable_WhenRetrievingAllElementFromHashtable_ThenReturnsAllUsers()
         {
-            var newUsers = _hashtableDemo.UpdateAnElementInHashTable(_users);
-            User user = (User)newUsers[1];
+            var users = HashtableDemo.RetrieveAllElementsFromHashTable(_userHashtable);
+
+            Assert.IsType<List<User>>(users);
+            Assert.NotEmpty(users);
+            Assert.Equal(5, users.Count);
```
jshergal, line 78: This assertion would be more maintainable and clearer if it said:
Assert.Equal(_userHashtable.Count, users.Count), as that is the actual assertion that is intended.

collections-csharp/HashtableInCSharp/Tests/HashtableInCSharpTest.cs
```diff
@@ -5,71 +5,120 @@ namespace Tests
...
+        }
+
+        [Fact]
+        public void GivenEmptyHashtable_WhenUpdatingElementInHashtable_ThenReturnsUpdatedHashTable()
+        {
+            var newUsers = HashtableDemo.UpdateElementInHashTable(_userHashtable, 3);
```
jshergal, line 84: The name of the method implies that the provided _userHashtable is going to be updated. While the method is returning a hashtable, it is not cloning it, it is returning _userHashtable. All that is fine, but in the assertions, since the method is claiming to update the provided Hashtable, it seems like the assertions should be against _userHashtable. Also, since you are asserting that the count did not change, you should store the current count before the operation and then in the assertion for the size, the count before the operation is the expected value that you are asserting. This helps remove the reliance on magic numbers and makes the intent clearer when reading the code.

collections-csharp/HashtableInCSharp/Tests/HashtableInCSharpTest.cs
```diff
@@ -5,71 +5,120 @@ namespace Tests
...
             var userLastName = user.LastName;
 
+            Assert.Equal(5, newUsers.Count);
             Assert.Equal("Henry", userFirstName);
             Assert.Equal("Stafford", userLastName);
         }
 
         [Fact]
-        public void GivenAnEmptyHashtable_WhenRemovingAnElementFromHashtable_ThenReturnsUpdatedHashTable()
+        public void GivenEmptyHashtable_WhenRemovingElementFromHashtable_ThenReturnsUpdatedHashTable()
```
jshergal, line 95: In this test, since you are verifying a removal of a particular element by key, there should be an assertion that the key exists before the removal, and then an assertion that the key no longer exists after the removal. Also, a size assertion is nice, but you should store the current size before the removal and you can assert that the new size is one less than the old size.

collections-csharp/HashtableInCSharp/HashtableInCSharp/HashtableDemo.cs
```diff
@@ -4,85 +4,99 @@ namespace HashtableInCSharp
...
-                { 3, new User() { FirstName = "Alberto", LastName = "Daci"} },
+                { 1, new User() { Id = 1, FirstName = "Rafa", LastName = "Lopez" } },
+                { 2, new User() { Id = 2, FirstName = "Michael", LastName = "Sam" } },
+                { 3, new User() { Id = 3, FirstName = "Sam", LastName = "Manalt"} },
+                { 4, new User() { Id = 4, FirstName = "Lone", LastName = "Costa"} },
+                { 5, new User() { Id = 5, FirstName = "Alberto", LastName = "Daci"} },
             };
             
-            var thirdHashTable = new Hashtable(dictionary);
+            var userHashTable = new Hashtable(dictionary);
```
jshergal, line 32: You could just add the return statement on this line

collections-csharp/HashtableInCSharp/HashtableInCSharp/HashtableDemo.cs
```diff
@@ -4,85 +4,99 @@ namespace HashtableInCSharp
...
 
-            return secondHashTable;
+            return userHashTable;
         }
 
-        public Hashtable AddToHashTable(Hashtable hashTable)
+        public static Hashtable AddSampleDataToHashTable(Hashtable userHashTable, List<User> userList)
```
jshergal, line 37: This comment applies to all of the methods that take a Hashtable as an argument. Since each of the methods is actually modifying the passed in Hashtable, it is confusing to also have it return that same Hashtable. It would make sense in a Fluent programming style, Builder type of pattern, but since the main idea of this class is methods showing how to perform specific operations on a Hashtable, I think making these methods void would be make the intent clearer. It would also help to clarify the unit tests where the operation is being called and the return value is being captured and then the assertions are being made on the return value. It could lead to confusion as to thinking the operations are not modifying the original, but rather returning a copy that has had the operations applied to it (like you have with the elements of the Immutable collections library).

collections-csharp/HashtableInCSharp/HashtableInCSharp/HashtableDemo.cs
```diff
@@ -0,0 +1,85 @@
...
+            }
+
+            return userHashTable;
+        }
+
+        public static User RetrieveElementFromHashTable(Hashtable userHashTable, int id)
+        {
+            if (userHashTable.ContainsKey(id))
+            {
+                User user = (User)userHashTable[id];
```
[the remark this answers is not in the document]
rbychiang: Thanks @jshergal.
@cjaynjoku Can you please apply this change? We don't need a local var here.

collections-csharp/HashtableInCSharp/HashtableInCSharp/HashtableDemo.cs
```diff
@@ -0,0 +1,85 @@
...
+
+                return user;
+            }
+
+            return default;
+        }
+
+        public static IList<User> RetrieveAllElementsFromHashTable(Hashtable userHashTable)
+        {
+            var userList = new List<User>();
```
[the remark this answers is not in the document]
rbychiang: Thanks @jshergal.
@cjaynjoku, can you please set the initial capacity `userHashTable.Count` when creating a new `List<User>` instance?

PR #1142, opened 2023-05-30
CM-726: WireMock.NET

Banovvv:


dotnet-client-libraries/WireMock.NET/WireMockNet/PlanetsService.cs
```diff
@@ -0,0 +1,28 @@
...
+
+namespace WireMockNet
+{
+    public class PlanetsService
+    {
+        private readonly HttpClient _httpClient;
+
+        public PlanetsService(HttpClient httpClient)
+        {
+            _httpClient = httpClient ??
```
jshergal, line 12: Because you have enabled nullables for the project (which is the default with new projects starting in .NET 6.0), the HttpClient is attributed to never be null so the check is unnecessary, and most likely the IDE would even be showing you that httpClient cannot be null here.

dotnet-client-libraries/WireMock.NET/WireMockNet/Planet.cs
```diff
@@ -0,0 +1,25 @@
+﻿namespace WireMockNet
+{
+    public class Planet
```
jshergal, line 3: This is a good candidate for a record, which will give you the Equality check for free and will then help to make the assertions in the test cleaner

dotnet-client-libraries/WireMock.NET/WireMock.NET.Tests/PlanetsServiceTests.cs
```diff
@@ -0,0 +1,85 @@
...
+    .RespondWith(
+        Response.Create()
+            .WithStatusCode(HttpStatusCode.OK)
+            .WithBodyAsJson(planet));
+
+            // Act
+            var result = await _planetService.GetPlanetByIdAsync(planet.Id);
+
+            // Assert
+            result.Should().NotBeNull();
```
jshergal, line 52: By changing Planet to be a record OR by implementing IEquatable, you can clean make these assertions clearer as this code is basically performing an equality check (with the absence of the ID property, which, since you are checking everything else and are expecting to get the planet with the given ID, this test should probably assert as well).

dotnet-client-libraries/WireMock.NET/WireMockNet/PlanetsService.cs
```diff
@@ -0,0 +1,28 @@
...
+                throw new ArgumentNullException(nameof(httpClient));
+        }
+
+        public async Task<Planet?> GetPlanetByIdAsync(int id)
+        {
+            var response = await _httpClient.GetAsync(_httpClient.BaseAddress + $"planets/{id}");
```
jshergal, line 18: No need to create the aboslute URI since the BaseAddress is already set on the HttpClient, you can just give the relative URI for the endpoint.

dotnet-client-libraries/WireMock.NET/WireMockNet/PlanetsService.cs
fl0tschi: Resolved in a separate PR https://github.com/CodeMazeBlog/CodeMazeGuides/pull/1182

dotnet-client-libraries/WireMock.NET/WireMockNet/Planet.cs
fl0tschi: Resolved in a separate PR https://github.com/CodeMazeBlog/CodeMazeGuides/pull/1182

dotnet-client-libraries/WireMock.NET/WireMock.NET.Tests/PlanetsServiceTests.cs
fl0tschi: Resolved in a separate PR https://github.com/CodeMazeBlog/CodeMazeGuides/pull/1182

PR #1145, opened 2023-05-31
CM-130: Guid class in c#

KonstantinosP7:


PhilBroderick (approved):
Looks alright.

csharp-basic-topics/GuidClassInCSharp/GuidClassInCSharpTests/GuidClassInCSharpUnitTests.cs
```diff
@@ -0,0 +1,54 @@
...
+    public class Tests
+    {
+        [Test]
+        public void GivenCheckGuidEqualityMethod_WhenEqualGuids_ThenTrue()
+        {
+            var guidA = new Guid("B5724FB4-2FA0-4E68-B108-93C24A1EA6DF");
+            var guidB = new Guid("B5724FB4-2FA0-4E68-B108-93C24A1EA6DF");
+            var checkEquality = CheckGuidEquality(guidA, guidB);
+
+            Assert.That(checkEquality, Is.EqualTo(true));
```
jshergal, line 14: Since you are using the constraint model approach of NUnit, this would be better as: Assert.That(checkEquality, Is.True)

csharp-basic-topics/GuidClassInCSharp/GuidClassInCSharpTests/GuidClassInCSharpUnitTests.cs
```diff
@@ -0,0 +1,54 @@
...
+        }
+
+        [Test]
+        public void GivenCheckGuidEqualityMethod_WhenNotEqualGuids_ThenFalse()
+        {
+            var guidA = Guid.NewGuid();
+            var guidB = Guid.NewGuid();
+            var checkEquality = CheckGuidEquality(guidA, guidB);
+
+            Assert.That(checkEquality, Is.EqualTo(false));
```
jshergal, line 24: See above comment about Is.True constraint.

csharp-basic-topics/GuidClassInCSharp/GuidClassInCSharpTests/GuidClassInCSharpUnitTests.cs
```diff
@@ -0,0 +1,54 @@
...
+            Assert.That(emptyGuid, Is.EqualTo(createdGuid));
+        }
+
+        [Test]
+        public void GivenCreateRandomGuidMethod_WhenCalled_ThenValidGuid()
+        {
+            var createdGuid = CreateRandomGuid();
+            var isValidGuid = Guid.TryParse(createdGuid.ToString(), out _);
+
+            Assert.That(isValidGuid, Is.EqualTo(true));
```
jshergal, line 42: Should be Is.True

csharp-basic-topics/GuidClassInCSharp/GuidClassInCSharp/Methods/GuidToByteString.cs
```diff
@@ -0,0 +1,12 @@
+﻿namespace GuidClassInCSharp
+{
+    partial class GuidClassInCSharpMethods
+    {
+        public static string GuidToByteString(Guid inputGuid)
+        {
+            var byteArray = inputGuid.ToByteArray();
```
jshergal, line 7: Since you are targetting .NET 7, you can use a stackalloc Span here and call the guid.TryWriteBytes to save on the array allocation.
Span<byte> dest = stackalloc byte[16];
inputGuid.TryWriteBytes(dest);

csharp-basic-topics/GuidClassInCSharp/GuidClassInCSharp/Methods/GuidToByteString.cs
```diff
@@ -0,0 +1,12 @@
...
+
+            return BitConverter.ToString(byteArray);
```
jshergal, line 9: This may be fine, if for the purpose of the article you want a hex string separated by dashes, otherwise, since you are targetting .NET 6.0+, you can use the Convert.ToHexString method. I understand that in this case the performance doesn't really matter, but when possible we should try to show best practices even in our sample code.
https://learn.microsoft.com/en-us/dotnet/api/system.convert.tohexstring?view=net-7.0

csharp-basic-topics/GuidClassInCSharp/GuidClassInCSharp/Methods/GuidToStringMethod.cs
```diff
@@ -0,0 +1,10 @@
+﻿namespace GuidClassInCSharp
+{
+    public partial class GuidClassInCSharpMethods
+    {
+        public static string GuidToString(Guid guid)
```
jshergal, line 5: Guid.ToString takes format arguments, it would be good to add an additional format parameter to show the options available with a call to ToString (i.e. ToString("X") or ToString("N"))

csharp-basic-topics/GuidClassInCSharp/GuidClassInCSharpTests/GuidClassInCSharpUnitTests.cs
```diff
@@ -0,0 +1,54 @@
...
+        public void GivenCreateEmptyGuidMethod_WhenCalled_ThenGuidEmpty()
+        {
+            var emptyGuid = new Guid("00000000-0000-0000-0000-000000000000");
+            var createdGuid = CreateEmptyGuid();
+
+            Assert.That(emptyGuid, Is.EqualTo(createdGuid));
+        }
+
+        [Test]
+        public void GivenCreateRandomGuidMethod_WhenCalled_ThenValidGuid()
```
jshergal, line 37: For completeness this should maybe have an additional Assert that the created random GUID is not the empty GUID

csharp-basic-topics/GuidClassInCSharp/GuidClassInCSharpTests/GuidClassInCSharpUnitTests.cs
```diff
@@ -0,0 +1,54 @@
+using static GuidClassInCSharp.GuidClassInCSharpMethods;
+
+namespace GuidClassInCSharpTests
+{
+    public class Tests
+    {
+        [Test]
```
jshergal, line 7: Maybe this test would be clearer with a TestCase attribute in place of the Test attribute.
[TestCase("B5724FB4-2FA0-4E68-B108-93C24A1EA6DF")]
public void GivenCheckGuidEqualityMethod_WhenEqualGuids_ThenTrue(string guidValue)

csharp-basic-topics/GuidClassInCSharp/GuidClassInCSharpTests/GuidClassInCSharpUnitTests.cs
```diff
@@ -0,0 +1,54 @@
...
+        public void GivenCreateRandomGuidMethod_WhenCalled_ThenValidGuid()
+        {
+            var createdGuid = CreateRandomGuid();
+            var isValidGuid = Guid.TryParse(createdGuid.ToString(), out _);
+
+            Assert.That(isValidGuid, Is.EqualTo(true));
+        }
+
+        [Test]
+        public void GivenGuidToByteStringMethod_WhenGivenGuid_ThenByteArray()
```
jshergal, line 46: This is another good candidate for the TestCase attribute

csharp-basic-topics/GuidClassInCSharp/GuidClassInCSharp/Methods/GuidToByteString.cs
KonstantinosP7: In this specific case, I wanted to show the separate hexadecimal digits and notify that they are a collection of hexadecimal pairs. That said, I preferred "31-A7-5A-C8-0C-24-D5-46-98-A8-ED-98-84-97-F8-03" over "31A75AC80C24D54698A8ED988497F803".

csharp-basic-topics/GuidClassInCSharp/GuidClassInCSharp/Methods/GuidToByteString.cs
KonstantinosP7: Here, I wanted to demonstrate the .ToByteArray() extension explicitly, and I tried to keep it simple. Should I leave it as it is or add a new section with the extension .TryWriteBytes() ?

csharp-basic-topics/GuidClassInCSharp/GuidClassInCSharp/Methods/GuidToByteString.cs
jshergal: 👍I thought that may be the case, just commented to be sure that there was a specific goal

csharp-basic-topics/GuidClassInCSharp/GuidClassInCSharp/Methods/GuidToByteString.cs
jshergal: In my  opinion both are simple and the latter approach (the one I suggested) gives an extra nudge to noalloc/lowalloc code writing, which is an important factor when talking about microservices, mobile devices, embedded systems, etc. Basically, my feeling is even if we are on a desktop, we are still gaining performance and reducing allocation overhead so why not? When people start looking at doing things like .NET MAUI then you have no idea where the eventual code may run. I do understand the trap of early optimization, but in this case it is basically the same code, just with an allocation removed, so we aren't prematurely optimizing and we are not complicating the code. The only caveat with this sample would be if someone was calling it in a recursive loop, but someone who saw this code and wanted to use it in that context would already be aware of that and could make the change to use ToByteArray themselves.

That is my feeling - you should check it with your editor, I am just adding some basic counsel.

PR #1150, opened 2023-06-02
CM-742: Permutations of a String in C#

MPrtenjak:
Code for the article: "Permutations of a String in C#" (CM-742)

### DIFF
--- a/csharp-algorithms/PermutationsOfAString/.editorconfig
+++ b/csharp-algorithms/PermutationsOfAString/.editorconfig
@@ -1,4 +1,0 @@
-[*.{cs,vb}]
-indent_style = space
-indent_size = 4
-tab_width = 4
--- a/csharp-algorithms/PermutationsOfAString/.editorconfig
+++ b/csharp-algorithms/PermutationsOfAString/.editorconfig
@@ -1,0 +1,4 @@
+[*.{cs,vb}]
+indent_style = space
+indent_size = 4
+tab_width = 4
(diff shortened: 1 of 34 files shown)


csharp-algorithms/PermutationsOfAString/.editorconfig
jshergal, line 1: This file shouldn't be included in the PR

MPrtenjak:
Hi,

It is included in all my other code samples on code-maze, and a few other authors also include it. Not only should it be included, but it must be included so the code is properly aligned according to code-maze guides.

jshergal:
> Hi,
> 
> It is included in all my other code samples on code-maze, and a few other authors also include it. Not only should it be included, but it must be included so the code is properly aligned according to code-maze guides.

As per the Author guidlines it should not be included:

> make sure to only include the relevant changes that are necessary for your article and nothing else: 
> - no formatting changes
> - no IDE artifacts (Visual Studio stuff)
> - no build output (bin, debug, etc)
> - basically, nothing that isn’t critical to the article

This file is an IDE artifact and on top of that it will override a persons local [VisualStudio settings](https://learn.microsoft.com/en-us/visualstudio/ide/create-portable-custom-editor-options?view=vs-2022).

Regarding the Code-Maze alignments, that is correct as to how things appear in an article. If you need these settings locally so that your code is formatted correctly for the article, that is perfectly reasonable, but it should not end up getting checked in as it is a VisualStudio artifact.

MPrtenjak:
Hi,

I have removed the .editorconfig file from the project

PR #1157, opened 2023-06-04
CM-213: Difference Between == and Equals() in C#

ahmedryanfaiyaz:


csharp-basic-topics/EqualityOperatorVsEqualsMethodInCSharp/EqualityOperatorVsEqualsMethodInCSharp/Program.cs
```diff
@@ -0,0 +1,119 @@
...
+                classInstance1.Equals(classInstance2) // false
+            );
+
+            // Struct
+            var structInstance1 = new MyStruct("Ahmed", 2);
+            var structInstance2 = new MyStruct("Ahmed", 2);
+
+            CompareResults(
+                ReferenceEquals(structInstance1, structInstance2), // false
+                null, // structInstance1 == structInstance2 throws a compilation error
```
jshergal, line 97: To fix the compiler error you need to override == in the struct. You just need to implement the IEquatable<T> interface along with overriding Equals.
https://learn.microsoft.com/en-us/dotnet/csharp/programming-guide/statements-expressions-operators/how-to-define-value-equality-for-a-type
For the purpose of consistency with the other code lines, it is probably worth implementing. You could then update the article text with the explanation of why it was necessary to implement the interface and why you may want to do that for any given class or struct. You could use that as a nice segway into discussing the advantages of records. We also now have record structs so you could explore both class records and struct records if so desired.

PR #1159, opened 2023-06-05
CM-697: EntityFrameworkCoreVsDapper

SteliosPantelopoulos:
Initial commit


dotnet-efcore/EntityFrameworkCoreVsDapper/Tests/PersonRepositoryDapperUnitTests.cs
```diff
@@ -0,0 +1,34 @@
...
+
+            Assert.True(1000 == queryResults.Count());
+        }
+
+        [Fact]
+        public void When_Quering10000Persons_ThenReturn10000Objects()
+        {
+            var queryResults = repository.Get_10000_Persons();
+
+            Assert.True(10000 == queryResults.Count());
```
jshergal, line 23: Since the result is a List you can use the List.Count property directly rather than the IEnumerable LINQ Count function.

dotnet-efcore/EntityFrameworkCoreVsDapper/Tests/PersonRepositoryDapperUnitTests.cs
```diff
@@ -0,0 +1,34 @@
...
+        }
+
+        [Fact]
+        public void When_Quering10000Persons_ThenReturn100000Objects()
+        {
+            var queryResults = repository.Get_100000_Persons();
+
+            Assert.True(100000 == queryResults.Count());
```
jshergal, line 31: See earlier comment about the List.Count property

dotnet-efcore/EntityFrameworkCoreVsDapper/Tests/PersonRepositoryDapperUnitTests.cs
```diff
@@ -0,0 +1,34 @@
...
+    public class PersonRepositoryDapperUnitTests
+    {
+        private PersonsRepositoryDapper repository = new PersonsRepositoryDapper();
+
+        [Fact]
+        public void When_Quering1000Persons_ThenReturn1000Objects()
+        {
+            var queryResults = repository.Get_1000_Persons();
+
+            Assert.True(1000 == queryResults.Count());
```
jshergal, line 15: Would be cleaner to use both List.Count property and Assert.Equal(1000, queryReults.Count), as the intention is an equality check.

dotnet-efcore/EntityFrameworkCoreVsDapper/Tests/PersonRepositoryEFCoreUnitTests.cs
```diff
@@ -0,0 +1,35 @@
...
+    public class PersonRepositoryEFCoreUnitTests
+    {
+        private PersonsRepositoryEFCore repository = new PersonsRepositoryEFCore();
+
+        [Fact]
+        public void When_Quering1000Persons_ThenReturn1000Objects()
+        {
+            var queryResults = repository.Get_1000_Persons();
+
+            Assert.True(1000 == queryResults.Count());
```
jshergal, line 15: Assert.Equal and List.Count would be better here

dotnet-efcore/EntityFrameworkCoreVsDapper/Tests/PersonRepositoryEFCoreUnitTests.cs
```diff
@@ -0,0 +1,35 @@
...
+        }
+
+        [Fact]
+        public void When_Quering10000Persons_ThenReturn10000Objects()
+        {
+            var queryResults = repository.Get_10000_Persons();
+
+            Assert.True(10000 == queryResults.Count());
```
jshergal, line 23: Assert.Equal and List.Count would be better here

dotnet-efcore/EntityFrameworkCoreVsDapper/Tests/PersonRepositoryEFCoreUnitTests.cs
```diff
@@ -0,0 +1,35 @@
...
+        }
+
+        [Fact]
+        public void When_Quering10000Persons_ThenReturn100000Objects()
+        {
+            var queryResults = repository.Get_100000_Persons();
+
+            Assert.True(100000 == queryResults.Count());
```
jshergal: Assert.Equal and List.Count would be better here

dotnet-efcore/EntityFrameworkCoreVsDapper/EntityFrameworkCoreVsDapper/Repositories/PersonsRepositoryDapper.cs
```diff
@@ -0,0 +1,52 @@
...
+{
+    public class PersonsRepositoryDapper : IPersonsRepository
+    {
+        //ADD THE CONNECTION STRING TO YOUR DATABASE HERE
+        private string ConnectionString = "";
+
+        [Benchmark]
+        public List<Person> Get_1000_Persons()
+        {
+            using (var dbConnection = new SqlConnection(ConnectionString))
```
jshergal, line 16: You can switch to the using declaration to reduce nesting since the project is targetting .NET 7

dotnet-efcore/EntityFrameworkCoreVsDapper/EntityFrameworkCoreVsDapper/Repositories/PersonsRepositoryDapper.cs
```diff
@@ -0,0 +1,52 @@
...
+                dbConnection.Close();
+
+                return items;
+            }
+        }
+
+        [Benchmark]
+        public List<Person> Get_10000_Persons()
+        {
+            using (var dbConnection = new SqlConnection(ConnectionString))
```
jshergal, line 29: You can switch to the using declaration to reduce nesting since the project is targetting .NET 7

dotnet-efcore/EntityFrameworkCoreVsDapper/EntityFrameworkCoreVsDapper/Repositories/PersonsRepositoryDapper.cs
```diff
@@ -0,0 +1,52 @@
...
+                dbConnection.Close();
+
+                return items;
+            }
+        }
+
+        [Benchmark]
+        public List<Person> Get_100000_Persons()
+        {
+            using (var dbConnection = new SqlConnection(ConnectionString))
```
jshergal, line 42: You can switch to the using declaration to reduce nesting since the project is targetting .NET 7

dotnet-efcore/EntityFrameworkCoreVsDapper/EntityFrameworkCoreVsDapper/Repositories/PersonsRepositoryDapper.cs
```diff
@@ -0,0 +1,52 @@
...
+
+namespace EntityFrameworkCoreVsDapper.Repositories
+{
+    public class PersonsRepositoryDapper : IPersonsRepository
+    {
+        //ADD THE CONNECTION STRING TO YOUR DATABASE HERE
+        private string ConnectionString = "";
+
+        [Benchmark]
+        public List<Person> Get_1000_Persons()
```
jshergal, line 14: Instead of copy/pasted code, this would be better with a private method that takes a count parameter and then each of these methods internally could call it. You could even change the interface to be a single method with a count parameter, and then use a [parameterized benchmark](https://andreyakinshin.github.io/bdn-docs-wip/articles/features/parameterization.html#sample-introarguments) to run the three versions. Either way is probably okay, but for sure the internal duplicate code should be extracted to a private method. The same concept applies for the EFCore code.

dotnet-efcore/EntityFrameworkCoreVsDapper/EntityFrameworkCoreVsDapper/Repositories/PersonsRepositoryDapper.cs
```diff
@@ -11,42 +11,15 @@ public class PersonsRepositoryDapper : IPersonsRepository
...
-            using (var dbConnection = new SqlConnection(ConnectionString))
-            {
-                dbConnection.Open();
-                var items = (List<Person>)dbConnection.Query<Person>("SELECT TOP(100000) * FROM Persons");
-                dbConnection.Close();
-
-                return items;
-            }
+            dbConnection.Open();
+            return (List<Person>)dbConnection.Query<Person>($"SELECT TOP({personsNumber}) * FROM Persons");
```
jshergal, line 22: I may not be correct on this, so please disregard if I am not. I have not used Dapper all that much, but I was thinking that the Query<T>() method returns an IEnumerable<T>. If that is the case, then you should be calling ToList() rather than casting. If you look at the Dapper source code, within the SqlMapper, ther result of the Query call calls to an internal QueryImpl method which involves yield return statements. Maybe there is magic in the compiler that fixes this up, which is why I am mentioning it, but you, the editor and head code reviewer can decide on correctness.

dotnet-efcore/EntityFrameworkCoreVsDapper/Tests/PersonRepositoryDapperUnitTests.cs
```diff
@@ -8,27 +8,11 @@ public class PersonRepositoryDapperUnitTests
         private PersonsRepositoryDapper repository = new PersonsRepositoryDapper();
 
         [Fact]
```
jshergal, line 10: You could make this a Theory and just pass in the 100, 1000, and 10000 values. It will shorten the code and make it more maintainable.

PR #1160, opened 2023-06-05
code for action and func delegate article is added

smhnkmr:
conde code and tests for action and func delegate evaluation article


smhnkmr:
ignore this PR

PR #1164, opened 2023-06-06
CM-375: Compare Byte Arrays in .NET

martinchege324:
First draft


collections-arrays/CompareByteArraysInCsharp/CompareByteArraysInCsharp/CompareByteArrays.cs
```diff
@@ -0,0 +1,147 @@
...
+                var hashFirstArray = hashInstance.ComputeHash(firstArray);
+                var hashSecondArray = hashInstance.ComputeHash(secondArray);
+
+                return hashFirstArray.SequenceEqual(hashSecondArray);
+            }
+        }
+
+        [ArgumentsSource(nameof(SampleByteArray))]
+        [Benchmark]
+        public bool CompareUsingHashSet(byte[] firstArray, byte[] secondArray)
```
jshergal, line 79: This method will produce an incorrect equality comparison. You can easily produce two different arrays with different ordering or even different lengths that will return true

collections-arrays/CompareByteArraysInCsharp/CompareByteArraysInCsharp/CompareByteArrays.cs
```diff
@@ -0,0 +1,147 @@
...
+    public class CompareByteArrays
+    {
+        private readonly Random _rand;
+        private readonly byte[] _byteArray;  
+
+        public CompareByteArrays()
+        {
+            _rand = new Random();
+            _byteArray = GenerateRandomArray(10000000);
+        }
```
PhilBroderick, line 20: Space required after this bracket.

collections-arrays/CompareByteArraysInCsharp/CompareByteArraysInCsharp/CompareByteArrays.cs
```diff
@@ -0,0 +1,147 @@
...
+            }
+
+            return true;
+        }
+
+        [ArgumentsSource(nameof(SampleByteArray))]
+        [Benchmark]
+        public bool CompareUsingIStructuralEquatable(byte[] firstArray, byte[] secondArray)
+        {
+
```
PhilBroderick, line 120: Unnecessary space.

collections-arrays/CompareByteArraysInCsharp/CompareByteArraysInCsharp/CompareByteArrays.cs
```diff
@@ -0,0 +1,147 @@
...
+        private static extern int memcmp(byte[] firstArray, byte[] secondArray, int size);
+        [ArgumentsSource(nameof(SampleByteArray))]
+        [Benchmark]
+        public bool CompareUsingPInvoke(byte[] firstArray, byte[] secondArray)
+        {
+            if (firstArray == null || secondArray == null || firstArray.Length != secondArray.Length)
+                return false;
+
+            return memcmp(firstArray, secondArray, firstArray.Length) == 0;
+        }
```
PhilBroderick, line 137: Space after this.

PhilBroderick (changes requested):
Couple of formatting fixes, plus the comment from Jeff that needs to be investigated.

collections-arrays/CompareByteArraysInCsharp/CompareByteArraysInCsharp/CompareByteArrays.cs
```diff
@@ -0,0 +1,147 @@
...
+        public bool CompareUsingIStructuralEquatable(byte[] firstArray, byte[] secondArray)
+        {
+
+            if (firstArray == null || secondArray == null || firstArray.Length != secondArray.Length)
+                return false;
+
+            return StructuralComparisons.StructuralEqualityComparer.Equals(firstArray, secondArray);
+        }
+
+        [DllImport("msvcrt.dll", CallingConvention = CallingConvention.Cdecl)]
```
jshergal, line 127: This is a Windows platform specific DllImport, but the method exists in the standard C library on other platforms. Unfortunately the naming is different. The C library on other platforms is libc. You just need a #If around this so that it gets the correct library on other platforms:
#if WIN32
[DllImport("msvcrt", CallingConvention ...
#else
[DllImport("libc", CallingConvention ...

collections-arrays/CompareByteArraysInCsharp/CompareByteArraysInCsharp/CompareByteArrays.cs
```diff
@@ -0,0 +1,147 @@
...
+        {
+            var firstSet = new HashSet<byte>(firstArray);
+            var secondSet = new HashSet<byte>(secondArray);
+
+            return firstSet.SetEquals(secondSet);
+        }
+
+        [ArgumentsSource(nameof(SampleByteArray))]
+        [Benchmark]
+        public bool CompareUsingBinaryEquality(byte[] firstArray, byte[] secondArray)
```
jshergal, line 89: Since this is already unsafe code, and you are focused on byte arrays, you can improve the performance of this method by performing most of the comparisons as ints and then just doing the remaining couple of bytes as a byte comparison.

collections-arrays/CompareByteArraysInCsharp/CompareByteArraysInCsharp/CompareByteArrays.cs
```diff
@@ -0,0 +1,147 @@
...
+    {
+        private readonly Random _rand;
+        private readonly byte[] _byteArray;  
+
+        public CompareByteArrays()
+        {
+            _rand = new Random();
+            _byteArray = GenerateRandomArray(10000000);
+        }
+        public IEnumerable<object[]> SampleByteArray()
```
jshergal, line 21: The current benchmark setup is in a sense invalid as you have two cases of the same size, one being the worst case scenario where the arrays are equal, and the other being two unequal arrays. Since the unequal case is created randomly, the benchmark will vary greatly and very possible could find a difference on the very first character of the array.
I could be wrong on this, but my gut feeling is a better way to benchmark would be to show a worst case as you have with one of the cases, and then have a second case where the two are the same except for changing one value in the middle. If you want to also have a case where the first element is different so you are dealing with the overhead into the method and a quick return, then that could be a third test case. For those I would add a label or something so it is clear in the results which case is which.

collections-arrays/CompareByteArraysInCsharp/CompareByteArraysInCsharp/CompareByteArrays.cs
```diff
@@ -0,0 +1,147 @@
...
+            }
+
+            return true;
+        }
+
+        [ArgumentsSource(nameof(SampleByteArray))]
+        [Benchmark]
+        public bool CompareUsingHash(byte[] firstArray, byte[] secondArray) 
+        {
+            using (var hashInstance = SHA256.Create())
```
jshergal, line 68: I originally avoided adding this comment as this method should be used for the purpose of an equality check, but came back and decided to add it so you would have the information.
You don't need the using and the creation of the hasher, you can just call the static method SHA256.HashData

jshergal: Just realized that there is an error in my comment here - what I meant to say was "this method should NOT be used for the purpose of an equality check"

collections-arrays/CompareByteArraysInCsharp/CompareByteArraysInCsharp/CompareByteArrays.cs
jshergal: I think I may have added this comment elsewhere, but this method could be enhanced even further by making use of the the Vector class or if you wanted to hand code it a bit more - Vector128 and calling into SIMD and AVX classes where available.

collections-arrays/CompareByteArraysInCsharp/CompareByteArraysInCsharp/CompareByteArrays.cs
```diff
@@ -64,75 +64,102 @@ public bool CompareUsingForLoop(byte[] firstArray, byte[] secondArray)
...
-        private byte[] GenerateRandomArray(int size) 
+        private byte[] GenerateOrderedArray(bool firstElement, bool middleElement, int size)
         {
-            var byteArray = new byte[size];
+            var intArray = new int[size];
+            var byteArray = new byte[size * sizeof(int)];
+
+            for (int i = 0; i < size; i++)
+            {
+                if (firstElement == true && i == 0)
```
jshergal, line 146: You can simplify this with just if (firstElement && i == 0). There is no need to check == for boolean values. The same applies for the line with middleElement.

This whole method could be simplified a little bit (in my opinion, feel free to disagree). You could just use the single byte array declaration. Then you have the option of doing it as an unsafe block, fixing the array and then casting the fixed pointer to an int pointer. Or you could use the MemoryMarshal.GetArrayDataReference method to take a reference to the array (no need for an unsafe block). You can then use the Unsafe.As to get an int reference to the byte array. You just have to be careful with your index value, but you would then be looking at the byte array as an int array so it would remove the need for the extra array allocation as well as the copy. Again, that is just my opinion.

martinchege324: @jshergal, that's a great idea. I will be pushing another version shortly.

PhilBroderick (approved):

PR #1166, opened 2023-06-06
 How to use Lazy<T> in C#: CM-723

NarmadhaDodda:
Added examples for using Lazy<T> in C# to improve performance by deferring object initialization until it's actually needed.


csharp-advanced-topics/UsingLazyInCSharp/HowtoUseLazyInCSharp/Cache.cs
```diff
@@ -0,0 +1,28 @@
...
+        }
+
+        public void Add(string key, object value)
+        {
+            cacheDictionary[key] = value;
+        }
+
+        public object Get(string key)
+        {
+            if (cacheDictionary.ContainsKey(key))
```
jshergal, line 20: This should use Dictionary.TryGetValue rather than performing a double key lookup in the expected/happy path.

strings-csharp/ConvertStringToSpan/ConvertStringToSpan/StringExample.cs
```diff
@@ -0,0 +1,59 @@
...
+        }
+
+        [Benchmark]
+        public unsafe Span<char> ConvertStringToSpanUsingUnsafe()
+        {
+            unsafe
+            {
+                fixed (char* ptr = _myString)
+                {
+                    return new Span<char>(ptr, _myString.Length);
```
jshergal, line 30: I know this is code from an already published article that somehow bled into this PR, but this code is very dangerous. The Span is targeting heap memory that has been fixed, after the return the fixed block has ended and the target heap object could be moved and now, in theory the Span could be pointing to memory that no longer contains the target string.

PR #1173, opened 2023-06-09
CM-366: Converting Hex Strings to Byte array

pradeepradyumna:


strings-csharp/HexToByteArray/Tests/Test.cs
```diff
@@ -0,0 +1,88 @@
...
+        public void WhenInvalidFormatHexPassed_ThenConverionFail()
+        {
+            var expectedResult = "FormatException";
+            try
+            {
+                var result = _convertHex.FromHexString("Hello");
+            }
+            catch (Exception ex)
+            {
+                Assert.Equal(expectedResult, ex.GetType().Name);
```
jshergal, line 33: xUnit has exception assertions, you should use them. The syntax is:
Assert.Throws<exT>, so in this case it would be:
Assert.Throws<FormatException>(() => _convertHex.FromHexString("Hello"));

strings-csharp/HexToByteArray/Tests/Test.cs
```diff
@@ -0,0 +1,88 @@
...
+            }
+        }
+
+        [Fact]
+        public void WhenNullatHexPassed_ThenConverionFail()
+        {
+            var expectedResult = "ArgumentNullException";
```
jshergal, line 40: See previous comment about Assert.Throws

strings-csharp/HexToByteArray/Tests/Test.cs
```diff
@@ -0,0 +1,88 @@
...
+            }
+        }
+
+        [Fact]
+        public void WhenNullatHexPassedToAlternative_ThenConverionFail()
+        {
+            var expectedResult = "ArgumentNullException";
+            try
+            {
+                var result = _convertHex.FromHexStringAlternative("");
```
jshergal, line 80: Woudn't it be expected that a blank string returns an empty array. That is what the new built in convert does.

strings-csharp/HexToByteArray/HexToByteArray/ConvertHex.cs
```diff
@@ -0,0 +1,35 @@
...
+        public byte[] FromHexString(string hex)
+        {
+            try
+            {
+                return Convert.FromHexString(hex);
+            }
+            catch (Exception)
+            {
+
+                throw;
```
jshergal, line 14: If you are just catching an exception so you can rethrow it, you should just remove the try/catch blockk

strings-csharp/HexToByteArray/HexToByteArray/ConvertHex.cs
```diff
@@ -0,0 +1,35 @@
...
+            }
+        }
+
+        public byte[] FromHexStringAlternative(string hex)
+        {
+            try
+            {
+                int length = hex.Length;
```
jshergal, line 22: use var

strings-csharp/HexToByteArray/HexToByteArray/ConvertHex.cs
```diff
@@ -0,0 +1,35 @@
...
+                byte[] bytes = new byte[length / 2];
```
jshergal, line 23: use var

strings-csharp/HexToByteArray/HexToByteArray/ConvertHex.cs
```diff
@@ -0,0 +1,35 @@
...
+                for (int i = 0; i < length; i += 2)
+                    bytes[i / 2] = Convert.ToByte(hex.Substring(i, 2), 16);
```
jshergal, line 25: You should be using a span here instead of Substring. This call is creating lots of small temporary strings. If you added a benchmark that used a longer string, something like 1024 characters or 2048 you would really start to see the effect of this. By taking the string as a ReadOnlySpan you can just slice the characters you want (span.Slice(i, 2)). Instead of Convert.ToByte you can use byte.Parse or TryParse. With Parse it will just throw an exception if the characters aren't in a proper format. Note that you have to provide the NumberStyles.HexNumber for Parse to work properly.

strings-csharp/HexToByteArray/Tests/Test.cs
```diff
@@ -0,0 +1,88 @@
...
+        private const string HexString = "FF008000";
+        public Test()
+        {
+            _convertHex = new ConvertHex();
+        }
+
+        [Fact]
+        public void WhenValidHexPassed_ThenConverionSuccessful()
+        {
+            byte[] expectedResult = { 255, 0, 128, 0 };
```
jshergal, line 17: Since you are reusing the same expected result along with a const string, it would be better to make this a member of the class that is defined alongside the const HexString

strings-csharp/HexToByteArray/HexToByteArray/ConvertHex.cs
```diff
@@ -0,0 +1,35 @@
...
+                return Convert.FromHexString(hex);
+            }
+            catch (Exception)
+            {
+
+                throw;
+            }
+        }
+
+        public byte[] FromHexStringAlternative(string hex)
```
jshergal, line 18: I did a simple test with a 64 character hex string (32 byte resultant array), and by making just a couple of those small changes the code will perform twice as fast and drop memory allocations from 1K to < 60 bytes. Here is the code (note, I also removed the try/catch)

`public byte[] FromHexStringAlternativeUpdated(ReadOnlySpan<char> hex)`
`{`
`        int length = hex.Length;`
`        byte[] bytes = GC.AllocateUninitializedArray<byte>(length >> 1);`
`        for (int i = 0; i < length; i += 2)`
`            bytes[i >> 1] = byte.Parse(hex.Slice(i, 2),`
`                System.Globalization.NumberStyles.HexNumber);`
`        return bytes;`
 `   }`

Here is the benchmark results:
|                          Method |     Mean |    Error |   StdDev |   Gen0 | Allocated |
|-------------------------------- |---------:|---------:|---------:|-------:|----------:|
|        FromHexStringAlternative | 675.0 ns | 12.05 ns | 16.89 ns | 0.5159 |    1080 B |
| FromHexStringAlternativeUpdated | 338.8 ns |  1.17 ns |  1.09 ns | 0.0267 |      56 B |

strings-csharp/HexToByteArray/HexToByteArray/ConvertHex.cs
```diff
@@ -0,0 +1,35 @@
+﻿namespace HexToByteArray
+{
+    public class ConvertHex
```
jshergal, line 3: Since there is no instance data in this class, the whole thing can be made static

strings-csharp/HexToByteArray/HexToByteArray/Program.cs
```diff
@@ -0,0 +1,22 @@
+﻿// See https://aka.ms/new-console-template for more information
```
jshergal, line 1: This line can be removed

strings-csharp/HexToByteArray/HexToByteArray/Program.cs
```diff
@@ -0,0 +1,22 @@
...
+using BenchmarkDotNet.Running;
+using HexToByteArray;
+
+string hexString = "FF008000"; // Hex representaion of color System.Drawing.Color.Green
+
+var convertHex = new ConvertHex();
+
+// Run the built-in first and display the result
+var bytes = convertHex.FromHexString(hexString);
+foreach (var b in bytes)
```
jshergal, line 12: This could be extracted to a local helper method, something like "PrintArray". It might also be nice to add some simple formatting. Something like "{1,2,3,4}" so that the array output is more readable. You could use the string.Join method to comma separate the array.

strings-csharp/HexToByteArray/HexToByteArray/Program.cs
```diff
@@ -0,0 +1,22 @@
...
+    Console.WriteLine(b);
+
+// Run the alternative approach and display the result
+var bytesAlternatives = convertHex.FromHexStringAlternative(hexString);
+foreach (var b in bytesAlternatives)
+    Console.WriteLine(b);
+
+// Benchmark
+var summary = BenchmarkRunner.Run<BenchmarkClass>();
```
jshergal, line 21: You can leave off the variable here and just call BenchmarkRunner.Run.

strings-csharp/HexToByteArray/Tests/Test.cs
```diff
@@ -0,0 +1,88 @@
+using HexToByteArray;
+
+namespace Tests
+{
+    public class Test
+    {
+        private readonly ConvertHex _convertHex;
```
jshergal, line 7: By making the ConvertHex class static, you can remove this line as well as the test constructor

strings-csharp/HexToByteArray/Tests/Test.cs
pradeepradyumna: Done. Thanks!

strings-csharp/HexToByteArray/Tests/Test.cs
pradeepradyumna: Done. Thanks!

strings-csharp/HexToByteArray/Tests/Test.cs
pradeepradyumna: Done. Thanks!

strings-csharp/HexToByteArray/HexToByteArray/ConvertHex.cs
pradeepradyumna: Done. Thanks!

strings-csharp/HexToByteArray/HexToByteArray/ConvertHex.cs
pradeepradyumna: Done. Thanks!

strings-csharp/HexToByteArray/Tests/Test.cs
```diff
@@ -0,0 +1,88 @@
...
+        private const string HexString = "FF008000";
+        public Test()
```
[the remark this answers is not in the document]
pradeepradyumna: Since I have made the ConvertHex class static now, I removed the constructor.

strings-csharp/HexToByteArray/HexToByteArray/ConvertHex.cs
pradeepradyumna: Done. Thanks!

strings-csharp/HexToByteArray/HexToByteArray/ConvertHex.cs
pradeepradyumna: Done. Thanks!

strings-csharp/HexToByteArray/HexToByteArray/Program.cs
pradeepradyumna: Done. Thanks!

strings-csharp/HexToByteArray/HexToByteArray/Program.cs
pradeepradyumna: Done. Thanks!

strings-csharp/HexToByteArray/Tests/Test.cs
pradeepradyumna: Done. Thanks!

strings-csharp/HexToByteArray/Tests/Test.cs
```diff
@@ -0,0 +1,88 @@
...
+            }
+        }
+
+        [Fact]
+        public void WhenNullatHexPassed_ThenConverionFail()
+        {
+            var expectedResult = "ArgumentNullException";
+            try
+            {
+                var result = _convertHex.FromHexString("");
```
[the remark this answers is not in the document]
pradeepradyumna: You are right. After using Assert.Throw, I handled it differently now.

strings-csharp/HexToByteArray/Tests/Test.cs
```diff
@@ -4,85 +4,66 @@ namespace Tests
...
-            {
-                var result = _convertHex.FromHexString("Hello");
-            }
-            catch (Exception ex)
-            {
-                Assert.Equal(expectedResult, ex.GetType().Name);
-            }
+            var ex = Assert.Throws<FormatException>(() => ConvertHex.FromHexString("Hello"));
+
+            Assert.Equal("FormatException", ex.GetType().Name);
```
jshergal, line 25: The generic Assert.Throws<T> already verifies that the thrown exception is of type <T>, so this line can be removed. We also don't need to store the return of throws in a local variable.

strings-csharp/HexToByteArray/Tests/Test.cs
```diff
@@ -4,85 +4,66 @@ namespace Tests
...
-            var expectedResult = "ArgumentNullException";
-            try
-            {
-                var result = _convertHex.FromHexString("");
-            }
-            catch (Exception ex)
-            {
-                Assert.Equal(expectedResult, ex.GetType().Name);
-            }
+            string? nullHexString = null;
```
jshergal, line 31: We can just pass null in to the method to check behavior in a null situation. Another note in relation to that is that with nullable enabled, the compiler is checking that stuff and using the annotations to surface that back to the caller of the method, so the method signature itself is conveying that null is not an acceptable value. When that is the case, we can skip checking the behavior around a null argument, as our API is saying "Null is not acceptable", which means results of passing in null are by definition "undefined".

strings-csharp/HexToByteArray/Tests/Test.cs
```diff
@@ -4,85 +4,66 @@ namespace Tests
...
+            var ex = Assert.Throws<ArgumentNullException>(() => ConvertHex.FromHexString(nullHexString));
+
+            Assert.Equal("ArgumentNullException", ex.GetType().Name);
```
jshergal, line 34: Same as previous, we don't need to check the type as the Throws<T> already did.
Related to that, just for learning experience. A cleaner way to write a test that needs to verify a certain type would be to use the Assert.IsType<T>() assertion.

strings-csharp/HexToByteArray/Tests/Test.cs
```diff
@@ -4,85 +4,66 @@ namespace Tests
...
+        }
+
+        [Fact]
+        public void WhenEmptyHexPassedPassed_ThenConverionFail()
+        {
+            var result = ConvertHex.FromHexString(string.Empty);
+
+            Assert.Equal(_expectedEmptyResult, result);
```
jshergal, line 42: Here we can just use Assert.Empty

strings-csharp/HexToByteArray/Tests/Test.cs
```diff
@@ -4,85 +4,66 @@ namespace Tests
...
-            {
-                var result = _convertHex.FromHexStringAlternative("Hello");
-            }
-            catch (Exception ex)
-            {
-                Assert.Equal(expectedResult, ex.GetType().Name);
-            }
+            var ex = Assert.Throws<FormatException>(() => ConvertHex.FromHexStringAlternative("Hello"));
+
+            Assert.Equal("FormatException", ex.GetType().Name);
```
jshergal, line 58: See earlier comments

strings-csharp/HexToByteArray/Tests/Test.cs
```diff
@@ -4,85 +4,66 @@ namespace Tests
...
-            {
-                var result = _convertHex.FromHexStringAlternative("");
-            }
-            catch (Exception ex)
-            {
-                Assert.Equal(expectedResult, ex.GetType().Name);
-            }
+            var result = ConvertHex.FromHexStringAlternative(string.Empty);
+
+            Assert.Equal(_expectedEmptyResult, result);
```
jshergal, line 66: We can use Assert.Empty

strings-csharp/HexToByteArray/Tests/Test.cs
pradeepradyumna: Done. Thanks!

strings-csharp/HexToByteArray/Tests/Test.cs
pradeepradyumna: Makes sense. I have removed this test case. Thanks!

strings-csharp/HexToByteArray/Tests/Test.cs
pradeepradyumna: Done. Thanks!

strings-csharp/HexToByteArray/Tests/Test.cs
pradeepradyumna: Done. Thanks!

strings-csharp/HexToByteArray/Tests/Test.cs
pradeepradyumna: Done. Thanks!

strings-csharp/HexToByteArray/HexToByteArray/ConvertHex.cs
```diff
@@ -1,13 +1,15 @@
-﻿namespace HexToByteArray
+﻿using System.Runtime.CompilerServices;
+
```
jshergal, line 2: Putting this comment here as it applies across each method.

- The signature can be made consistent across all of them so that each method takes a `ReadOnlySpan<char>`
- Each one can start with the same defensive block:
`if (source.StartsWith("0x")) source = source[2..];`
`if (source.IsEmpty) return Array.Empty<byte>();`
`if (source.Length % 2 != 0) throw new ArgumentException("Invalid hex string", nameof(source));`

strings-csharp/HexToByteArray/HexToByteArray/ConvertHex.cs
```diff
@@ -16,5 +18,157 @@ public static byte[] FromHexStringAlternative(ReadOnlySpan<char> hex)
                     System.Globalization.NumberStyles.HexNumber);
             return bytes;
         }
+
+        public static byte[] FromHexWithPointers(string hex)
```
jshergal, line 22: This method is not using pointers, it is performing a modular arithmetic calculation. So it would be better to rename it to something like FromHexWithModularArithmetic, or something along those lines. Note, this is one of the ones I mentioned in the review document and have in my PR with the calculation cleaned up and a method extracted to make the code a little more readable (if you notice, the calculation for both firstChar and secondChar are identical). You could just copy the method directly from my PR.

strings-csharp/HexToByteArray/HexToByteArray/ConvertHex.cs
```diff
@@ -16,5 +18,157 @@ public static byte[] FromHexStringAlternative(ReadOnlySpan<char> hex)
...
+        {
+            if (hex.Length % 2 != 0) throw new ArgumentException("Invalid hex string", nameof(hex));
+
+            var res = new byte[hex.Length / 2]; // Allocate memory
```
jshergal, line 26: In the case of these examples, we are always going to set each value of the array, so it is more performant to use [GC.AllocateUninitializedArray<T>](https://learn.microsoft.com/en-us/dotnet/api/system.gc.allocateuninitializedarray). This creates an unarray but it does not perform the step to "zero" it, thus saving a few instructions. Also, instead of doing the division we can use a left shift of 1 to divide the number in half. This generates a shift instruction instead of a division instruction.

jshergal:
In the test class, you are only testing with uppercase Hex, but the methods also support lowercase. There are also separate methods for testing with empty, etc. If you take a look at the test cases in my PR, you will see instead of using `Fact` I am using `Theory` and passing in several different test cases. Short, long, lowercase, empty, etc. This allows us to write a single test method that is executed multiple times testing different input data.
We still need to test the cases where exceptions are expected and so those will remain as separated test methods, but we can simplify the code and enhance the tests in one fell swoop by making use of the `[Theory]` attribute.

strings-csharp/HexToByteArray/HexToByteArray/ConvertHex.cs
jshergal: Just as a follow-up on the name change to this method. In C# the idea of a pointer is very specific, and their use requires an unsafe code block. We can do something similar without unsafe in more modern C# using spans, references and the Unsafe methods, but even then, we should not use the word pointer as a pointer and a reference are different things. I tried writing this method using all three options and found that there was almost no performance difference since most of the time in this method is spent in the calculation and not in the retrieving of data from the source array and copying to the destination array.

PR #1175, opened 2023-06-11
CM-758: Test Data Generation with AutoFixture in .NET

Banovvv:


dotnet-client-libraries/TestDataGenerationWithAutoFixture/TestDataGenerationWithAutoFixture/Department.cs
```diff
@@ -0,0 +1,45 @@
...
+        }
+
+        public decimal CalculateAverageSalary()
+        {
+            if (!Employees.Any())
+            {
+                return 0;
+            }
+
+            decimal sumOfSalaries = Employees.Sum(e => e.Salary);
```
jshergal, line 40: You could use the LINQ Average extension here.

dotnet-client-libraries/TestDataGenerationWithAutoFixture/TestDataGenerationWithAutoFixture/Employee.cs
```diff
@@ -0,0 +1,32 @@
...
+    public class Employee
+    {
+        public string FirstName { get; set; }
+        public string? LastName { get; set; }
+        public int? Age { get; set; }
+        public decimal Salary { get; set; }
+
+        public Employee
+            (string firstName,
+            decimal salary)
```
jshergal, line 12: This constructor should just call into the other constructor (Constructor Chaining)

dotnet-client-libraries/TestDataGenerationWithAutoFixture/TestDataGenerationWithAutoFixture/TestDataGenerationWithAutoFixture.csproj
```diff
@@ -0,0 +1,10 @@
+<Project Sdk="Microsoft.NET.Sdk">
+
+  <PropertyGroup>
+    <OutputType>Exe</OutputType>
```
jshergal, line 4: Since the console app is doing nothing but printing "Hello World", would it be better to make this just a simple library by changing this line?

Banovvv: Done, good point.

PR #1179, opened 2023-06-13
CM-405: How to Efficiently Randomize an Array in C#

Banovvv:


collections-arrays/HowToEfficientlyRandomizeAnArray/HowToEfficientlyRandomizeAnArray/ArrayFunctions.cs
```diff
@@ -0,0 +1,42 @@
+﻿namespace HowToEfficientlyRandomizeAnArray
+{
+    public class ArrayFunctions
+    {
+        private readonly Random _rnd;
+
+        public ArrayFunctions()
```
jshergal, line 7: No need for the constructor, you can initialize _rnd directly at the declaration

collections-arrays/HowToEfficientlyRandomizeAnArray/HowToEfficientlyRandomizeAnArray/ArrayFunctions.cs
```diff
@@ -0,0 +1,42 @@
...
+        {
+            _rnd = new Random();
+        }
+
+        public int[] GerOrderedArray(int numberOfElements)
+        {
+            int[] array = new int[numberOfElements];
```
jshergal, line 14: You can just use var here

collections-arrays/HowToEfficientlyRandomizeAnArray/HowToEfficientlyRandomizeAnArray/ArrayBenchmarks.cs
```diff
@@ -0,0 +1,43 @@
...
+        [GlobalSetup]
+        public void Setup()
+        {
+            _arrayFunctions = new ArrayFunctions();
+            _guidArray = _arrayFunctions.GerOrderedArray(ARRAY_SIZE);
+            _randomArray = _arrayFunctions.GerOrderedArray(ARRAY_SIZE);
+            _fisherYatesArray = _arrayFunctions.GerOrderedArray(ARRAY_SIZE);
+        }
+
+        [Benchmark]
```
jshergal, line 25: Note that our benchmarks here are not an apples to oranges comparison and are showing a skewed result both in performance and memory allocation. With the Guid and Random methods we are using LINQ and returning a copy of the original array. In the FisherYates we are modifying the array in place. To make the comparisons fair and to make our helper library fair and consistent, we should modify the FisherYates method to first make a copy of the array and then operate on the copy and return the copy.
We could add a second FisherYates method also that we call InPlace, thus making it clear that it will operate on the array directly.
Something else that might be helpful to point out in the article would be some different use cases, as in some cases using the LINQ method and operating on the IEnumerable may be preferable so that we don't 1. Mutate our original array, and 2. We don't have to pay the memory cost of copying the array. Maybe we are just trying to grab the first 10 randomly ordered items from a collection for a test, or for a card game implementation of some kind (although then we probably are looking at Cryptographic random generators, but I digress).

collections-arrays/HowToEfficientlyRandomizeAnArray/HowToEfficientlyRandomizeAnArray/Program.cs
```diff
@@ -0,0 +1,12 @@
+﻿using BenchmarkDotNet.Running;
```
jshergal, line 1: Since this is targetting .NET 7.0 we can drastically simplify this by making use of top level statements.

Banovvv: Moved to top-level statements.

collections-arrays/HowToEfficientlyRandomizeAnArray/HowToEfficientlyRandomizeAnArray/ArrayFunctions.cs
Banovvv: Hi Jeff, I'm not really familiar with `Random.Shared` but very good pointer! What would you say would be the best approach:
1. Declare `private static Random _rnd = Random.Shared` and then use it
2. Just call `Random.Shared` directly where needed - `array.OrderByDescending(x => Random.Shared.Next()).ToArray();`

collections-arrays/HowToEfficientlyRandomizeAnArray/HowToEfficientlyRandomizeAnArray/ArrayBenchmarks.cs
jshergal: That is exactly what I would expect in terms of performance, the FisherYates algorithm is very efficient for shuffling, what I was focused more on is in the memory usage. You can still point out the efficiency of the algorithm, but in the article it would be good to point out that it is operating in place and thus mutating the array that is passed in. That could be made clearer in the message signature by making the method void and making it a ref parameter, which would force the caller to add a ref and thus think about the fact that this is modifying the array. That is probably outside the scope of the article, but for a benchmark, it is always good to, as much as possible, make the comparisons equal. If one method returns a copy, then the other should return a copy, otherwise the benchmark is skewed/biased in one direction as there is always going to be more memory usage for the copy and there will of course be the time necessary to do the copy.

collections-arrays/HowToEfficientlyRandomizeAnArray/HowToEfficientlyRandomizeAnArray/ArrayFunctions.cs
jshergal: I would just call it directly, that way it is clear and you can simplify the code by removing the member variable. It doesn't really increase the character count too much in the other locations. This is a good article talking about Random and the Random.Shared introduced in .NET 6.0: https://andrewlock.net/building-a-thread-safe-random-implementation-for-dotnet-framework/

Banovvv: Thanks, this is very helpful!

collections-arrays/HowToEfficientlyRandomizeAnArray/HowToEfficientlyRandomizeAnArray/ArrayBenchmarks.cs
Banovvv: Great, I've implemented the method that copies the array as well. Thanks again!

PR #1182, opened 2023-06-13
CM-726 Testing Using WireMock.NET

fl0tschi:
Quick fixes for the article about using WireMock.NET for integration testing.

Includes switch to record class to make test slimmer and remove a redundant null check in the constructor of PlanetsService.

### DIFF
--- a/dotnet-client-libraries/WireMock.NET/WireMockNet/Planet.cs
+++ b/dotnet-client-libraries/WireMock.NET/WireMockNet/Planet.cs
@@ -1,11 +1,4 @@
 ﻿namespace WireMockNet
 {
-    public record class Planet
-    {
-        public int Id { get; set; }
-        public string Name { get; set; }
-        public double Diameter { get; set; }
-        public int NumberOfMoons { get; set; }
-        public bool HasAtmosphere { get; set; }
-    }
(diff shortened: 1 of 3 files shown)


dotnet-client-libraries/WireMock.NET/WireMockNet/Planet.cs
```diff
@@ -1,25 +1,11 @@
 ﻿namespace WireMockNet
 {
-    public class Planet
+    public record class Planet
     {
         public int Id { get; set; }
```
jshergal, line 5: This should either be {get; init;} or be switched to the primary constructor type syntax. Records are a special type of readonly and are not supposed to support set on members. There is the special with operator syntax for constructing a copy with modified values. I guess I have never tested, but I am surprised the compiler didn't give a warning on this.

jshergal: I just checked the docs, and sure enough, you can create modifiable properties on a record, but by default they are init only. Thanks for making me check the docs 😀

fl0tschi: I like the primary constructor apparently will switch to this. Thanks for the advices

PR #1184, opened 2023-06-15
[fix] : Base path route net 6

UdhayarajanJ:
Actually base path route mapped successfully but swagger we try access not found resource. I just add bit of code here.


jshergal:
You have downgraded the .NET target for this PR and so we cannot accept it. Also, in doing some reading through ASP.NET Core documentation and issues, there is a [specific issue](https://github.com/dotnet/AspNetCore.Docs/pull/25769) related to .NET 6.0 which necessitated a call to UseRouting(), and I believe that is ultimately what led to your issue. The issue was resolved in .NET 7.0. Could you please try running the example code in .NET 7.0 and see if this resolves the issue?

CodeMazeBlog:
The fix was added, it maps the route correctly now.

PR #1185, opened 2023-06-16
CM-761: Const vs readonly

SteliosPantelopoulos:
Created the code for the topic Const vs readonly


csharp-intermediate-topics/ConstVsReadonlyInCSharp/Tests/CircleCalculatorUnitTests.cs
```diff
@@ -0,0 +1,25 @@
...
+    public class CircleCalculatorUnitTests
+    {
+        private CircleCalculator _circleCalculator = new CircleCalculator();
+
+        [Fact]
+        public void WhenCalculatingCircumferenceFor3_ThenReturn18_84()
+        {
+            var result = _circleCalculator.GetCircumference(3);
+
+            Assert.Equal(18.84, result);
```
jshergal, line 14: When dealing with "Equal" and floating point types, it is better to compare using a tolerance. Especially with .NET now being cross platform, depending on the processor where this code is running, it would be possible for there to be a slight variation within the result. This is a good discussion on the topic:
https://roundwide.com/equality-comparison-of-floating-point-numbers-in-csharp/

csharp-intermediate-topics/ConstVsReadonlyInCSharp/Tests/CircleCalculatorUnitTests.cs
```diff
@@ -0,0 +1,25 @@
...
+        }
+
+        [Fact]
+        public void WhenCalculatingAccurateCircumferenceFor3_ThenReturn18_849555921540002()
+        {
+            var result = _circleCalculator.GetAccurateCircumference(3);
+
+            Assert.Equal(18.849555921540002, result);
```
jshergal, line 22: See earlier comment about equality of floating points

csharp-intermediate-topics/ConstVsReadonlyInCSharp/ConstVsReadonlyInCSharp/CircleCalculator.cs
```diff
@@ -0,0 +1,19 @@
+﻿namespace ConstVsReadonlyInCSharp
+{
+    public class CircleCalculator
+    {
+        public const double E = 2.71;
```
jshergal, line 5: This is never used nor called out in the article. Maybe it would be better to just remove it.

csharp-intermediate-topics/ConstVsReadonlyInCSharp/ConstVsReadonlyInCSharp/MathCalculator.cs
```diff
@@ -0,0 +1,25 @@
+﻿namespace ConstVsReadonlyInCSharp
+{
+    public class MathCalculator
+    {
+        public double Add(double a, double b)
+        {
+            return a + b;
```
jshergal, line 7: Since all of these methods are one line and very simple, the code would be a lot shorter if it used expression body statements.

csharp-intermediate-topics/ConstVsReadonlyInCSharp/ConstVsReadonlyInCSharp/MathCalculator.cs
```diff
@@ -0,0 +1,25 @@
+﻿namespace ConstVsReadonlyInCSharp
+{
+    public class MathCalculator
```
jshergal, line 3: There is no instance data and none of the methods rely on instance data. This class and it's methods should be static

csharp-intermediate-topics/ConstVsReadonlyInCSharp/ConstVsReadonlyInCSharp/TaxCalculator.cs
```diff
@@ -0,0 +1,25 @@
+﻿namespace ConstVsReadonlyInCSharp
+{
+    public class TaxCalculator
+    {
+        private readonly double _countryVAT;
+        private readonly double _euroToDollarConversionRate = 1.08;
+        private readonly MathCalculator _calculator = new MathCalculator();
```
jshergal, line 7: By making the `MathCalculator` class static, this line can be removed and you can just use the static methods where necessary in the code

csharp-intermediate-topics/ConstVsReadonlyInCSharp/ConstVsReadonlyInCSharp/TaxCalculator.cs
```diff
@@ -0,0 +1,25 @@
+﻿namespace ConstVsReadonlyInCSharp
+{
+    public class TaxCalculator
```
jshergal, line 3: As a general rule in C#, whenever we are dealing with currency, we should use the decimal type rather than the double type due to floating point errors. I know this is a simple example, but whenever possible we should reflect industry standards and practices

csharp-intermediate-topics/ConstVsReadonlyInCSharp/Tests/MathCalculatorUnitTests.cs
```diff
@@ -0,0 +1,41 @@
...
+    public class MathCalculatorUnitTests
+    {
+        private MathCalculator mathCalculator = new MathCalculator();
+
+        [Fact]
+        public void WhenAdding1And3_TheReturn4()
+        {
+            var result = mathCalculator.Add(1, 3);
+
+            Assert.Equal(4, result);
```
jshergal, line 14: Since the result is a double, it would be better to list the expected as `4.0` to make it clear, when reading the code, that the expected result is also a double.

csharp-intermediate-topics/ConstVsReadonlyInCSharp/Tests/TaxCalculatorUnitTests.cs
```diff
@@ -0,0 +1,25 @@
...
+    public class TaxCalculatorUnitTests
+    {
+        private TaxCalculator _taxCalculator = new TaxCalculator(0.20);
+
+        [Fact]
+        public void WhenCalculatingTaxForValue15InEuros_ThenReturn3()
+        {
+            var result = _taxCalculator.CalculateCountryVatInEuro(15);
+
+            Assert.Equal(3, result);
```
jshergal, line 14: Floating point equality compare should use a tolerance

csharp-intermediate-topics/ConstVsReadonlyInCSharp/ConstVsReadonlyInCSharp/CircleCalculator.cs
SteliosPantelopoulos: I use it in my article as an example of how to declare a public const

csharp-intermediate-topics/ConstVsReadonlyInCSharp/ConstVsReadonlyInCSharp/MathCalculator.cs
SteliosPantelopoulos: I use it as an example of how to assign a complex type to a readonly field

csharp-intermediate-topics/ConstVsReadonlyInCSharp/ConstVsReadonlyInCSharp/TaxCalculator.cs
SteliosPantelopoulos: I use it as an example of how to assign a complex type to a readonly field

csharp-intermediate-topics/ConstVsReadonlyInCSharp/ConstVsReadonlyInCSharp/CircleCalculator.cs
jshergal: It is in the code listing and the previous paragraph mentions the idea, but as it is not being explicitly called out after the code listing and not being used anywhere, it seems like noise. A better location for it, would be in the MathCalculation class and then could be used. You could just have a method in the program app that performs a calculation with E to show how it is being used. It isn't related to circle calculations (or at least the ones being demonstrated). Even within the article, which I read, I didn't really see E being referred to and when looking at the code in the article I was trying to figure out what it was doing and how it related to a circle. Just my two cents, your editor will make the final call.

csharp-intermediate-topics/ConstVsReadonlyInCSharp/ConstVsReadonlyInCSharp/MathCalculator.cs
jshergal: Missed that use case. My bad on that. I was looking at it trying to figure out why it wasn't static.

csharp-intermediate-topics/ConstVsReadonlyInCSharp/ConstVsReadonlyInCSharp/TaxCalculator.cs
jshergal: Missed that being the use case.

csharp-intermediate-topics/ConstVsReadonlyInCSharp/Tests/CircleCalculatorUnitTests.cs
```diff
@@ -0,0 +1,25 @@
...
+
+            Assert.Equal(18.84, result, 3);
+        }
+
+        [Fact]
+        public void WhenCalculatingAccurateCircumferenceFor3_ThenReturn18_849555921540002()
+        {
+            var result = _circleCalculator.GetAccurateCircumference(3);
+
+            Assert.Equal(18.849555921540002, result, 12);
```
jshergal, line 22: See previous comment about floating point comparison

csharp-intermediate-topics/ConstVsReadonlyInCSharp/Tests/MathCalculatorUnitTests.cs
```diff
@@ -0,0 +1,41 @@
...
+    public class MathCalculatorUnitTests
+    {
+        private MathCalculator mathCalculator = new MathCalculator();
+
+        [Fact]
+        public void WhenAdding1And3_TheReturn4()
+        {
+            var result = mathCalculator.Add(1, 3);
+
+            Assert.Equal((decimal)4.0, result);
```
jshergal, line 14: See previous comments about floating point precision and equality, although decimal is precise to 28 decimal places, it still performs rounding and so can still produce incorrect equality results. Also, no need for the cast, this can be written as a decimal literal with M

csharp-intermediate-topics/ConstVsReadonlyInCSharp/Tests/MathCalculatorUnitTests.cs
```diff
@@ -0,0 +1,41 @@
...
+        }
+
+        [Fact]
+        public void WhenSubtracting5And3_TheReturn2()
+        {
+            var result = mathCalculator.Subtract(5, 3);
+
+            Assert.Equal((decimal)2.0, result);
```
jshergal, line 22: See previous comment about decimal literal and floating point equality

csharp-intermediate-topics/ConstVsReadonlyInCSharp/Tests/MathCalculatorUnitTests.cs
```diff
@@ -0,0 +1,41 @@
...
+        }
+
+        [Fact]
+        public void WhenMultiplying10And3_TheReturn30()
+        {
+            var result = mathCalculator.Multiply(10, 3);
+
+            Assert.Equal((decimal)30.0, result);
```
jshergal, line 30: See previous comments

csharp-intermediate-topics/ConstVsReadonlyInCSharp/Tests/MathCalculatorUnitTests.cs
```diff
@@ -0,0 +1,41 @@
...
+        }
+
+        [Fact]
+        public void WhenDividing10And5_TheReturn2()
+        {
+            var result = mathCalculator.Divide(10, 5);
+
+            Assert.Equal((decimal)2.0, result);
```
jshergal, line 38: See previous comments

csharp-intermediate-topics/ConstVsReadonlyInCSharp/Tests/TaxCalculatorUnitTests.cs
```diff
@@ -0,0 +1,25 @@
...
+
+            Assert.Equal(3, result, 5);
+        }
+
+        [Fact]
+        public void WhenCalculatingTaxForValue15InDollars_ThenReturn3_24()
+        {
+            var result = _taxCalculator.CalculateCountryVatInDollars(15);
+
+            Assert.Equal((decimal)3.24, result, 5);
```
jshergal, line 22: decimal literal can be used

SteliosPantelopoulos:
retest this please

PR #1194, opened 2023-06-20
CM-668: Should We Use Record or Class or Struct in C#

alvaromon:
PR for https://content.code-maze.com/browse/CM-668


blazor-features/ClassStructOrRecordInCSharp/ApiData.cs
```diff
@@ -0,0 +1,15 @@
...
+using System.Threading.Tasks;
+
+namespace ClassStructOrRecordInCSharp
+{
+    public record ApiData
+    {
+        public int Id { get; set; }
+    }
+
+    public record OtherApiData(string id);
```
jshergal, line 14: The naming convention for primary arguments on records is to use PascalCase as these are translated into init only properties.

blazor-features/ClassStructOrRecordInCSharp/GeoLocation.cs
```diff
@@ -0,0 +1,17 @@
...
+using System.Runtime.CompilerServices;
+using System.Text;
+using System.Threading.Tasks;
+
+namespace ClassStructOrRecordInCSharp
+{
+    public struct GeoLocation
+    {
+        public string latitude;
+        public string longitude;
```
jshergal, line 13: Naming convention for public fields is Pascal Case

PR #1201, opened 2023-06-26
CM-775 Determining if a string consists only of numeric characters

GergoVandor:


jshergal:
This PR is missing unit tests. From the [CodeMaze authors guide](https://docs.google.com/document/d/1PRj3IPPPNTF-NU-XGQqthflVd3m8tkX7c2JGn4gfX74/)

> - your code must include tests to verify the behavior added
> - all the tests should be inside the “Tests” project
> - all tests should follow our standard naming convention: 
> FooUnitTest.cs // unit tests
> FooIntegrationTest.cs or FooIntTest.cs // integration tests
> FooLiveTest.cs // tests that require a running component

strings-csharp/DeterminingIfStringIsNumericOnly/DeterminingIfStringIsNumericOnly/IsStringNumber.cs
```diff
@@ -0,0 +1,49 @@
...
+    private static partial Regex IsDigitRegex();
+    
+    [Benchmark]
+    public bool IsNumericWithNewRegex() => new Regex("^\\d+$").IsMatch("123456789123456789");
+
+    [Benchmark]
+    public bool IsNumericWithCompiledRegex() => IsDigitRegex().IsMatch("123456789123456789");
+
+    [Benchmark]
+    public bool IsNumericWithTryParse() => int.TryParse("123456789123456789", out _);
```
jshergal, line 18: Note that this method will return false when value is greater than int.MaxValue. For example, this string, which is clearly an integer and contains only digits, will return false "21474836471"

strings-csharp/DeterminingIfStringIsNumericOnly/DeterminingIfStringIsNumericOnly/IsStringNumber.cs
```diff
@@ -0,0 +1,49 @@
+using System.Text.RegularExpressions;
+using BenchmarkDotNet.Attributes;
+
+namespace DeterminingIfStringIsNumericOnly;
+
```
jshergal, line 5: It would be nice to add the summaryorderer here so that the benchmarks are ordered either slowest to fastest or fastest to slowest

jshergal (changes requested):
There are some changes that should be made in the code, and as noted in another comment. The solution is missing unit tests. Related to one of the code issues is an issue in the article, namely this line which returns false due to the fact that the string is > int.MaxValue:
`public bool IsNumericWithTryParse() => int.TryParse("123456789123456789", out _);`

strings-csharp/DeterminingIfStringIsNumericOnly/DeterminingIfStringIsNumericOnly/IsStringNumber.cs
GergoVandor: You are right, thanks for pointing it out, I added a remark about this to the article and written unit tests to verify this

strings-csharp/DeterminingIfStringIsNumericOnly/DeterminingIfStringIsNumericOnly/IsStringNumber.cs
```diff
@@ -0,0 +1,49 @@
...
+public partial class IsStringNumber
```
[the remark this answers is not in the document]
GergoVandor: Good idea thanks, done

strings-csharp/DeterminingIfStringIsNumericOnly/DeterminingIfStringIsNumericOnly/IsStringNumber.cs
GergoVandor: Added

strings-csharp/DeterminingIfStringIsNumericOnly/DeterminingIfStringIsNumericOnly/IsStringNumber.cs
```diff
@@ -0,0 +1,60 @@
...
+    [Arguments(IntegerString)]
+    public bool IsNumericWithNewRegex(string input) => new Regex("^\\d+$").IsMatch(input);
+
+    [Benchmark]
+    [Arguments(IntegerString)]
+    public bool IsNumericWithCompiledRegex(string input) => IsDigitRegex().IsMatch(input);
+
+    [Benchmark]
+    [Arguments(IntegerString)]
+    public bool IsNumericWithTryParse(string input) => int.TryParse(input, out _);
```
jshergal, line 25: You should either remove this method from the benchmark, or change the IntegerString to a value that is <= int.MaxValue, otherwise you have an invalid benchmark as all of the other methods are benchmarking the "happy" case, while this is benchmarking the failure case.

strings-csharp/DeterminingIfStringIsNumericOnly/Tests/IsStringNumberUnitTests.cs
```diff
@@ -0,0 +1,171 @@
...
+
+namespace Tests;
+
+public class IsStringNumberUnitTests
+{
+    [Test]
+    public void GivenIsNumericWithNewRegex_WhenInputIsIntegerOnly_ThenReturnsTrue()
+    {
+        const string input = "1234567890";
+        var isStringNumber = new IsStringNumber();
```
jshergal, line 11: Since this is being done in each test, we could make it a private static readonly member of the class and just reference it from there (I believe this works in NUnit, I am not as familiar with the current state of NUnit - if that is not the case, then we can still add a member variable and use the OneTimeSetup attribute in conjunction with TestFixture to handle the initialization)

strings-csharp/DeterminingIfStringIsNumericOnly/Tests/IsStringNumberUnitTests.cs
```diff
@@ -0,0 +1,171 @@
+using DeterminingIfStringIsNumericOnly;
+
+namespace Tests;
+
+public class IsStringNumberUnitTests
```
jshergal, line 5: Since with all but one exception (at least from a quick perusal) there are two constants being used for the test cases: a true case and a false case. These constants could be elevated to the class level and referenced there.
If we wanted to reduce the amount of code in the test class, we could go a step further and make use of the TestCaseSource attribute and add a simple static property that yielded the two strings along with the expected boolean outcome. Then we can just add two parameters to each of the tests (with the exception of the large integer for the int.TryParse case): string input and bool expectedResult

strings-csharp/DeterminingIfStringIsNumericOnly/DeterminingIfStringIsNumericOnly/IsStringNumber.cs
GergoVandor: I changed the string to be less than int.MaxValue

strings-csharp/DeterminingIfStringIsNumericOnly/Tests/IsStringNumberUnitTests.cs
GergoVandor: Extracted them, it's just a habit of mine that I declare everything inside the test (or in an instance method and call it from every test), this way unit tests can run in parallel and I don't have to think about whether the system under test is thread-safe or not. However here it shouldn't be an issue since there is no state in the IsStringNumber class, so I made these static properties

strings-csharp/DeterminingIfStringIsNumericOnly/Tests/IsStringNumberUnitTests.cs
GergoVandor: Added TestCase attributes to tests using the static properties (except for the TryParse when the input is larger than int.MaxValue, there I just passed a long string to the TestCase attribute)

PR #1223, opened 2023-07-04
CM-49: Locking Mechanism in C#

bozo-spoljaric:


csharp-intermediate-topics/LockingMechanismInCSharp/LockingMechanismInCSharp/DeadLock.cs
```diff
@@ -0,0 +1,55 @@
...
+
+        private static void IncrementCounter1()
+        {
+            lock (_firstLockObject)
+            {
+                Thread.Sleep(1000);
+
+                lock (_secondLockObject)
+                {
+                    for (int i = 1; i <= 100000; i++)
```
jshergal, line 31: This comment applies across all of the for loops in the code. The more standard approach for a loop that is counting is to start from 0 and check `<` as opposed to starting at 1 and performing `<=`. There are cases where you may need to start at 1 or some other index, but in the standard "I want to do something 100 times", industry practice is to start at 0. Since that is a standard, it also makes it a half step quicker for someone stepping into the codebase to parse what is going on with the loop. Seeing it start at 1 gives an immediate pause to try to think why it isn't starting at 0.

csharp-intermediate-topics/LockingMechanismInCSharp/Tests/LockingMechanismInCSharpTests.cs
```diff
@@ -0,0 +1,64 @@
...
+{
+    [TestClass]
+    public class LockingMechanismInCSharpTests
+    {
+        [TestMethod]
+        public void WhenMultithreadWithoutLock_ThenDifferentResultSmallerThan3000()
+        {
+            var result = MultithreadWithoutLocking.Execute();
+
+            Assert.IsTrue(result < 300000, "Result is not smaller than 300000");
```
jshergal, line 13: I have an AMD Processor with 8 cores. Running the `Execute` in a loop of 10,000 iterations I found that the correct 300,000 result occurred 9728 times. I am not exactly sure what a better test would be. It's a hard thing to test since the failure is non-deterministic and is very hardware-dependent. Maybe running it in a loop (with a minor change in the Execute method to reset the counter to 0) until it produces the wrong result. Sorry, I don't have a better suggestion offhand, but this test as it is, will non-deterministicly cause the tests to fail.

csharp-intermediate-topics/LockingMechanismInCSharp/LockingMechanismInCSharp/DeadLock.cs
bozo-spoljaric: fixed, also changed int to var

csharp-intermediate-topics/LockingMechanismInCSharp/Tests/LockingMechanismInCSharpTests.cs
bozo-spoljaric: Unfortunately, I couldn't find a meaningful way to write this test, so I just removed it.

PR #1230, opened 2023-07-06
CM-250: Sorting Generic List in C#

jshergal:
Initial commit of code for CM250


collections-lists/SortingGenericList/src/SortingGenericList.Library/MySortedList.cs
```diff
@@ -0,0 +1,99 @@
...
+        if (index < 0) return false;
+
+        _list.RemoveAt(index);
+
+        return true;
+    }
+
+    public int Count => _list.Count;
+
+    public bool IsReadOnly => false;
```
fl0tschi, line 98: `Count` and `IsReadOnly` should be moved underneath the `Capacity` property.

fl0tschi (changes requested):
Only the two properties need to be moved, then we are good with the code itself.

For the future - you don't need to fix it for this code review - the tests should have the pattern `When..._Then...`

fl0tschi (approved):

PR #1235, opened 2023-07-08
CM-697: Entity framework Vore vs Dapper

SteliosPantelopoulos:


dotnet-efcore/EntityFrameworkCoreVsDapper/EntityFrameworkCoreVsDapper/Repositories/PersonsRepository.cs
```diff
@@ -0,0 +1,64 @@
...
+            }
+        }
+
+        [Benchmark]
+        [Arguments(10000)]
+        [Arguments(100000)]
+        [Arguments(1000000)]
+        public List<Person> GetXPersonsEFCore(int personsNumber)
+        {
+            using (_context)
```
jshergal, line 51: This using needs to be removed as we are disposing of the context in the `Cleanup` method. Also, when using this class outside of a benchmark context, we could run into use after dispose issues. As a general rule, if we are not the one creating the object, we should not be the one to dispose of it. `_context` being a member variable indicates that our class should be made IDisposable and implement the Dispose pattern.

dotnet-efcore/EntityFrameworkCoreVsDapper/EntityFrameworkCoreVsDapper/Repositories/PersonsRepository.cs
```diff
@@ -0,0 +1,64 @@
...
+            _context = new PersonsContext();
+        }
+
+        [Benchmark]
+        [Arguments(10000)]
+        [Arguments(100000)]
+        [Arguments(1000000)]
+        public List<Person> GetXPersonsDapper(int personsNumber)
+        {
+            using (_dbConnection)
```
jshergal, line 37: This using needs to be removed as the object is being cleanedup/disposed of in the `Cleanup` method. This also means that we are violating the use after dispose rule as `Cleanup` will be called after this method executes, meaning the call to `_dbConnection.Close()` is executed on a disposed object.

dotnet-efcore/EntityFrameworkCoreVsDapper/EntityFrameworkCoreVsDapper/Program.cs
```diff
@@ -1,15 +1,18 @@
...
     static void Main(string[] args)
     {
-        var efCoreRepository = new PersonsRepositoryEFCore();
-        efCoreRepository.GetXPersons(1000);
+        var personsRepository = new PersonsRepository();
+        personsRepository.GetXPersonsEFCore(1000);
 
-        var dapperRepository = new PersonsRepositoryDapper();
-        dapperRepository.GetXPersons(1000);
+        personsRepository.GetXPersonsDapper(1000);
```
jshergal, line 15: We aren't doing anything with these lines of code, so we could just remove them. Within the benchmark methods we are showing how to call the GetXPersons methods, so there isn't necessarily a need to also highlight it here.

dotnet-efcore/EntityFrameworkCoreVsDapper/EntityFrameworkCoreVsDapper/Program.cs
```diff
@@ -1,15 +1,18 @@
...
 
         BenchmarkRunner.Run(typeof(Program).Assembly);
```
jshergal, line 17: This line can be simplified to `BenchmarkRunner.Run<PersonsRepository>();`

jshergal (changes requested):

dotnet-efcore/EntityFrameworkCoreVsDapper/EntityFrameworkCoreVsDapper/Program.cs
SteliosPantelopoulos: I need these lines for my article code to show how I can call these methods. Indeed when you see all the code it's clear how to call them, but looking at the code sections of the article it's not that clear.

PR #1237, opened 2023-07-09
CM-743 Refactoring Couplers in C#

zanbor:


csharp-refactoring/RefactoringCouplers/RefactoringCouplers/FeatureEnvy/Correct/Ticket.cs
```diff
@@ -0,0 +1,33 @@
...
+
+    private bool Available { get; set; }
+    private int Price { get; init; }
+
+    public bool IsAvailable()
+    {
+        return Available;
+    }
+
+    private bool CanAfford(decimal points)
```
jshergal, line 18: Since the method is private and both `point` in the `BuyTicket` method and `Price` are of type `int`, we should probably just make this an int here as well. No need for the additional precision when we are not using it.
I think an argument could be made that `Price` could be changed to a decimal, since we are representing a currency.

zanbor: Right, my bad! I wanted to keep it simple and go with `int`s everywhere. I corrected it in both Ticket classes, as well as in the article. Thanks :)

csharp-refactoring/RefactoringCouplers/RefactoringCouplers/FeatureEnvy/Correct/Customer.cs
```diff
@@ -0,0 +1,22 @@
+﻿namespace RefactoringCouplers.FeatureEnvy.Correct;
+public class Customer
+{
+    public Customer(string name)
+    {
+        this.Name = name;
```
mahrncic, line 6: No need to use the `this` keyword since there is a difference between the property and the parameter (the first letter is capitalized for the property). You also don't use it in your other classes, so it would be better if we are consistent 😄

csharp-refactoring/RefactoringCouplers/RefactoringCouplers/FeatureEnvy/Incorrect/Customer.cs
```diff
@@ -0,0 +1,25 @@
+﻿namespace RefactoringCouplers.FeatureEnvy.Incorrect;
+public class Customer
+{
+    public Customer(string name)
+    {
+        this.Name = name;
```
mahrncic, line 6: Same as for previous `Customer` class, let's remove `this` keyword.

csharp-refactoring/RefactoringCouplers/RefactoringCouplers/FeatureEnvy/Incorrect/Ticket.cs
```diff
@@ -0,0 +1,27 @@
...
+    public Ticket(bool available, int price)
+    {
+        Available = available;
+        Price = price;
+    }
+
+    private bool Available { get; set; }
+    public int Price { get; init; }
+
+    public bool IsAvailable()
```
mahrncic, line 13: No need for this method I think

mahrncic (changes requested):
Looks good, I like your examples, proposed a couple of changes 😄

csharp-refactoring/RefactoringCouplers/RefactoringCouplers/FeatureEnvy/Correct/Ticket.cs
```diff
@@ -0,0 +1,33 @@
...
+        Price = price;
+    }
+
+    private bool Available { get; set; }
+    private int Price { get; init; }
+
+    public bool IsAvailable()
+    {
+        return Available;
+    }
```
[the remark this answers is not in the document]
zanbor: Good point, but I changed the property to `{ get; private set; }` because it is possible to change it outside the constructor (but not outside the `Ticket` class) :)

zanbor:
> Looks good, I like your examples, proposed a couple of changes 😄

Thank you, I'm glad to hear that! I updated the code according to your suggestions, thanks for all the comments :)

mahrncic (approved):

mahrncic (approved):

PR #1241, opened 2023-07-11
CM-718: Introduction to PDF Manipulation With iText

MPrtenjak:
Initial code for the article


dotnet-client-libraries/ITextBasics/ITextBasics/BigDocument.cs
```diff
@@ -0,0 +1,87 @@
...
+        private Paragraph CreatePictureParagraph(string imageFile)
+        {
+            var imageData = ImageDataFactory.Create(imageFile);
+            var image = new Image(imageData);
+
+            return new Paragraph()
+                .Add(image);
+        }
+
+        private string LoremIpsum(int numberOfWords)
```
jshergal, line 71: No need to change anything, just an FYI in case you are not aware of the package. There is a faker library called [Bogus](https://github.com/bchavez/Bogus) that includes a Lorem Ipsum generator to make paragraphs, sentences, number of words, etc.

jshergal, line 71: You could make this method static.

dotnet-client-libraries/ITextBasics/ITextBasics/BigDocument.cs
```diff
@@ -0,0 +1,87 @@
...
+        {
+            var loremIpsumText = "Lorem ipsum dolor sit amet, consectetur adipiscing elit, sed do eiusmod tempor incididunt ut labore et dolore magna aliqua. Ut enim ad minim veniam, quis nostrud exercitation ullamco laboris nisi ut aliquip ex ea commodo consequat. Duis aute irure dolor in reprehenderit in voluptate velit esse cillum dolore eu fugiat nulla pariatur. Excepteur sint occaecat cupidatat non proident, sunt in culpa qui officia deserunt mollit anim id est laborum.";
+            var words = loremIpsumText.Split(' ');
+
+            var sentences = new List<string>();
```
jshergal, line 76: This could be done with a StringBuilder and then calls to AppendJoin. We could also create the StringBuilder with a starting capacity based on the number of words. Maybe something like numberOfWords * 10 or instead of 10 we could use the average length of the Ipsum words.

dotnet-client-libraries/ITextBasics/ITextBasics/Program.cs
```diff
@@ -0,0 +1,40 @@
...
+var pdfCreator = new RealPdfCreator(console);
+var helloWorld = new HelloWorld();
+var bigDocument = new BigDocument();
+
+while (true)
+{
+    userMenu.DisplayOptions();
+    var userAction = userMenu.GetSelection();
+
+    if (userAction == UserMenu.UserAction.Exit)
```
jshergal, line 16: This could be moved inside the switch statement

dotnet-client-libraries/ITextBasics/ITextBasics/BigDocument.cs
MPrtenjak: Hi,

I Did hear about Bogus library, but I really do not use a lot of external libraries and I would never use an external library for something so simple :)... But thanks to let me know that even something like this exists 👍

MPrtenjak: Yes, that is true. I will make it static.

Any function that does not use any internal object state can be static. But should it be? If this would be "my program" I would not make it static just because otherwise I would have to go through the codebase and make any function that can be static into a static function.

dotnet-client-libraries/ITextBasics/ITextBasics/BigDocument.cs
MPrtenjak: HI,

In the light of my previous answer. If this is something I really have to do, no problem, of course, I can do it. But again! Realy? Should we do that in the code that is meant for beginners? What will anyone gain from that?

dotnet-client-libraries/ITextBasics/ITextBasics/BigDocument.cs
```diff
@@ -0,0 +1,87 @@
...
+            var image = new Image(imageData);
+
+            return new Paragraph()
+                .Add(image);
+        }
+
+        private string LoremIpsum(int numberOfWords)
+        {
+            var loremIpsumText = "Lorem ipsum dolor sit amet, consectetur adipiscing elit, sed do eiusmod tempor incididunt ut labore et dolore magna aliqua. Ut enim ad minim veniam, quis nostrud exercitation ullamco laboris nisi ut aliquip ex ea commodo consequat. Duis aute irure dolor in reprehenderit in voluptate velit esse cillum dolore eu fugiat nulla pariatur. Excepteur sint occaecat cupidatat non proident, sunt in culpa qui officia deserunt mollit anim id est laborum.";
+            var words = loremIpsumText.Split(' ');
```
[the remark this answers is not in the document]
MPrtenjak: That is true, and I did it like this, but "premature optimization is the root of evil". 

What did we gain from that? The code is now harder to follow. Variables that were in a function (hidden) are now outside in the object. Why? Is this a function that we should optimize to that extend?

dotnet-client-libraries/ITextBasics/ITextBasics/Program.cs
MPrtenjak: Ok, corrected.

But why I programmed it like that in the first place? Because 'Exit' is a special action, out of the ordinary flow, and that is why is in my code outside of other similar actions.

dotnet-client-libraries/ITextBasics/Tests/BigDocumentTest.cs
```diff
@@ -0,0 +1,18 @@
...
+    [TestClass]
+    public class BigDocumentTest
+    {
+        [TestMethod]
+        public void GivenCreateCommand_WhenRunning_ThenExpectCreatedFile()
+        {
+            var bigDocument = new BigDocument();
+            var tmpFileName = Path.GetTempFileName();
+            var runner = new PdfRunnerTest();
+            runner.DoTest(tmpFileName, bigDocument.Create);
```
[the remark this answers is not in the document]
MPrtenjak: Files in TEMP folder do not have to be deleted. I chose to do so because I like to keep everything clean... to some extend 😉.

I would not use an external library for this, I would also not develop a new class with a disposable pattern just for that... No need to complicate things and program - clean code!

So I just solve it with a basic try - finally structure.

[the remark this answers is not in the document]
jshergal: While it is true, they do not have to be deleted, it is considered good practice to perform housekeeping. We can cause problems for other programs if there are too many files in $TEMP. In older versions of .NET there were issues related to creating too many that would cause software to fail. Related you can read [this answer on SO](https://stackoverflow.com/a/27708677).
Regarding an external library or an IDisposable, those were just suggestions of ways in which the problem could be solved. However you choose to address it is fine. I wrote a class years ago and published it in a few places (first to google code, then to CodeNet or whatever the MS one was called before they bought GitHub) and finally moved it all to GitHub.  It is a great utility class and I have needed it in lots of projects where I am dealing with large temp files and don't want to leak them into the TEMP folder. That is why I threw out that idea, thinking maybe you already have a utility class like that that you keep around. I like the IDisposable exactly for the purpose of clean code. The class is called TemporaryFile and does the very best it can to delete the file on Dispose 😀

dotnet-client-libraries/ITextBasics/ITextBasics/Program.cs
jshergal: It is easier to read and cleaner to place it within the switch statement, since we are already switching over almost all of the other enum states. It makes more sense from a code readability flow to be within the switch. I would argue that the `Unknown` case should also be included explicitly for completeness. To me the "Exit" action is not special, but if you want to make it as a special case and treat it differently, then I think the while loop should become a do..while and the exit condition of the loop should be this if statement (with the opposite check of course). Then you don't need to add that case into the switch and it better captures the intent of the code, and has the benefit of removing a while(true). In general, a while(true) loop with only one exit condition is an anti-pattern, since that style of loop having only one break/exit condition can generally be rewritten as a do..while.

dotnet-client-libraries/ITextBasics/ITextBasics/BigDocument.cs
jshergal: It is a commonly used library for unit testing and you understand correctly, I was just pointing it out as a "in case you didn't know" statement, no arguments with keeping the method.

jshergal: As a best practice, methods which do not access instance data should be static. Both VS and ReSharper will point this out as a violation of code analysis rule [CA1822](https://learn.microsoft.com/en-us/dotnet/fundamentals/code-analysis/quality-rules/ca1822).

dotnet-client-libraries/ITextBasics/ITextBasics/BigDocument.cs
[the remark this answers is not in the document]
jshergal: If you feel promoting it to the class level as a constant is a problem, no worries there, it should be a local constant though. That maintains the clarity and is a best practice. Internal constant data should be marked constant. It does have performance implications, which in the context of the article and the code is irrelevant, except that in our articles we should be striving to adhere to industry best practices.

The bigger issue is in the `Split` method here which is creating small arrays on the heap each time the method is called. Again, I understand the context of a simple article and so the performance is not an issue, but our goal as authors is to present code that is clean, readable and adheres to best practices. Best practice would be a static readonly array  within this class initialized like this:
`private static readonly string[] LoremIpsumWords = "Lorem ipsum dolor sit amet, consectetur adipiscing elit, sed do eiusmod tempor incididunt ut labore et dolore magna aliqua. Ut enim ad minim veniam, quis nostrud exercitation ullamco laboris nisi ut aliquip ex ea commodo consequat. Duis aute irure dolor in reprehenderit in voluptate velit esse cillum dolore eu fugiat nulla pariatur. Excepteur sint occaecat cupidatat non proident, sunt in culpa qui officia deserunt mollit anim id est laborum.".Split(' ');`
You could move the string constant to it's own const declaration to deal with word wrapping, etc, but initializing it once would follow best practice.

dotnet-client-libraries/ITextBasics/ITextBasics/BigDocument.cs
jshergal: Following on to my answer on the previous point, our goal is to provide code that follows best practices. In this case, rewriting this code with a StringBuilder will make the code shorter, easier to follow and follow best practices.

dotnet-client-libraries/ITextBasics/ITextBasics/BigDocument.cs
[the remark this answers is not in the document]
jshergal: Here is a [fiddle](https://dotnetfiddle.net/T7ZE3X) with the method rewritten using a StringBuilder, a class constant and a static readonly array. Notice the method is shorter and I would argue easier to read.

dotnet-client-libraries/ITextBasics/ITextBasics/BigDocument.cs
MPrtenjak: I repaired all methods that can be static to static.

dotnet-client-libraries/ITextBasics/ITextBasics/BigDocument.cs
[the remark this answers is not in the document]
MPrtenjak: Thanks for the fiddle. I have implemented your code.

dotnet-client-libraries/ITextBasics/ITextBasics/BigDocument.cs
MPrtenjak: I used your code, so it is now using StringBuilder

dotnet-client-libraries/ITextBasics/ITextBasics/Program.cs
MPrtenjak: i implemented all options in the case statement.

dotnet-client-libraries/ITextBasics/Tests/UserMenuTest.cs
```diff
@@ -0,0 +1,32 @@
...
+
+namespace Tests
+{
+    [TestClass]
+    public class UserMenuTest
+    {
+        [TestMethod]
+        public void GivenKey_WhenGettingUserInput_ThenExpectCorrectAction()
+        {
+            StringBuilder screenOutput = new();
```
jshergal, line 14: Since there is no checking of what has been printed to the console, and this is the only place where `FakeConsole` is being used. It seems like it would be better to remove this variable and then update `FakeConsole` to perform noops on the write and clear calls.
If there is an intended test case to test that UserMenu is writing the proper output to the console, then we should add a test around that as well.
In that case, we could also update `FakeConsole` to create its own internal buffer and add accessor methods around that. It would make the constructor cleaner and remove a dependency on callers, who would no longer have to care about how it was internally storing things. Maybe it could be using a list of strings, maybe a stringbuilder, maybe some kind of rolling array buffer, etc. The caller no longer has to care about that, it just calls the accessor methods to see what was written.

MPrtenjak: Corrected class and add a test for a number of output lines on the console.

PR #1245, opened 2023-07-12
CM-166: XML Deserialization in C#

falgunmodispade:


jshergal:
Hi @falgunmodispade, when creating a PR, please create a draft PR when the code is not ready for review.

xml-csharp/XMLDeserializationInCsharp/XMLDeserializationInCsharp/Program.cs
```diff
@@ -0,0 +1,53 @@
...
+                             "<Books><Book>" +
+                             "<Title>The Catcher in the Rye</Title>" +
+                             "<Author>J.D. Salinger</Author>" +
+                             "</Book>" +
+                             "<Book>" +
+                             "<Title>To Kill a Mockingbird</Title>" +
+                             "<Author>Harper Lee</Author>" +
+                             "</Book>" +
+                             "</Books>" +
+                             "</Library>";
```
jshergal, line 32: We can use a raw string literal here.

xml-csharp/XMLDeserializationInCsharp/XMLDeserializationInCsharp/Program.cs
```diff
@@ -0,0 +1,53 @@
...
+                Console.WriteLine($"Title: {book.Title}, Author: {book.Author}");
+            }
+            #endregion
+        }
+
+        #region Deserialization Method
+        public static T DeserializeXmlData<T>(string xmlData)
+        {
+            var serializer = new XmlSerializer(typeof(T));
+            using (var reader = new StringReader(xmlData))
```
jshergal, line 46: We can reduce nesting here with a using statement.

xml-csharp/XMLDeserializationInCsharp/XMLDeserializationTests/XMLDeserializationTest.cs
```diff
@@ -0,0 +1,59 @@
...
+                              "<Books><Book>" +
+                              "<Title>The Catcher in the Rye</Title>" +
+                              "<Author>J.D. Salinger</Author>" +
+                              "</Book>" +
+                              "<Book>" +
+                              "<Title>To Kill a Mockingbird</Title>" +
+                              "<Author>Harper Lee</Author>" +
+                              "</Book>" +
+                              "</Books>" +
+                              "</Library>";
```
jshergal, line 41: We can use a raw string literal here

xml-csharp/XMLDeserializationInCsharp/XMLDeserializationInCsharp/Person.cs
```diff
@@ -0,0 +1,14 @@
...
+using System.Text;
+using System.Threading.Tasks;
+
+namespace XMLDeserializationInCsharp
+{
+    public class Person
+    {
+        public string Name { get; set; }
+        public int Age { get; set; }
+    }
```
jshergal, line 13: This would be a good use case example to show deserializing of records. I would recommend that we change this to a record and then we will be able to highlight how deserialization works with records. The big catch is that we need to define a parameterless constructor on the object, but we can define it privately.
`public record Person(string Name, int Age)
{
    private Person() : this("", int.MinValue) {}
}`

xml-csharp/XMLDeserializationInCsharp/XMLDeserializationInCsharp/Book.cs
```diff
@@ -0,0 +1,18 @@
...
+
+namespace XMLDeserializationInCsharp
+{
+    public class Book
+    {
+        [XmlElement("Title")]
+        public string Title { get; set; }
+
+        [XmlElement("Author")]
+        public string Author { get; set; }
```
jshergal, line 16: This would also be a good candidate showing how to use a record. Instead of doing the Person class, we could use this one and show how to assign attributes in a positional record statement.
`public record Book([property: XmlElement("Title")] string Title, ...)`

CodeMazeBlog: Hi @jshergal. Thanks for these suggestions. They are great as always. I will talk to the author to make these changes. Also, I would like to see class deserialization as well, so, we can just add a simple section for record deserialization and apply your changes.

xml-csharp/XMLDeserializationInCsharp/XMLDeserializationInCsharp/Program.cs
```diff
@@ -2,51 +2,82 @@
...
     {
         static void Main(string[] args)
         {
+            #region Deserialization_Of_Class
             #region Simple Deserialization
-            var xmlData = "<Person>" +
-                        "<Name>John Doe</Name>" +
-                        "<Age>30</Age>" +
-                        "</Person>";
+            var xmlData = @"<Person>
```
jshergal, line 11: This works, but my suggestion was to make it a raw string literal, that way it won't have a ton of extra unnecessary whitespace in the string. Raw string literals start with """. [Here is a link](https://learn.microsoft.com/en-us/dotnet/csharp/language-reference/proposals/csharp-11.0/raw-string-literal) to the proposal documentation which explains it more. It is perfect for defining json and xml in code and preserving the formatting and spacing properly. Both VS and Rider handle it properly when formatting the code.

jshergal:
The // Arrange, // Act, and // Assert comments are not needed in the unit tests.

falgunmodispade:
> The // Arrange, // Act, and // Assert comments are not needed in the unit tests.

I have removed comments too from Test File

xml-csharp/XMLDeserializationInCsharp/XMLDeserializationInCsharp/Program.cs
falgunmodispade: Hi @jshergal Thank you for the detailed answer i have changed string to raw string literal now you can check

jshergal:
The build is failing due to an issue with the unit tests. Double check the new strings as there is some unexpected spacing in elements. Secondarily, there are many warnings related to non-nullable fields. It would be good to clean those up. The options are to allow null, or to initialize to a non-null value.
Example of initializing to non-null in the LibraryRecord class:
`public List<BookRecord> Books { get; init; } = new();`

Example of allowing null:
` public List<BookRecord>? Books { get; init; }`

Also, the `public static T DeserializeXmlData<T>` in Program.cs has the same type of issue as the XmlSerializer can return null. It is explicitly defined with this signature: `public object? Deserialize(`. My recommendation would be to adjust the signature of DeserializeXmlData to `T? DeserializeXmlData<T>`

falgunmodispade:
> The build is failing due to an issue with the unit tests. Double check the new strings as there is some unexpected spacing in elements. Secondarily, there are many warnings related to non-nullable fields. It would be good to clean those up. The options are to allow null, or to initialize to a non-null value. Example of initializing to non-null in the LibraryRecord class: `public List<BookRecord> Books { get; init; } = new();`
> 
> Example of allowing null: ` public List<BookRecord>? Books { get; init; }`
> 
> Also, the `public static T DeserializeXmlData<T>` in Program.cs has the same type of issue as the XmlSerializer can return null. It is explicitly defined with this signature: `public object? Deserialize(`. My recommendation would be to adjust the signature of DeserializeXmlData to `T? DeserializeXmlData<T>`

Hi @jshergal now the build is succeded and I have applied your suggestions

PR #1248, opened 2023-07-12
CM-228 Merging Dictionaries In CSharp. First commit.

KonstantinosP7:


collections-csharp/MergingDictionariesInCSharp/MergingDictionariesInCSharp.Benchmark/MergingMethodsBenchmark.cs
```diff
@@ -0,0 +1,93 @@
...
+    private Dictionary<int, string> _dictionaryA;
+    private Dictionary<int, string> _dictionaryB;
+
+    [GlobalSetup]
+    private void Setup()
+    {
+        _dictionaryA = new Dictionary<int, string>();
+        _dictionaryB = new Dictionary<int, string>();
+
+        Random random = new Random();
```
jshergal, line 19: Since we are not using a seed, this can be changed to use [Random.Shared](https://learn.microsoft.com/en-us/dotnet/api/system.random.shared?view=net-7.0). The same applies in the GenerateRandomValue method.

collections-csharp/MergingDictionariesInCSharp/MergingDictionariesInCSharp.Tests/UnitTests.cs
```diff
@@ -0,0 +1,99 @@
...
+    public void HashSetMethod_ShouldMergeDictionaries()
+    {
+        Dictionary<int, string> mergedDictionary = UsingHashSetMethod(_dictionaryA, _dictionaryB);
+
+        Dictionary<int, string> expectedDictionary = GetExpectedDictionary();
+
+        CollectionAssert.AreEqual(expectedDictionary, mergedDictionary);
+    }
+
+    private Dictionary<int, string> GetExpectedDictionary()
```
jshergal, line 87: This is more of an opinion, but I feel like the tests would be clearer if this dictionary was also created in the SetUp method, then you see the expected alongside dictionaries a and b all within one block of code. It could then just be a member variable alongside _dictionaryA and _dictionaryB. Maybe called _expectedMergedDictionary.

collections-csharp/MergingDictionariesInCSharp/MergingDictionariesInCSharp/Methods/ConcatMethod.cs
```diff
@@ -0,0 +1,14 @@
+﻿namespace MergingDictionariesInCSharp
+{
+    public partial class Methods
+    {
+        public static Dictionary<TKey, TValue> ConcatMethod<TKey, TValue>(params Dictionary<TKey, TValue>[] dictionaries)
```
jshergal, line 5: We should be sure there is also a test showing the expected ArugmentException when the two dictionaries contain the same key. This is explicitly called out in the article that the dictionaries must contain unique keys, so it would be good to have a test confirming that as well.

collections-csharp/MergingDictionariesInCSharp/MergingDictionariesInCSharp/Methods/UnionMethod.cs
```diff
@@ -0,0 +1,15 @@
+﻿namespace MergingDictionariesInCSharp
+{
+    public partial class Methods
+    {
+        public static Dictionary<TKey, TValue> UnionMethod<TKey, TValue>(params Dictionary<TKey, TValue>[] dictionaries)
```
jshergal, line 5: We should add a test for the expected ArgumentException when the dictionaries key sets are not mutually exclusive.

collections-csharp/MergingDictionariesInCSharp/MergingDictionariesInCSharp/Methods/UsingHashSetMethod.cs
```diff
@@ -0,0 +1,25 @@
+﻿namespace MergingDictionariesInCSharp
+{
+    public partial class Methods
+    {
+        public static Dictionary<TKey, TValue> UsingHashSetMethod<TKey, TValue>(params Dictionary<TKey, TValue>[] dictionaries)
```
jshergal, line 5: This method will not work as expected. It is working with the test cases that are being thrown at it, but the basis of the method is expecting that both HashSets are storing the keys and values in the same order, but this is explicitly not guaranteed by HashSet.

collections-csharp/MergingDictionariesInCSharp/MergingDictionariesInCSharp/Program.cs
```diff
@@ -0,0 +1,30 @@
...
+Console.WriteLine("Using ForEach: ");
+PrintDictionary(ForEachMethod(dictionaryA, dictionaryB));
+Console.WriteLine("Using GroupBy: ");
+PrintDictionary(GroupByMethod(dictionaryA, dictionaryB));
+Console.WriteLine("Using Lookup: ");
+PrintDictionary(LookupMethod(dictionaryA, dictionaryB));
+Console.WriteLine("Using Union: ");
+PrintDictionary(UnionMethod(dictionaryA, dictionaryB));
+Console.WriteLine("Using Hashet: ");
+PrintDictionary(UnionMethod(dictionaryA, dictionaryB));
```
jshergal, line 30: I believe UsingHashSetMethod is intended here.

collections-csharp/MergingDictionariesInCSharp/MergingDictionariesInCSharp.Benchmark/MergingMethodsBenchmark.cs
PhilBroderick: This can also be a `var`.

collections-csharp/MergingDictionariesInCSharp/MergingDictionariesInCSharp.Benchmark/MergingMethodsBenchmark.cs
```diff
@@ -0,0 +1,93 @@
...
+
+        // Generate random values for dictionaryA
+        for (int i = 0; i < 100; i++)
+        {
+            int key = random.Next(100);
+            string value = GenerateRandomValue();
```
PhilBroderick, line 25: Likewise - `vars`.

collections-csharp/MergingDictionariesInCSharp/MergingDictionariesInCSharp.Benchmark/MergingMethodsBenchmark.cs
```diff
@@ -0,0 +1,93 @@
...
+            }
+        }
+
+        // Generate random values for dictionaryB
+        foreach (var kvp in _dictionaryA)
+        {
+            int key = random.Next(100, 200);
+            string value = GenerateRandomValue();
+
+            if (!_dictionaryB.ContainsKey(key) && !_dictionaryA.ContainsKey(key))
```
PhilBroderick, line 43: Do we need to check if the key is in Dictionary A? As we use 100 for the upper bound when seeding Dictionary A, it will not be present in that dictionary, so we can safely use it in dictionary B and not worry about checking if the key exists in Dictionary A.

collections-csharp/MergingDictionariesInCSharp/MergingDictionariesInCSharp.Benchmark/MergingMethodsBenchmark.cs
```diff
@@ -0,0 +1,93 @@
...
+            {
+                _dictionaryB.Add(key, value);
+            }
+        }
+    }
+
+    private string GenerateRandomValue()
+    {
+        Random random = new Random();
```
PhilBroderick, line 52: `var`

collections-csharp/MergingDictionariesInCSharp/MergingDictionariesInCSharp.Tests/UnitTests.cs
PhilBroderick: I would agree with @jshergal here. There is no need for this method and his suggestion would be clearer.

collections-csharp/MergingDictionariesInCSharp/MergingDictionariesInCSharp.Benchmark/MergingMethodsBenchmark.cs
KonstantinosP7: Changed to Random.Shared.
Used var.

collections-csharp/MergingDictionariesInCSharp/MergingDictionariesInCSharp.Tests/UnitTests.cs
KonstantinosP7: Moved _expecetdMergedDictionary to initiate like _dictionaryA, _dictionaryB.

collections-csharp/MergingDictionariesInCSharp/MergingDictionariesInCSharp.Benchmark/MergingMethodsBenchmark.cs
KonstantinosP7: Removed check for existing in dictionaryA

collections-csharp/MergingDictionariesInCSharp/MergingDictionariesInCSharp.Benchmark/MergingMethodsBenchmark.cs
KonstantinosP7: Changed to var

collections-csharp/MergingDictionariesInCSharp/MergingDictionariesInCSharp.Benchmark/MergingMethodsBenchmark.cs
KonstantinosP7: used var & Random.Shared

collections-csharp/MergingDictionariesInCSharp/MergingDictionariesInCSharp/Methods/ConcatMethod.cs
KonstantinosP7: Added Exception test

collections-csharp/MergingDictionariesInCSharp/MergingDictionariesInCSharp/Methods/UnionMethod.cs
KonstantinosP7: Added Argument Exception test

collections-csharp/MergingDictionariesInCSharp/MergingDictionariesInCSharp/Program.cs
KonstantinosP7: Corrected it

collections-csharp/MergingDictionariesInCSharp/MergingDictionariesInCSharp/Methods/UsingHashSetMethod.cs
KonstantinosP7: You are correct Jeff. HAshet collection does not guarantee that the items are stored in order we add them. I changed it to List instead of HashSet.

collections-csharp/MergingDictionariesInCSharp/MergingDictionariesInCSharp.Benchmark/MergingMethodsBenchmark.cs
```diff
@@ -0,0 +1,94 @@
...
+    {
+        _dictionaryA = new Dictionary<int, string>();
+        _dictionaryB = new Dictionary<int, string>();
+
+        var random = Random.Shared;
+
+        // Generate random values for dictionaryA
+        for (var i = 0; i < 100; i++)
+        {
+            var key = random.Next(100);
```
PhilBroderick, line 25: Now that I think about it, is there any point in generating random keys if we are simply iterating from 0-99. Moreover, if we find a key that already exists, we are decrementing `i` so we will always end up with 100 entries. It would simply be easier to add `i` as the key and generate a random value. Same for `_dictionaryB`, but 100-199.

collections-csharp/MergingDictionariesInCSharp/MergingDictionariesInCSharp.Tests/UnitTests.cs
```diff
@@ -0,0 +1,109 @@
...
+            { 2, "Banana" },
+            { 3, "Orange" },
+            { 4, "Grapes" },
+            { 5, "Watermelon" },
+            { 6, "Pineapple" }
+        };
+    }
+
+    [Test]
+    public void ConcatMethod_ShouldMergeDictionaries()
```
PhilBroderick, line 39: These should all be `Given_When_Then`.

PhilBroderick (changes requested):

collections-csharp/MergingDictionariesInCSharp/MergingDictionariesInCSharp.Benchmark/MergingMethodsBenchmark.cs
KonstantinosP7: Ok, implemented it as described.

collections-csharp/MergingDictionariesInCSharp/MergingDictionariesInCSharp.Tests/UnitTests.cs
KonstantinosP7: Changed the names here in every test method

collections-csharp/MergingDictionariesInCSharp/MergingDictionariesInCSharp.Benchmark/MergingMethodsBenchmark.cs
```diff
@@ -11,45 +12,29 @@ public class MergingMethodsBenchmark
...
-            }
+            var randomValue = GenerateRandomValue();
+            _dictionaryB.Add(i, randomValue);            
         }
     }
 
     private string GenerateRandomValue()
     {
-        Random random = new Random();
+        var random = Random.Shared;
```
jshergal, line 37: No need for the variable, just use Random.Shared directly. It is static and thread safe.

PhilBroderick (approved):

PR #1249, opened 2023-07-13
CM-722: How to use the Mutex Class in C#

saraelsa:


csharp-advanced-topics/MutexClassInCSharp/MutexClassInCSharp/Program.cs
```diff
@@ -0,0 +1,16 @@
...
+
+mutex.WaitOne();
+
+for (int number = 1; number <= 50; number++)
+{
+    File.AppendAllText(fileName, $"{number} ");
+    Thread.Sleep(100);
+}
+
+mutex.ReleaseMutex();
```
jshergal, line 13: This could be extracted into a new public static class with just this single method. We could also add a try/finally around the mutex set and release to ensure  releasing even in the presence of an exception. This would have the added benefit of removing the need for the empty class and the reflection call in the testing assembly. We could simply call the function from the test assembly. This main console then could still also call this code.

Just a thought. I might be missing something, but it seems to me that would be cleaner.

csharp-advanced-topics/MutexClassInCSharp/MutexClassInCSharp.Tests/ProgramTests.cs
```diff
@@ -0,0 +1,63 @@
+namespace MutexClassInCSharp.Tests;
+
+public sealed class ProgramTests
+{
+    private const string fileName = "numbers.txt";
+
+    private Action<string[]> _mainMethod
+        = typeof(AssemblyAccessor).Assembly.EntryPoint!.CreateDelegate<Action<string[]>>();
```
jshergal, line 8: See my note about moving the `Main` code into a public static class.

csharp-advanced-topics/MutexClassInCSharp/MutexClassInCSharp.Tests/ProgramTests.cs
```diff
@@ -0,0 +1,63 @@
...
+
+    private async Task RunProgram()
+        => await Task.Run(() => _mainMethod(new string[] { }));
+
+    private async Task EnsureFileConsistency()
+    {
+        string fileContents = await File.ReadAllTextAsync(fileName);
+        File.Delete(fileName);
+
+        List<int> numbers = fileContents
```
jshergal, line 43: we can use var here

csharp-advanced-topics/MutexClassInCSharp/MutexClassInCSharp.Tests/ProgramTests.cs
```diff
@@ -0,0 +1,63 @@
...
+            .TrimEnd()
+            .Split(' ')
+            .Select(int.Parse)
+            .ToList();
+
+        int nextExpectedNumber = 1;
+
+        foreach (int number in numbers)
```
jshergal, line 51: we can use var here

csharp-advanced-topics/MutexClassInCSharp/MutexClassInCSharp.Tests/ProgramTests.cs
```diff
@@ -0,0 +1,63 @@
...
+        {
+            Assert.Equal(nextExpectedNumber, number);
+
+            nextExpectedNumber = nextExpectedNumber + 1;
```
jshergal, line 55: We can combine these two lines into one:
`Assert.Equal(nextExpectedNumber++, number)`

saraelsa:
Thanks for your suggestions! I have implemented the changes and I believe this PR should automatically get updated with those.

PR #1253, opened 2023-07-14
CM-201: How do I generate a stream from a string?

cjaynjoku:


strings-csharp/GenerateStreamFromStringInCSharp/GenerateStreamFromStringInCSharp/GenerateStreamFromStringBenchmark.cs
```diff
@@ -0,0 +1,86 @@
+﻿using BenchmarkDotNet.Attributes;
+using BenchmarkDotNet.Order;
+
+namespace GenerateStreamFromStringInCSharp
+{
+    [MemoryDiagnoser]
+    [Orderer(SummaryOrderPolicy.FastestToSlowest)]
+    public class GenerateStreamFromStringBenchmark
+    {
+        private static readonly string BenchmarkString
```
jshergal, line 10: We can make this const

strings-csharp/GenerateStreamFromStringInCSharp/GenerateStreamFromStringInCSharp/GenerateStreamFromStringMethods.cs
```diff
@@ -0,0 +1,22 @@
...
+            using var stream = new MemoryStream(encoding.GetByteCount(sampleString));
+            using var writer = new StreamWriter(stream, encoding);
+            writer.Write(sampleString);
+            stream.Position = 0;
+
+            return stream;
+        }
+
+        public static Stream ConciselyGenerateStreamFromString(string sampleString, Encoding? encoding = null)
+            => new MemoryStream((encoding ?? Encoding.UTF8).GetBytes(sampleString ?? string.Empty));
```
jshergal, line 20: Due to nullable references being enabled, we can remove the null check on the string.

strings-csharp/GenerateStreamFromStringInCSharp/Tests/GenerateStreamFromStringTests.cs
```diff
@@ -0,0 +1,34 @@
+using GenerateStreamFromStringInCSharp;
+
+namespace Tests
+{
+    public class GenerateStreamFromStringTests
+    {
+        private readonly string _testString
```
jshergal, line 7: We can make this a const

strings-csharp/GenerateStreamFromStringInCSharp/Tests/GenerateStreamFromStringTests.cs
```diff
@@ -0,0 +1,34 @@
...
+        Nullam aliquam auctor fringilla. Nam id orci lacus. Sed placerat justo vitae sem auctor, ac molestie lacus sodales. 
+        Nullam tincidunt lacus ex, ac hendrerit elit tincidunt in. Integer tincidunt accumsan nisl, id aliquet leo convallis in. 
+        Morbi pulvinar odio et est vulputate viverra. Maecenas nibh libero, sollicitudin quis ex ut, egestas auctor mi. 
+        Fusce varius facilisis mauris, ut volutpat libero volutpat in. Aliquam faucibus hendrerit nibh vel porttitor.
+        """;
+
+        [Fact]
+        public void WhenUsingGenerateStreamFromString_ThenReturnsStream()
+        {
+            Stream resultStream = GenerateStreamFromStringMethods.GenerateStreamFromString(_testString);
```
jshergal, line 19: We can use var here

strings-csharp/GenerateStreamFromStringInCSharp/Tests/GenerateStreamFromStringTests.cs
```diff
@@ -0,0 +1,34 @@
...
+
+            Assert.NotNull(resultStream);
+            Assert.IsAssignableFrom<Stream>(resultStream);
```
jshergal, line 22: This assert is redundant as the type returned from the GenerateStream method is a Stream.

strings-csharp/GenerateStreamFromStringInCSharp/GenerateStreamFromStringInCSharp/GenerateStreamFromStringMethods.cs
```diff
@@ -0,0 +1,22 @@
...
+
+namespace GenerateStreamFromStringInCSharp
+{
+    public static class GenerateStreamFromStringMethods
+    {
+        public static Stream GenerateStreamFromString(string sampleString, Encoding? encoding = null)
+        {
+            encoding ??= Encoding.UTF8;
+
+            using var stream = new MemoryStream(encoding.GetByteCount(sampleString));
```
jshergal, line 11: This using statement needs to be removed as this is the return object and therefore the caller is responsible for it. Also, we should never dispose of something and then return it. That leads to use after dispose errors.

strings-csharp/GenerateStreamFromStringInCSharp/GenerateStreamFromStringInCSharp/GenerateStreamFromStringMethods.cs
karimkod, line 20: in addition to @jshergal review, I'd like to suggest we write this as a full body method instead of an expression body method, and we should have a line there that clearly shows the conversion to bytes. 
Altough this is concise in term of number of lines but it's doing a lot of things in one line and it may confuse the reader.

strings-csharp/GenerateStreamFromStringInCSharp/Tests/GenerateStreamFromStringTests.cs
karimkod, line 22: instead of having asserts that are not very strong like not null and IsAssignableFrom<Stream> you can read back the stream to a string and check if the resultant string is equal to _testString, it's a better test, and it protect from forgetting to reset the stream to position 0. 
you can do this in both tests.

karimkod (changes requested):

strings-csharp/GenerateStreamFromStringInCSharp/GenerateStreamFromStringInCSharp/GenerateStreamFromStringMethods.cs
```diff
@@ -0,0 +1,30 @@
+﻿using System.Text;
+
+namespace GenerateStreamFromStringInCSharp
+{
+    public static class GenerateStreamFromStringMethods
+    {
+        public static Stream GenerateStreamFromStringWithStreamWriterAndMemoryStream(string sampleString, Encoding? encoding = null)
```
karimkod, line 7: As explained in the article review, please break this into multiple lines. Read the guidelines about code formatting if you want a reminder to see how this should be applied properly.

strings-csharp/GenerateStreamFromStringInCSharp/GenerateStreamFromStringInCSharp/GenerateStreamFromStringMethods.cs
```diff
@@ -0,0 +1,30 @@
...
+            var stream = new MemoryStream(encoding.GetByteCount(sampleString));
+            var writer = new StreamWriter(stream, encoding);
+            writer.Write(sampleString);
+            writer.Flush();
+            stream.Position = 0;
+
+            return stream;
+        }
+
+        public static Stream GenerateStreamFromStringWithMemoryStreamAndGetBytes(string sampleString, Encoding? encoding = null)
```
karimkod, line 20: As explained in the article review, please break this into multiple lines. Read the guidelines about code formatting if you want a reminder to see how this should be applied properly.

karimkod (changes requested):

strings-csharp/GenerateStreamFromStringInCSharp/Tests/GenerateStreamFromStringTests.cs
```diff
@@ -15,9 +15,9 @@ public class GenerateStreamFromStringTests
         """;
 
         [Fact]
-        public void WhenUsingGenerateStreamFromStringWithStreamWriterAndMemoryStream_ThenReturnsStream()
+        public void WhenUsingMemoryStreamAndStreamWriter_ThenReturnsStream()
         {
-            var resultStream = GenerateStreamFromStringMethods.GenerateStreamFromStringWithStreamWriterAndMemoryStream(_testString);
+            var resultStream = GenerateStreamFromStringMethods.UseMemoryStreamAndStreamWriter(_testString);
 
             var reader = new StreamReader(resultStream, Encoding.UTF8);
```
jshergal: These are missing using statements

strings-csharp/GenerateStreamFromStringInCSharp/Tests/GenerateStreamFromStringTests.cs
```diff
@@ -26,9 +26,9 @@ public void WhenUsingGenerateStreamFromStringWithStreamWriterAndMemoryStream_The
         }
 
         [Fact]
-        public void WhenUsingGenerateStreamFromStringWithMemoryStreamAndGetBytes_ThenReturnsStream()
+        public void WhenUsingGetBytesAndMemoryStream_ThenReturnsStream()
         {
-            var resultStream = GenerateStreamFromStringMethods.GenerateStreamFromStringWithMemoryStreamAndGetBytes(_testString);
+            var resultStream = GenerateStreamFromStringMethods.UseGetBytesAndMemoryStream(_testString);
 
             var reader = new StreamReader(resultStream, Encoding.UTF8);
```
jshergal: These are missing using statements

strings-csharp/GenerateStreamFromStringInCSharp/GenerateStreamFromStringInCSharp/GenerateStreamFromStringMethods.cs
```diff
@@ -4,19 +4,29 @@ namespace GenerateStreamFromStringInCSharp
...
-        public static Stream GenerateStreamFromString(string sampleString, Encoding? encoding = null)
+        public static Stream
+        UseMemoryStreamAndStreamWriter(string sampleString, Encoding? encoding = null)
         {
             encoding ??= Encoding.UTF8;
 
-            using var stream = new MemoryStream(encoding.GetByteCount(sampleString));
-            using var writer = new StreamWriter(stream, encoding);
+            var stream = new MemoryStream(encoding.GetByteCount(sampleString));
+            var writer = new StreamWriter(stream, encoding);
```
jshergal, line 13: This should be disposed, but we need to explicitly invoke the constructor that allows us to leave the stream open:
https://learn.microsoft.com/en-us/dotnet/api/system.io.streamwriter.-ctor?view=net-7.0#system-io-streamwriter-ctor(system-io-stream-system-text-encoding-system-int32-system-boolean)

We can pass in -1 as a buffer size to have it use the default internal buffer size (1024), or we can specify a buffer size, if we want to be explicit.

strings-csharp/GenerateStreamFromStringInCSharp/Tests/GenerateStreamFromStringTests.cs
```diff
@@ -14,21 +15,25 @@ public class GenerateStreamFromStringTests
...
         [Fact]
-        public void WhenUsingGenerateStreamFromString_ThenReturnsStream()
+        public void WhenUsingMemoryStreamAndStreamWriter_ThenReturnsStream()
         {
-            Stream resultStream = GenerateStreamFromStringMethods.GenerateStreamFromString(_testString);
+            var resultStream = GenerateStreamFromStringMethods.UseMemoryStreamAndStreamWriter(_testString);
 
-            Assert.NotNull(resultStream);
-            Assert.IsAssignableFrom<Stream>(resultStream);
+            var reader = new StreamReader(resultStream, Encoding.UTF8);
```
jshergal: We need usings here

strings-csharp/GenerateStreamFromStringInCSharp/Tests/GenerateStreamFromStringTests.cs
```diff
@@ -14,21 +15,25 @@ public class GenerateStreamFromStringTests
...
         }
 
         [Fact]
-        public void WhenUsingConciselyGenerateStreamFromString_ThenReturnsStream()
+        public void WhenUsingGetBytesAndMemoryStream_ThenReturnsStream()
         {
-            Stream resultStream = GenerateStreamFromStringMethods.ConciselyGenerateStreamFromString(_testString);
+            var resultStream = GenerateStreamFromStringMethods.UseGetBytesAndMemoryStream(_testString);
+
+            var reader = new StreamReader(resultStream, Encoding.UTF8);
```
jshergal, line 33: We need usings here

jshergal (commented):
Some usings needed

strings-csharp/GenerateStreamFromStringInCSharp/GenerateStreamFromStringInCSharp/GenerateStreamFromStringMethods.cs
jshergal: Note - my comment here is explicitly related to the StreamWriter, not the MemoryStream as that is being returned from the method.

strings-csharp/GenerateStreamFromStringInCSharp/Tests/GenerateStreamFromStringTests.cs
```diff
@@ -0,0 +1,39 @@
...
+        Morbi pulvinar odio et est vulputate viverra. Maecenas nibh libero, sollicitudin quis ex ut, egestas auctor mi. 
+        Fusce varius facilisis mauris, ut volutpat libero volutpat in. Aliquam faucibus hendrerit nibh vel porttitor.
+        """;
+
+        [Fact]
+        public void WhenUsingMemoryStreamAndStreamWriter_ThenReturnsStream()
+        {
+            var resultStream = GenerateStreamFromStringMethods.UseMemoryStreamAndStreamWriter(_testString);
+
+            var reader = new StreamReader(resultStream, Encoding.UTF8);
```
jshergal, line 22: We need usings here

strings-csharp/GenerateStreamFromStringInCSharp/GenerateStreamFromStringInCSharp/GenerateStreamFromStringMethods.cs
```diff
@@ -10,7 +10,7 @@ public static Stream
             encoding ??= Encoding.UTF8;
 
             var stream = new MemoryStream(encoding.GetByteCount(sampleString));
-            var writer = new StreamWriter(stream, encoding);
+            var writer = new StreamWriter(stream, encoding, -1);
```
jshergal, line 13: This is still missing the using statement and we need to explicitly call the `StreamWriter` overload that leaves the stream open.
https://learn.microsoft.com/en-us/dotnet/api/system.io.streamwriter.-ctor?view=net-7.0#system-io-streamwriter-ctor(system-io-stream-system-text-encoding-system-int32-system-boolean)

strings-csharp/GenerateStreamFromStringInCSharp/GenerateStreamFromStringInCSharp/GenerateStreamFromStringMethods.cs
```diff
@@ -0,0 +1,32 @@
...
+            using var writer = new StreamWriter(stream, encoding, -1, true);
+            writer.Write(sampleString);
+            writer.Flush();
+            stream.Position = 0;
+
+            return stream;
+        }
+
+        public static Stream
+        UseGetBytesAndMemoryStream(string sampleString, Encoding? encoding = null)
```
karimkod, line 22: Instead of breaking this method into multiple lines, let's keep it in one line and change its name to be GetStreamWithGetBytes()
```suggestion
        public static Stream GetStreamWithGetBytes(string sampleString, Encoding? encoding = null)
```

strings-csharp/GenerateStreamFromStringInCSharp/GenerateStreamFromStringInCSharp/GenerateStreamFromStringMethods.cs
```diff
@@ -0,0 +1,32 @@
+﻿using System.Text;
+
+namespace GenerateStreamFromStringInCSharp
+{
+    public static class GenerateStreamFromStringMethods
+    {
+        public static Stream
+        UseMemoryStreamAndStreamWriter(string sampleString, Encoding? encoding = null)
```
karimkod, line 8: Instead of breaking this method into multiple lines, let's keep it in one line and change its name to be GetStreamWithStreamWriter()
```suggestion
        public static Stream GetStreamWithStreamWriter(string sampleString, Encoding? encoding = null)
```

karimkod (changes requested):

karimkod (changes requested):

PR #1262, opened 2023-07-17
CM-7: Source Generator

fabriciogd:


csharp-advanced-topics/SourceGeneratorInCSharp/Generator/ExampleGenerator.cs
```diff
@@ -0,0 +1,34 @@
...
+    $@"namespace {callingEntrypoint.ContainingNamespace.Name}
+{{
+    public static class HelloWorld
+    {{
+        public static void SayHello()
+        {{
+            Console.WriteLine(""Hello From Generator"");
+        }}
+    }}
+}}";
```
jshergal, line 25: A raw string literal would make this cleaner and would remove the need to escape all the brackets and the quote marks. We can still use string interpolation within raw string literals.

csharp-advanced-topics/SourceGeneratorInCSharp/Generator/Helper/SourceGenerationHelper.cs
```diff
@@ -0,0 +1,18 @@
...
+    public class SourceGenerationHelper
+    {
+        public const string Attribute = @"
+namespace Generator
+{
+    [System.AttributeUsage(System.AttributeTargets.Class)]
+    public class GenerateServiceAttribute : System.Attribute
+    {
+    }
+}";
```
jshergal, line 16: If we use a raw string literal here, we can make the const fit into normal formatting.

PR #1271, opened 2023-07-22
Add Action-Func delegates project Mahdi Ghanbari

mahdignb:


mahdignb:
This is the project for the Action and Func delegates in C#
 evaluation article

PR #1273, opened 2023-07-22
CM-707 How to Use Moq to Return a Value That Was Passed Into a Method

asfd1234:


dotnet-testing/UseMoqToReturnPassedInValue/UseMoqToReturnPassedInValue/Printer.cs
```diff
@@ -0,0 +1,16 @@
+﻿namespace UseMoqToReturnPassedInValue
```
jshergal, line 1: Use file level namespace to reduce nesting

dotnet-testing/UseMoqToReturnPassedInValue/Tests/UseMoqToReturnPassedInValueTest.cs
```diff
@@ -0,0 +1,51 @@
...
+            Assert.AreEqual("Hello", result1);
+            Assert.AreEqual("Hello", result2);
+        }
+        
+        [TestMethod]
+        public void GivenMockSetupToReturnPassedInValue_WhenCallingMock_ThenValueReturned()
+        {
+            var mockTranslator = new Mock<IPrinter>();
+            mockTranslator.Setup(iTranslator => iTranslator.Print(It.IsAny<string>())).Returns((string input) => { return input; });
+            IPrinter printer = mockTranslator.Object;
```
jshergal, line 29: Use var

dotnet-testing/UseMoqToReturnPassedInValue/Tests/UseMoqToReturnPassedInValueTest.cs
```diff
@@ -0,0 +1,51 @@
...
+
+            string result1 = printer.Print("Hello");
+            string result2 = printer.Print("Hola");
```
jshergal, line 32: We can use var here. Also it would be good to make both of these strings constants. Something like:
`const string expectedResult1 = "Hello"`
and then we can use those strings again in the assertions.

dotnet-testing/UseMoqToReturnPassedInValue/Tests/UseMoqToReturnPassedInValueTest.cs
```diff
@@ -0,0 +1,51 @@
...
+            Assert.AreEqual("Hello", result1);
+            Assert.AreEqual("Hola", result2);
+        }
+
+        [TestMethod]
+        public void GivenTwoDifferentInputs_WhenPrintingAll_ThenCorrectCountReturned()
+        {
+            var mockTranslator = new Mock<IPrinter>();
+            mockTranslator.Setup(iTranslator => iTranslator.Print(It.IsAny<string>())).Returns((string input) => { return input; });
+            IPrinter printer = mockTranslator.Object;
```
jshergal, line 43: Use var

dotnet-testing/UseMoqToReturnPassedInValue/Tests/UseMoqToReturnPassedInValueTest.cs
```diff
@@ -0,0 +1,51 @@
...
+            var printerQueue = new PrinterQueue(printer);
+
+            var characterCountPrinted = printerQueue.PrintAll(new[] { "Hello", "Hola" });
```
jshergal, line 46: Again we can make these strings constants, and then create an expected count from them and use that expected count in the assertion. This way we are mirroring in the expected result, what we expect to happen in the code.
`const string hello = "Hello";`
`const string hola = "Hola";`
`var expectedCount = hello.Length + hola.Length;`

dotnet-testing/UseMoqToReturnPassedInValue/UseMoqToReturnPassedInValue/PrinterQueue.cs
```diff
@@ -0,0 +1,37 @@
...
+    {
+        var characterCount = 0;
+
+        foreach (var job in jobs)
+        {
+            var result = _printer.Print(jobs[0]); //deliberate bug
+            characterCount += result.Length;
+        }
+
+        return characterCount;
```
jshergal, line 35: This method isn't being called, and so really could be removed. Personally, since it is being listed as a "Why" use case in the article, doesn't mean that it needs to be in the repo.
Second, to make the "bug" a bit more realistic, I would change the foreach into a for loop, as in general we use an indexer within a foreach loop.

dotnet-testing/UseMoqToReturnPassedInValue/Tests/UseMoqToReturnPassedInValueTest.cs
```diff
@@ -0,0 +1,51 @@
...
+
+            Assert.AreEqual("Hello", result1);
+            Assert.AreEqual("Hello", result2);
+        }
+        
+        [TestMethod]
+        public void GivenMockSetupToReturnPassedInValue_WhenCallingMock_ThenValueReturned()
+        {
+            var mockTranslator = new Mock<IPrinter>();
+            mockTranslator.Setup(iTranslator => iTranslator.Print(It.IsAny<string>())).Returns((string input) => { return input; });
```
karimkod, line 28: Better for clarity to break this into multiple lines.

dotnet-testing/UseMoqToReturnPassedInValue/Tests/UseMoqToReturnPassedInValueTest.cs
```diff
@@ -0,0 +1,51 @@
...
+
+            Assert.AreEqual("Hello", result1);
+            Assert.AreEqual("Hola", result2);
+        }
+
+        [TestMethod]
+        public void GivenTwoDifferentInputs_WhenPrintingAll_ThenCorrectCountReturned()
+        {
+            var mockTranslator = new Mock<IPrinter>();
+            mockTranslator.Setup(iTranslator => iTranslator.Print(It.IsAny<string>())).Returns((string input) => { return input; });
```
karimkod, line 42: Better for clarity to break this into multipe lines.

dotnet-testing/UseMoqToReturnPassedInValue/UseMoqToReturnPassedInValue/PrinterQueue.cs
karimkod: I totally agree with what @jshergal has pointed out. Not all code should be in the repository, especially a code to show a deliberate bug. You can remove this from the repo but keep it in the article.

karimkod (changes requested):
My reviews in addition to @jshergal reviews :D

PR #1277, opened 2023-07-23
 Evaluation topic on Action and Func Delegates by @ngeezy19

ngeezy19:
I wrote an evaluation topic on Action and Func Delegates and added Tests to the source code


ngeezy19:
close

PR #1282, opened 2023-07-24
Compare byte arrays

martinchege324:


collections-arrays/CompareByteArraysInCsharp/CompareByteArraysInCsharp/CompareByteArrays.cs
```diff
@@ -76,13 +76,15 @@ public unsafe bool CompareUsingBinaryEquality(byte[] firstArray, byte[] secondAr
...
+                var i = 0;
+
+                for (; i <= arrayLength - vectorSize; i += vectorSize)
                 {
                     if (!VectorEquality(pbtr1 + i, pbtr2 + i))
                         return false;
                 }
 
-                for (int i = arrayLength - (arrayLength % vectorSize); i < arrayLength; i++)
+                for (i = arrayLength - (arrayLength % vectorSize); i < arrayLength; i++)
```
jshergal, line 87: No need to set `i` here, it should already be at the proper index after exiting the previous for loop. See my comment on the previous commit showing the two loops

PR #1283, opened 2023-07-24
CM-160: How to return HTTP 500 from ASP.NET Core Web Api

Almir98:


aspnetcore-webapi/WebApiReturnHttp500/WebApiReturnHttp500/Models/User.cs
Banovvv, line 1: A simple `Record` can be used here.

aspnetcore-webapi/WebApiReturnHttp500/Tests/Usings.cs
```diff
@@ -0,0 +1,8 @@
+// Global using directives
```
mahrncic, line 1: No need for this comment, we already know that these are global using directives because each line starts with `global using`, so this comment can be removed 😄

aspnetcore-webapi/WebApiReturnHttp500/WebApiReturnHttp500/GlobalUsings.cs
```diff
@@ -0,0 +1,6 @@
+// Global using directives
```
mahrncic, line 1: Same as I mentioned before, there is no need for this comment

aspnetcore-webapi/WebApiReturnHttp500/WebApiReturnHttp500/Controllers/UserController.cs
```diff
@@ -0,0 +1,96 @@
...
+    }
+
+    [HttpGet]
+    [Route("GetUsersFirstMethod")]
+    public IActionResult GetUsersFirstMethod()
+    {
+        try
+        {
+            // Simulating an exception by explicitly throwing it
+            throw new Exception("Simulated exception");
```
mahrncic, line 21: You throw this exception in the same exact way in every controller method and put the same comment above it to explain what it does. This is fine, but why don't we try to extract this to a helper private method with the descriptive name?

So, I would create a private method called something like `SimulateException()` and then move the `throw new Exception("Simulated exception");` inside that method. After that, you can just call this helper inside every controller method without the comment, because the method name explains everything.

mahrncic (changes requested):

Banovvv (approved):

mahrncic (approved):

PR #1300, opened 2023-07-31
CM-799 : How to Mark a Method as Deprecated in C#

falgunmodispade:


csharp-advanced-topics/DeprecationInCsharp/DeprecationTests/DeprecationTest.cs
```diff
@@ -0,0 +1,30 @@
+namespace DeprecationTests
+{
+    [TestClass]
+    public class DeprecationTest
+    {
+        [TestMethod]
+        public void GivenInputString_WhenReverseString_Deprecated_ThenReturnReversedString()
+        {
+            string input = "Hello, World!";
```
jshergal, line 9: Since this string is used in both tests, along with the same expected output string, we can move both of them to the class level and make them constants.

csharp-advanced-topics/DeprecationInCsharp/DeprecationTests/DeprecationTest.cs
```diff
@@ -0,0 +1,30 @@
...
+
+            string reversed = StringUtils.ReverseString(input);
```
jshergal, line 11: We should use var here

csharp-advanced-topics/DeprecationInCsharp/DeprecationInCsharp/StringUtils.cs
```diff
@@ -0,0 +1,17 @@
+﻿public class StringUtils
```
jshergal, line 1: Since both methods are static, we can mark this class as static as well

csharp-advanced-topics/DeprecationInCsharp/DeprecationInCsharp/StringUtils.cs
```diff
@@ -0,0 +1,17 @@
...
+        Array.Reverse(charArray);
+        return new string(charArray);
+    }
+
+    public static string ReverseStringV2(string originalString)
+    {
+        var charArray = originalString.ToCharArray();
+        Array.Reverse(charArray);
+        return new string(charArray);
+    }
```
jshergal, line 16: Since we are showing how to mark something as obsolete, we could make this stand out more by actually performing the reverse a different way. I would recommend that we show the method from our article [How to Reverse a String](https://code-maze.com/csharp-reverse-string/). In particular, we could use the [TextEnumerator method](https://github.com/CodeMazeBlog/CodeMazeGuides/blob/main/strings-csharp/StringReverse/StringReverse/Methods/TextElementEnumeratorMethod.cs) which is basically the only proper way to reverse a non-latin/English Unicode string. It would also provide a way to add a link to the other article from within this article.
By adding in the more correct way as V2 we help show a motivation for why a method may become obsolete. Sometimes because it is slower, sometimes because it is broken, sometimes for other reasons or to support new features. But here in particular we can show the new method added to fix broken behavior.

falgunmodispade: Hi @jshergal i have implemented all of your suggestions and thanks for providing suggestions

csharp-advanced-topics/DeprecationInCsharp/DeprecationInCsharp/StringUtils.cs
```diff
@@ -0,0 +1,41 @@
+﻿using System.Globalization;
+
+namespace Deprecation
```
jshergal, line 3: We can use file scoped namespace here to reduce nesting.
https://learn.microsoft.com/en-us/dotnet/csharp/language-reference/proposals/csharp-10.0/file-scoped-namespaces

csharp-advanced-topics/DeprecationInCsharp/DeprecationInCsharp/StringUtils.cs
```diff
@@ -0,0 +1,41 @@
...
+{
+    public class StringUtils
+    {
+        public const string input = "Hello, World!";
+        public const string reverseInput = "!dlroW ,olleH";
```
jshergal, line 8: These should be in the testing class, be private and I would suggest that the second string be renamed expectedResult, or expectedOutput

csharp-advanced-topics/DeprecationInCsharp/DeprecationTests/DeprecationTest.cs
```diff
@@ -0,0 +1,29 @@
+using Deprecation;
```
jshergal, line 1: Since this applies across all tests, we can move it to the global usings

csharp-advanced-topics/DeprecationInCsharp/DeprecationInCsharp/Program.cs
```diff
@@ -0,0 +1,15 @@
+﻿namespace Deprecation
+{
+    public class Program
+    {
+        public static void Main()
+        {
+            var originalString = StringUtils.input;
```
jshergal, line 7: We can just declare a new constant string here as it will make the example clearer. The suggestion for the constant string in the testing was meant to only apply within the testing assembly where we have an input and an expected output. We are using the same input in all tests and have the same expected output in all tests, so it is natural to add them as class member constants there.

csharp-advanced-topics/DeprecationInCsharp/DeprecationInCsharp/Program.cs
```diff
@@ -0,0 +1,15 @@
...
+            var reversed = StringUtils.ReverseString(originalString); // Deprecated method
+            var reversedV2 = StringUtils.ReverseStringV2(originalString); // Recommended method
+
+            Console.WriteLine("Reversed (Deprecated Method): " + reversed);
+            Console.WriteLine("Reversed V2 (Recommended Method): " + reversedV2);
+        }
+    }
+}
```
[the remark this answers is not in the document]
falgunmodispade: Hi @jshergal, I have implemented all of your suggestions but I'm not sure about this one let me know if my changes are correct or not

csharp-advanced-topics/DeprecationInCsharp/DeprecationInCsharp/StringUtils.cs
```diff
@@ -1,11 +1,9 @@
 ﻿using System.Globalization;
 
-namespace Deprecation
+namespace Deprecation.StringUtils
```
jshergal, line 3: A file scoped namespace allows us to reduce nesting, so instead of having:
`namespace MyNamespace`
`{`
`    // some code`
`}`
we can write:
`namespace MyNamespace;`
`// some code`
No brackets needed and so we are able to reduce the nesting withing the file which aids in readability. You can read more here [File scoped namespace](https://learn.microsoft.com/en-us/dotnet/csharp/language-reference/proposals/csharp-10.0/file-scoped-namespaces). Sorry if I didn't link the documentation earlier, I had intended to.

csharp-advanced-topics/DeprecationInCsharp/DeprecationTests/DeprecationTest.cs
```diff
@@ -1,28 +1,29 @@
-using Deprecation;
-
+using Deprecation.StringUtils;
 namespace DeprecationTests
```
jshergal, line 2: You can take a look at the other comment regarding this to see what is meant by file scoped namesapce

csharp-advanced-topics/DeprecationInCsharp/DeprecationInCsharp/StringUtils.cs
falgunmodispade: Thanks for detailed answer i have implemented same you check

csharp-advanced-topics/DeprecationInCsharp/DeprecationTests/DeprecationTest.cs
```diff
@@ -1,29 +1,29 @@
-using Deprecation;
+using Deprecation.StringUtils;
```
jshergal, line 1: This can go in the global usings.

csharp-advanced-topics/DeprecationInCsharp/DeprecationTests/Usings.cs
```diff
@@ -1 +1,2 @@
-global using Microsoft.VisualStudio.TestTools.UnitTesting;
\ No newline at end of file
+global using Microsoft.VisualStudio.TestTools.UnitTesting;
+using Deprecation;
```
jshergal, line 2: Since the namespace was changed, this should be updated

csharp-advanced-topics/DeprecationInCsharp/DeprecationTests/DeprecationTest.cs
falgunmodispade: using Deprecation.StringUtils; moving it to global usings not working

csharp-advanced-topics/DeprecationInCsharp/DeprecationTests/Usings.cs
falgunmodispade: Hi @jshergal due to this obsolete method warning errors Jenkins build is failing

jshergal: > Since the namespace was changed, this should be updated

You didn't update the using here in the global using, you deleted it. That is why the build is failing. It also would have failed on your local machine. Be sure to always build before you commit.

jshergal: Here is the output:

> /var/lib/jenkins/workspace/codemaze-build-pr/csharp-advanced-topics/DeprecationInCsharp/DeprecationTests/DeprecationTest.cs(14,24): error CS0103: The name 'StringUtils' does not exist in the current context [/var/lib/jenkins/workspace/codemaze-build-pr/csharp-advanced-topics/DeprecationInCsharp/DeprecationTests/DeprecationTests.csproj]
> /var/lib/jenkins/workspace/codemaze-build-pr/csharp-advanced-topics/DeprecationInCsharp/DeprecationTests/DeprecationTest.cs(24,26): error CS0103: The name 'StringUtils' does not exist in the current context [/var/lib/jenkins/workspace/codemaze-build-pr/csharp-advanced-topics/DeprecationInCsharp/DeprecationTests/DeprecationTests.csproj]
> 
> Build FAILED.

csharp-advanced-topics/DeprecationInCsharp/DeprecationTests/Usings.cs
```diff
@@ -1,2 +1 @@
 global using Microsoft.VisualStudio.TestTools.UnitTesting;
-using Deprecation;
```
jshergal: You still need the global using here, but it should now be `Deprecation.StringUtils`. Also note that global usings require the word `global` in the front, so this line should be `global using Deprecation.StringUtils;`

csharp-advanced-topics/DeprecationInCsharp/DeprecationTests/DeprecationTest.cs
```diff
@@ -1,4 +1,4 @@
-
+using Deprecation.StringUtils;
```
jshergal, line 1: You only need this in one place. Either here or in global usings. Since this is the namespace under test in this testing assembly it makes sense for it to live in the global usings. This simplifies the file. Since you have now added it to the global using space it is redundant to have it here. If you are using VS it actually shows that the using here is not needed because it is now in the global usings.

image

mahrncic:
Closing this as a new PR will be created.

PR #1302, opened 2023-07-31
Added solution.

alintroncea:


jshergal:
The PR should have the Jira number in the title: 'CM-646: How to Solve HTTP Error 500.31"

PR #1332, opened 2023-08-09
CM-389 dynamically switching connection string

GergoVandor:


dotnet-efcore/DynamicDbContextSwitching/DynamicDbContextSwitching/DataSourceProvider.cs
```diff
@@ -0,0 +1,31 @@
...
+    public DataSource CurrentDataSource { get; set; }
+
+    private readonly IConfiguration _configuration;
+
+    public DataSourceProvider(IConfiguration configuration)
+    {
+        _configuration = configuration;
+    }
+    
+    public string? GetConnectionString()
```
mahrncic, line 16: I don't think the return type needs to be nullable here `string?` since the method will always return a string or it will throw an exception

mahrncic (changes requested):
Looks good, one small change required 😄

dotnet-efcore/DynamicDbContextSwitching/DynamicDbContextSwitching/DataSourceProvider.cs
```diff
@@ -0,0 +1,31 @@
+using Microsoft.Extensions.Configuration;
+
+namespace DynamicDbContextSwitching;
+
+public class DataSourceProvider : IDataSourceProvider
+{
+    public DataSource CurrentDataSource { get; set; }
+
+    private readonly IConfiguration _configuration;
```
mahrncic, line 9: The order here is not correct, private fields should always appear before the public properties

dotnet-efcore/DynamicDbContextSwitching/DynamicDbContextSwitching/DataSourceProvider.cs
GergoVandor: GetConnectionString will never throw an exception, if _configuration is null it will return null. If there is no "ConnectionStrings" section in the appsettings it will return null, and if the ConnectionStrings section has no value called Primary or Secondary it will also return null.
So I'd leave it as is because the UseSqLite method can also accept null as a value, probably it will throw an exception in case of a configuration error. And I'd not complicate the example by adding a custom ConfigurationException. Using the bang operator could be a solution, but then it will defeat the purpose of nullable reference types, so I wouldn't use it either

dotnet-efcore/DynamicDbContextSwitching/DynamicDbContextSwitching/DataSourceProvider.cs
GergoVandor: reordered them

dotnet-efcore/DynamicDbContextSwitching/DynamicDbContextSwitching/DataSourceProvider.cs
mahrncic: Makes sense, good point 👍

mahrncic (approved):

PR #1337, opened 2023-08-11
CM-714: Difference Between String, FormattableString, IFormattable in C#

martinchege324:
First draft

### DIFF
--- a/strings-csharp/StringFormattableString/StringFormattableString/FormattableStringsMethods.cs
+++ b/strings-csharp/StringFormattableString/StringFormattableString/FormattableStringsMethods.cs
@@ -28,11 +28,6 @@ namespace StringFormattableString
             return $"My name is {studentName} and I am {studentAge} years old.";
         }
 
-        public FormattableString FormattableSQLStringExample(string userName)
-        {
-            return $"SELECT * FROM Users WHERE Username = {userName}"; 
-        }
-
         public FormattableString FormattableStringDateExample(DateTime currentDate)
         {  
             return $"Today's date: {currentDate:D}";
--- a/strings-csharp/StringFormattableString/StringFormattableString/FormattableStringsMethods.cs
+++ b/strings-csharp/StringFormattableString/StringFormattableString/FormattableStringsMethods.cs
@@ -1,5 +1,6 @@
 ﻿using BenchmarkDotNet.Attributes;
 using BenchmarkDotNet.Order;
+using Microsoft.Extensions.Logging;
 
 namespace StringFormattableString
 {
@@ -8,22 +9,45 @@ namespace StringFormattableString
     [Orderer(SummaryOrderPolicy.FastestToSlowest)]
     public class FormattableStringsMethods
     {
-        [Benchmark]
-        public string StringExample() 
+        public IEnumerable<object> StringValues() 
         {
-            var studentName = "John";
-            var studentAge = 30;
+            yield return new object [] { "John", 30 };
+        }
 
+        [Benchmark]
+        [ArgumentsSource(nameof(StringValues))]
+        public string StringExample(string studentName, int studentAge) 
+        {
             return $"My name is {studentName} and I am {studentAge} years old.";
         }
 
         [Benchmark]
-        public FormattableString FormattableStringExample() 
+        [ArgumentsSource(nameof(StringValues))]
+        public FormattableString FormattableStringExample(string studentName, int studentAge) 
         {
-            var studentName = "Sean";
-            var studentAge = 40;
-            
             return $"My name is {studentName} and I am {studentAge} years old.";
+        }
+
+        public FormattableString FormattableSQLStringExample(string userName)
+        {
+            return $"SELECT * FROM Users WHERE Username = {userName}"; 
+        }
+
+        public FormattableString FormattableStringDateExample(DateTime currentDate)
+        {  
+            return $"Today's date: {currentDate:D}";
+        }
+
+        public FormattableString FormattableStringLoggingExample(LogLevel level, string message)
+        {
+            return $"[{DateTime.Now}] [{level}] {message}";
+        }
+
+        public FormattableString FormattableStringDynamicStringExample(string itemName, int itemCount)
+        {
+            var messageTemplate = $"You have {(itemCount > 1 ? "items" : "item")} in your cart.";
+
+            return $"{itemCount}: {itemName}; {messageTemplate}";
         }
     }
 }
--- a/strings-csharp/StringFormattableString/StringFormattableString/FormattableStringsMethods.cs
+++ b/strings-csharp/StringFormattableString/StringFormattableString/FormattableStringsMethods.cs
@@ -1,0 +1,25 @@
+﻿namespace StringFormattableString
+{
+    public class FormattableStringsMethods
+    {
+        public string StringExample() 
+        {
+            var studentName = "John";
+            var studentAge = 30;
+            var message = $"My name is {studentName} and I am {studentAge} years old.";
+
+            return message;
+        }
+
+        public FormattableString FormattableStringExample() 
+        {
+            var studentName = "John";
+            var studentAge = 30;
+            FormattableString message = $"My name is {studentName} and I am {studentAge} years old.";
+
+            return message;
+        }
+
+
+    }
+}
--- a/strings-csharp/StringFormattableString/StringFormattableString/FormattableStringsMethods.cs
+++ b/strings-csharp/StringFormattableString/StringFormattableString/FormattableStringsMethods.cs
@@ -1,7 +1,14 @@
-﻿namespace StringFormattableString
+﻿using BenchmarkDotNet.Attributes;
+using BenchmarkDotNet.Order;
+
+namespace StringFormattableString
 {
+    [MemoryDiagnoser]
+    [RankColumn]
+    [Orderer(SummaryOrderPolicy.FastestToSlowest)]
     public class FormattableStringsMethods
     {
+        [Benchmark]
         public string StringExample() 
         {
             var studentName = "John";
@@ -11,6 +18,7 @@
             return message;
         }
 
+        [Benchmark]
         public FormattableString FormattableStringExample() 
         {
             var studentName = "John";
@@ -19,7 +27,5 @@
 
             return message;
         }
-
-
     }
 }
--- a/strings-csharp/StringFormattableString/StringFormattableString/FormattableStringsMethods.cs
+++ b/strings-csharp/StringFormattableString/StringFormattableString/FormattableStringsMethods.cs
@@ -13,19 +13,17 @@ namespace StringFormattableString
         {
             var studentName = "John";
             var studentAge = 30;
-            var message = $"My name is {studentName} and I am {studentAge} years old.";
 
-            return message;
+            return $"My name is {studentName} and I am {studentAge} years old.";
         }
 
         [Benchmark]
         public FormattableString FormattableStringExample() 
         {
-            var studentName = "John";
-            var studentAge = 30;
-            FormattableString message = $"My name is {studentName} and I am {studentAge} years old.";
-
-            return message;
+            var studentName = "Sean";
+            var studentAge = 40;
+            
+            return $"My name is {studentName} and I am {studentAge} years old.";
         }
     }
 }
--- a/strings-csharp/StringFormattableString/StringFormattableString/Program.cs
+++ b/strings-csharp/StringFormattableString/StringFormattableString/Program.cs
@@ -1,0 +1,2 @@
+﻿// See https://aka.ms/new-console-template for more information
+Console.WriteLine("Hello, World!");
--- a/strings-csharp/StringFormattableString/StringFormattableString/Program.cs
+++ b/strings-csharp/StringFormattableString/StringFormattableString/Program.cs
@@ -1,4 +1,3 @@
 ﻿using BenchmarkDotNet.Running;
-using StringFormattableString;
 
-var summary = BenchmarkRunner.Run<FormattableStringsMethods>();
+BenchmarkRunner.Run<StringFormattableString.FormattableStringsMethods>();
--- a/strings-csharp/StringFormattableString/StringFormattableString/Program.cs
+++ b/strings-csharp/StringFormattableString/StringFormattableString/Program.cs
@@ -1,2 +1,4 @@
-﻿// See https://aka.ms/new-console-template for more information
-Console.WriteLine("Hello, World!");
+﻿using BenchmarkDotNet.Running;
+using StringFormattableString;
+
+var summary = BenchmarkRunner.Run<FormattableStringsMethods>();
--- a/strings-csharp/StringFormattableString/StringFormattableString/TemperatureFormat.cs
+++ b/strings-csharp/StringFormattableString/StringFormattableString/TemperatureFormat.cs
@@ -1,0 +1,24 @@
+﻿namespace StringFormattableString
+{
+    public class TemperatureFormat : IFormattable
+    {
+        private double temperatureValue;
+
+        public TemperatureFormat() { }
+
+        public TemperatureFormat(double temperatureValue)
+        {
+            this.temperatureValue = temperatureValue;
+        }
+
+        public string ToString(string format, IFormatProvider formatProvider)
+        {
+            if (format == "C")
+                return $"{temperatureValue} °C";
+            else if (format == "F")
+                return $"{temperatureValue * 9 / 5 + 32} °F";
+            else
+                return $"{temperatureValue + 273.15} K";
+        }
+    }
+}
--- a/strings-csharp/StringFormattableString/StringFormattableString/TemperatureFormat.cs
+++ b/strings-csharp/StringFormattableString/StringFormattableString/TemperatureFormat.cs
@@ -9,14 +9,11 @@
             _temperatureValue = temperatureValue;
         }
 
-        public string ToString(string format, IFormatProvider formatProvider)
+        public string ToString(string format, IFormatProvider _) => format switch
         {
-            if (format == "C")
-                return $"{_temperatureValue} C";
-            else if (format == "F")
-                return $"{_temperatureValue * 9 / 5 + 32} F";
-            else
-                return $"{_temperatureValue + 273.15} K";
-        }
+            "F" => $"{_temperatureValue * 9 / 5 + 32} F",
+            "K" => $"{_temperatureValue + 273.15} K",
+            _ => $"{_temperatureValue} C"
+        };
     }
 }
--- a/strings-csharp/StringFormattableString/StringFormattableString/TemperatureFormat.cs
+++ b/strings-csharp/StringFormattableString/StringFormattableString/TemperatureFormat.cs
@@ -6,7 +6,7 @@
 
         public TemperatureFormat(double temperatureValue)
         {
-            this._temperatureValue = temperatureValue;
+            _temperatureValue = temperatureValue;
         }
 
         public string ToString(string format, IFormatProvider formatProvider)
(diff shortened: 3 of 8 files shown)


strings-csharp/StringFormattableString/StringFormattableString/TemperatureFormat.cs
```diff
@@ -0,0 +1,24 @@
+﻿namespace StringFormattableString
+{
+    public class TemperatureFormat : IFormattable
+    {
+        private double temperatureValue;
```
PhilBroderick, line 5: Should be `_temperatureValue`

strings-csharp/StringFormattableString/StringFormattableStringTests/FormattableStringsUnitTests.cs
```diff
@@ -0,0 +1,53 @@
...
+using System.Runtime.CompilerServices;
+
+namespace StringFormattableStringTests
+{
+    [TestClass]
+    public class FormattableStringsUnitTests
+    {
+        private readonly FormattableStringsMethods _formattableStringMethods;
+        private readonly string _sampleString;
+        private readonly FormattableString _sampleFormattableString;
```
PhilBroderick, line 11: As these are only used in a single method each, we don't need to define then as private fields. This ties into a comment I left on the feedback doc. If we move these into the methods, it will make things clearer in the article.

strings-csharp/StringFormattableString/StringFormattableStringTests/FormattableStringsUnitTests.cs
```diff
@@ -0,0 +1,53 @@
...
+        public void GivenAStringObject_WhenStringOperationsApplied_VerifyAccurateResults()
+        {
+            var stringObj = "Jane";
+
+            Assert.AreEqual(stringObj.Count(), stringObj.Length);
+            Assert.IsFalse(stringObj.Equals(_sampleString));
+            Assert.AreEqual(_sampleString.CompareTo(stringObj), 1);
+            Assert.IsInstanceOfType(_sampleString, typeof(string));
+        }
+        [TestMethod]
```
PhilBroderick, line 30: Need a newline here.

PhilBroderick (changes requested):
Few changes required.

strings-csharp/StringFormattableString/StringFormattableString/TemperatureFormat.cs
```diff
@@ -0,0 +1,22 @@
+﻿namespace StringFormattableString
+{
+    public class TemperatureFormat : IFormattable
+    {
+        private double _temperatureValue;
+
+        public TemperatureFormat(double temperatureValue)
+        {
+            this._temperatureValue = temperatureValue;
```
jshergal, line 9: `this` is not needed.

strings-csharp/StringFormattableString/StringFormattableStringTests/FormattableStringsUnitTests.cs
```diff
@@ -0,0 +1,53 @@
...
+
+        [TestMethod]
+        public void GivenAFormattableStringObject_WhenFormattableStringOperationsApplied_VerifyAccurateResults()
+        {
+            var sampleFormattableString = _formattableStringMethods.FormattableStringExample();
+
+            Assert.AreEqual("John", sampleFormattableString.GetArgument(0));
+            Assert.AreEqual(sampleFormattableString.ArgumentCount, 2);
+            Assert.IsInstanceOfType(sampleFormattableString, typeof(FormattableString));
+            Assert.IsInstanceOfType(sampleFormattableString.ToString(), typeof(string));
```
jshergal, line 36: This `Assert` isn't necessary as it is in essence testing the framework. By definition calling `ToString` returns a string.

strings-csharp/StringFormattableString/StringFormattableStringTests/FormattableStringsUnitTests.cs
```diff
@@ -0,0 +1,53 @@
+using StringFormattableString;
+using System.Runtime.CompilerServices;
```
jshergal, line 2: This using is not needed.

strings-csharp/StringFormattableString/StringFormattableStringTests/FormattableStringsUnitTests.cs
```diff
@@ -0,0 +1,53 @@
...
+{
+    [TestClass]
+    public class FormattableStringsUnitTests
+    {
+        private readonly FormattableStringsMethods _formattableStringMethods;
+
+        public FormattableStringsUnitTests()
+        {
+            _formattableStringMethods = new FormattableStringsMethods();
+        }
```
jshergal, line 14: This could be simplified into a member initializer:
`private readonly FormattableStringsMethods _formattableStringMethods = new();`

strings-csharp/StringFormattableString/StringFormattableString/FormattableStringsMethods.cs
```diff
@@ -0,0 +1,31 @@
...
+    [MemoryDiagnoser]
+    [RankColumn]
+    [Orderer(SummaryOrderPolicy.FastestToSlowest)]
+    public class FormattableStringsMethods
+    {
+        [Benchmark]
+        public string StringExample() 
+        {
+            var studentName = "John";
+            var studentAge = 30;
```
jshergal, line 15: Since these values are shared between both benchmarks, and are constant strings, we could make them member constants of the benchmark class.

strings-csharp/StringFormattableString/StringFormattableString/FormattableStringsMethods.cs
```diff
@@ -0,0 +1,31 @@
...
+            var message = $"My name is {studentName} and I am {studentAge} years old.";
+
+            return message;
```
jshergal, line 18: The variable is not necessary, we can just return the construct.

strings-csharp/StringFormattableString/StringFormattableString/FormattableStringsMethods.cs
```diff
@@ -0,0 +1,31 @@
...
+        }
+
+        [Benchmark]
+        public FormattableString FormattableStringExample() 
+        {
+            var studentName = "John";
+            var studentAge = 30;
+            FormattableString message = $"My name is {studentName} and I am {studentAge} years old.";
+
+            return message;
```
jshergal, line 28: This variable is not necessary, we can just return it directly

strings-csharp/StringFormattableString/StringFormattableString/Program.cs
```diff
@@ -0,0 +1,4 @@
+﻿using BenchmarkDotNet.Running;
+using StringFormattableString;
+
+var summary = BenchmarkRunner.Run<FormattableStringsMethods>();
```
jshergal, line 4: The variable is not necessary.

martinchege324:
@jshergal, everything should be fine now.

strings-csharp/StringFormattableString/StringFormattableString/FormattableStringsMethods.cs
PhilBroderick: @martinchege324 it looks like this one hasn't been addressed.

PhilBroderick: Sorry indeed you have. I'm happy with this now.

strings-csharp/StringFormattableString/StringFormattableStringTests/FormattableStringsUnitTests.cs
```diff
@@ -0,0 +1,44 @@
...
+    {
+        private readonly FormattableStringsMethods _formattableStringMethods = new();
+    
+        [TestMethod]
+        public void GivenAStringObject_WhenStringOperationsApplied_VerifyAccurateResults()
+        {
+            var stringObj = "Jane";
+            var sampleString = _formattableStringMethods.StringExample();
+
+            Assert.AreEqual(stringObj.Count(), stringObj.Length);
```
jshergal, line 14: This assert isn't testing anything related to formattable strings. It is simply asserting that the Enumerable.Count method matches the result of string.Length.

strings-csharp/StringFormattableString/StringFormattableString/FormattableStringsMethods.cs
jshergal: > @PhilBroderick I used different values for both methods; let me know if this change is still required.

Since this is a benchmark we shouldn't be using different values for the interpolated string. It should be the same between the two benchmarks. It would be better to move these to class constants and then use the class constants within the benchmark methods.

strings-csharp/StringFormattableString/StringFormattableStringTests/FormattableStringsUnitTests.cs
```diff
@@ -0,0 +1,44 @@
...
+            Assert.IsFalse(stringObj.Equals(sampleString));
+            Assert.AreEqual(sampleString.CompareTo(stringObj), 1);
```
jshergal, line 16: As a whole, it is hard to grasp what we are actually testing in this unit test. It would probably make the tests clearer if the `FormattableStringMethods` methods were changed to take both a name and an age argument. Then in the tests we could verify that the returned result is the type expected and contains the arguments/values expected. Then we just need to update the benchmark to pass in the arguments using either the `Argument` or `ArgumentSource` attributes.

martinchege324:
@jshergal it should be fine now.

strings-csharp/StringFormattableString/StringFormattableString/FormattableStringsMethods.cs
```diff
@@ -8,22 +9,45 @@ namespace StringFormattableString
...
-            var studentName = "Sean";
-            var studentAge = 40;
-            
             return $"My name is {studentName} and I am {studentAge} years old.";
         }
+
+        public FormattableString FormattableSQLStringExample(string userName)
+        {
+            return $"SELECT * FROM Users WHERE Username = {userName}"; 
+        }
```
jshergal, line 34: The username string is not quoted properly, it should be `'{userName}'`, but  that being said I would highly recommend removing this example as it is bad practice. This code leads to SQL injection attacks. The statement should be written:
`"SELECT * FROM Users WHERE Username = @UserName"`
And then when it is used we add the parameter value like this:
`cmd.Parameters.Add(new SqlParameter("@UserName", userName));`

strings-csharp/StringFormattableString/StringFormattableStringTests/FormattableStringsUnitTests.cs
```diff
@@ -9,9 +11,10 @@ public class FormattableStringsUnitTests
         public void GivenAStringObject_WhenStringOperationsApplied_VerifyAccurateResults()
         {
             var stringObj = "Jane";
```
jshergal, line 13: This can go away as it doesn't really serve any purpose. A better way to write this test would be something like this:
`const string expectedStudentName = "John";`
`const int expectedStudentAge = 30;`
`var sampleString = _formattableStringMethods.StringExample(expectedStudentName, expectedStudentAge);`
`Assert.IsTrue(sampleString.Contains(expectedStudentAge);`
 etc.
Since we know exactly what string should be returned, since the method itself is building the string, we could even check just the whole expected string.

martinchege324:
@jshergal, I have applied your feedback; let me know if it's okay now.

jshergal:
> @jshergal, I have applied your feedback; let me know if it's okay now.

Yes, the change in that unit test looks good. There is other feedback I have added that still needs to be addressed.

One thing, in looking at the unit test where you made the recent changes, it feels to me like the check for the type should be the first assertion, since the other two assertions are based on the fact that the return value is actually a string. As a whole I am not sure the value that is added by asserting that the return value is a string, especially given that we are calling `Contains` on the return value, which will basically fail to compile if the object was not a string (with the exception of course that some other return type could have a `Contains` method). As a whole, if we want to keep that assertion, then from a logical flow perspective it should be the first assertion. We can think of the ordering in this way:

1. Setup
2. Run some operation which gives a result
3. Verify the result is of an expected type
4. Now verify that the result value matches what is expected

martinchege324:
@jshergal, it's sorted; thanks for your feedback.

strings-csharp/StringFormattableString/StringFormattableStringTests/FormattableStringsUnitTests.cs
```diff
@@ -0,0 +1,95 @@
...
+            Assert.IsInstanceOfType(sampleString, typeof(string)); 
+            Assert.IsTrue(sampleString.Contains(studentName));
+            Assert.IsTrue(sampleString.Contains(studentAge.ToString()));
+        }
+
+        [TestMethod]
+        public void GivenAFormattableStringObject_WhenFormattableStringOperationsApplied_VerifyAccurateResults()
+        {
+            var studentName = "Sean";
+            var studentAge = 40;
```
jshergal, line 26: Can be made constants

strings-csharp/StringFormattableString/StringFormattableStringTests/FormattableStringsUnitTests.cs
```diff
@@ -0,0 +1,95 @@
...
+            var sampleFormattableString = _formattableStringMethods.FormattableStringExample(studentName, studentAge);
+
+            Assert.AreEqual(studentName, sampleFormattableString.GetArgument(0));
+            Assert.AreEqual(studentAge, sampleFormattableString.GetArgument(1));
+            Assert.AreEqual(sampleFormattableString.ArgumentCount, 2);
+            Assert.IsInstanceOfType(sampleFormattableString, typeof(FormattableString));
```
jshergal, line 32: Same argument for reordering would apply here. We should try to think about our assertions as a process going from general to more specific, especially in the context of multiple assertions within a test. An assertion is our way of "proving" something and so the assertions should always follow a logical ordering and be related to either what we are:
1. setting up (i.e. if we are testing that a method adds something to an empty list, we could first assert that the list is empty as part of our setup, as a way of "proving" that the test is valid).
2. we are validating, but again in the presence of multiple assertions to move from general to specific.

strings-csharp/StringFormattableString/StringFormattableStringTests/FormattableStringsUnitTests.cs
```diff
@@ -0,0 +1,95 @@
...
+        }
+
+        [TestMethod]
+        public void GivenAFormattableStringSQLObject_WhenFormattableStringOperationsApplied_VerifyAccurateResults()
+        {
+            var userName = "John";
```
jshergal, line 38: Should be a constant

strings-csharp/StringFormattableString/StringFormattableStringTests/FormattableStringsUnitTests.cs
```diff
@@ -0,0 +1,95 @@
...
+
+        [TestMethod]
+        public void GivenAFormattableStringDateObject_WhenFormattableStringOperationsApplied_VerifyAccurateResults()
+        {
+            var currentDate = DateTime.Now;
+            var sampleFormattableString = _formattableStringMethods.FormattableStringDateExample(currentDate);
+
+            Assert.IsTrue(sampleFormattableString.ToString().Contains(currentDate.ToString("D")));
+            Assert.IsInstanceOfType(sampleFormattableString.GetArgument(0), typeof(DateTime));
+            Assert.IsInstanceOfType(sampleFormattableString, typeof(FormattableString));
```
jshergal, line 53: Should be first assertion

strings-csharp/StringFormattableString/StringFormattableStringTests/FormattableStringsUnitTests.cs
```diff
@@ -0,0 +1,95 @@
...
+        }
+
+        [TestMethod]
+        public void GivenAFormattableStringDateObject_WhenFormattableStringOperationsApplied_VerifyAccurateResults()
+        {
+            var currentDate = DateTime.Now;
+            var sampleFormattableString = _formattableStringMethods.FormattableStringDateExample(currentDate);
+
+            Assert.IsTrue(sampleFormattableString.ToString().Contains(currentDate.ToString("D")));
+            Assert.IsInstanceOfType(sampleFormattableString.GetArgument(0), typeof(DateTime));
```
jshergal, line 52: Should be second assertion

strings-csharp/StringFormattableString/StringFormattableStringTests/FormattableStringsUnitTests.cs
```diff
@@ -0,0 +1,95 @@
...
+            Assert.IsInstanceOfType(sampleFormattableString, typeof(FormattableString));
+        }
+
+        [TestMethod]
+        public void GivenAFormattableStringDateObject_WhenFormattableStringOperationsApplied_VerifyAccurateResults()
+        {
+            var currentDate = DateTime.Now;
+            var sampleFormattableString = _formattableStringMethods.FormattableStringDateExample(currentDate);
+
+            Assert.IsTrue(sampleFormattableString.ToString().Contains(currentDate.ToString("D")));
```
jshergal, line 51: Should be final assertion

strings-csharp/StringFormattableString/StringFormattableStringTests/FormattableStringsUnitTests.cs
```diff
@@ -0,0 +1,95 @@
...
+        public void GivenAFormattableStringLogObject_WhenFormattableStringOperationsApplied_VerifyAccurateResults()
+        {
+            var level = LogLevel.Information;
+            var message = "Logging method invoked";
+            var sampleFormattableString = _formattableStringMethods.FormattableStringLoggingExample(level, message);
+
+            Assert.IsInstanceOfType(sampleFormattableString.GetArgument(0), typeof(DateTime));
+            Assert.IsInstanceOfType(sampleFormattableString.GetArgument(1), typeof(LogLevel));
+            Assert.IsInstanceOfType(sampleFormattableString.GetArgument(2), typeof(string));
+            Assert.IsInstanceOfType(sampleFormattableString, typeof(FormattableString));
```
jshergal, line 66: Should be first assertion

strings-csharp/StringFormattableString/StringFormattableStringTests/FormattableStringsUnitTests.cs
```diff
@@ -0,0 +1,95 @@
...
+        }
+
+        [TestMethod]
+        public void GivenAFormattableDynamicStringObject_WhenFormattableStringOperationsApplied_VerifyAccurateResults()
+        {
+            var itemName = "Laptop";
+            var itemCount = 3;
```
jshergal, line 73: Can be const

strings-csharp/StringFormattableString/StringFormattableStringTests/FormattableStringsUnitTests.cs
```diff
@@ -0,0 +1,95 @@
...
+            var sampleFormattableString = _formattableStringMethods.FormattableStringDynamicStringExample(itemName, itemCount);
+
+            Assert.AreEqual(itemCount, sampleFormattableString.GetArgument(0));
+            Assert.AreEqual(itemName, sampleFormattableString.GetArgument(1));
+            Assert.IsInstanceOfType(sampleFormattableString, typeof(FormattableString));
```
jshergal, line 78: Should be first assetion

martinchege324:
@jshergal, let me know if I have missed anything.

strings-csharp/StringFormattableString/StringFormattableString/FormattableStringsMethods.cs
jshergal: @martinchege324 - This one still has not been resolved. The example should really just be removed as it is bad practice.

martinchege324:
@jshergal, it should be fine now

PR #1343, opened 2023-08-13
CM-771: Fastest Way to Read a Text File in C#

cjaynjoku:


files-csharp/FastestWayToReadATextFileInCsharp/FastestWayToReadATextFileInCsharp/Program.cs
```diff
@@ -0,0 +1,12 @@
+﻿using BenchmarkDotNet.Running;
+
+namespace FastestWayToReadATextFileInCsharp
```
jshergal, line 3: We can change this file to use top level statements which will greatly simplify this file.

files-csharp/FastestWayToReadATextFileInCsharp/FastestWayToReadATextFileInCsharp/WaysToReadATextFileInCsharp.cs
```diff
@@ -0,0 +1,86 @@
+﻿using System.Text;
+
+namespace FastestWayToReadATextFileInCsharp
```
jshergal, line 3: We can use file scoped namespace to reduce nesting in this file

files-csharp/FastestWayToReadATextFileInCsharp/FastestWayToReadATextFileInCsharp/WaysToReadATextFileInCsharp.cs
```diff
@@ -0,0 +1,86 @@
...
+            return stringBuilder.ToString();
+        }
+
+        public string UseBufferedStreamObject()
+        {
+            var stringBuilder = new StringBuilder();
+
+            using (var fileStream = new FileStream(_sampleFilePath, FileMode.Open))
+            using (var bufferedStream = new BufferedStream(fileStream))
+            using (var streamReader = new StreamReader(bufferedStream))
```
jshergal, line 72: For consistency, we should make these using declarations as we did in the previous code block.

files-csharp/FastestWayToReadATextFileInCsharp/Tests/WaysToReadATextFileInCsharpTests.cs
```diff
@@ -0,0 +1,81 @@
+﻿using FastestWayToReadATextFileInCsharp;
+
+namespace Tests
```
jshergal, line 3: We can reduce nesting with file scoped namespace

files-csharp/FastestWayToReadATextFileInCsharp/Tests/WaysToReadATextFileInCsharpTests.cs
```diff
@@ -0,0 +1,81 @@
...
+            = Path.Combine(Directory.GetParent(Directory.GetCurrentDirectory()).Parent.Parent.FullName, "SampleTextFile.txt");
+
+        private readonly WaysToReadATextFileInCsharp _textFileReader;
+
+        public WaysToReadATextFileInCsharpTests()
+        {
+            _textFileReader = new WaysToReadATextFileInCsharp(_sampleFilePath);
+        }
+
+        private readonly string _expected
```
jshergal, line 17: This can be made constant

files-csharp/FastestWayToReadATextFileInCsharp/Tests/WaysToReadATextFileInCsharpTests.cs
```diff
@@ -0,0 +1,81 @@
...
+        {
+            var result = _textFileReader.UseFileReadAllLines();
+
+            Assert.Equal(_expected, result);
+        }
+
+        [Fact]
+        public void WhenReadingATextFileWithFileReadAllText_ThenReturnsAString()
+        {
+            var result = _textFileReader.UseFileReadLines();
```
jshergal, line 36: This test is called ReadAllText, but we are using ReadAllLines as in the previous test.

files-csharp/FastestWayToReadATextFileInCsharp/Tests/WaysToReadATextFileInCsharpTests.cs
```diff
@@ -0,0 +1,81 @@
+﻿using FastestWayToReadATextFileInCsharp;
+
+namespace Tests
+{
+    public class WaysToReadATextFileInCsharpTests
```
jshergal, line 5: A note with all of the tests. The test case method name states that the expectation is the return of a string, but the assertions are that the methods return expected text. I do believe the expected text assertion is the correct assertion, but the test names should be modified to reflect what we are asserting.

files-csharp/FastestWayToReadATextFileInCsharp/Tests/WaysToReadATextFileInCsharpTests.cs
```diff
@@ -0,0 +1,81 @@
+﻿using FastestWayToReadATextFileInCsharp;
```
jshergal, line 1: Move this to the global usings

files-csharp/FastestWayToReadATextFileInCsharp/Tests/WaysToReadATextFileInCsharpTests.cs
jshergal: As an additional thought, since this is also the expected text from the sample file. Rather than having the sample file, we could simply write this text to a temp file and then use that temp file in the tests (being sure to clean up the temp file when we are done). This would also make the tests more robust as we would not be facing the issue of having to keep both the file and this string in synch.

files-csharp/FastestWayToReadATextFileInCsharp/FastestWayToReadATextFileInCsharp/WaysToReadATextFileInCsharpBenchmark.cs
```diff
@@ -0,0 +1,43 @@
+﻿using BenchmarkDotNet.Attributes;
+using BenchmarkDotNet.Order;
+
+namespace FastestWayToReadATextFileInCsharp
```
jshergal, line 4: We can use file scoped namespace to reduce nesting

files-csharp/FastestWayToReadATextFileInCsharp/FastestWayToReadATextFileInCsharp/WaysToReadATextFileInCsharpBenchmark.cs
```diff
@@ -0,0 +1,43 @@
...
+    [MemoryDiagnoser]
+    [Orderer(SummaryOrderPolicy.FastestToSlowest)]
+    public class WaysToReadATextFileInCsharpBenchmark
+    {
+        private static readonly string _sampleFilePath
+           = Path.Combine(Directory.GetParent(Directory.GetCurrentDirectory()).Parent.Parent.Parent.Parent.Parent.Parent.FullName, "BenchmarkTextFile.txt");
+
+        private WaysToReadATextFileInCsharp _textFileReader = new(_sampleFilePath);
+
+        [Benchmark]
```
jshergal, line 15: Since each of the benchmarked methods are returning a string, we should set the return type of the benchmark methods to match. This also helps to ensure no compiler optimizations that could turn the call into a noop since the return value is being dumped.

files-csharp/FastestWayToReadATextFileInCsharp/FastestWayToReadATextFileInCsharp/WaysToReadATextFileInCsharp.cs
```diff
@@ -0,0 +1,86 @@
...
+        }
+
+        public string UseFileReadAllText() => File.ReadAllText(_sampleFilePath);
+        
+        public string UseFileReadLines()
+        {
+            var stringArray = File.ReadLines(_sampleFilePath);
+
+            return string.Join(Environment.NewLine, stringArray);
+        }
```
jshergal, line 28: We should probably change both of these methods to use a StringBuilder as we will be benchmarking them. It is an unfair comparison, especially considering the size of the benchmark text file, to compare a call to String.Join with a large array in comparison with a StringBuilder.

files-csharp/FastestWayToReadATextFileInCsharp/FastestWayToReadATextFileInCsharp/WaysToReadATextFileInCsharp.cs
```diff
@@ -0,0 +1,86 @@
...
+            var stringArray = File.ReadAllLines(_sampleFilePath);
+
+            return string.Join(Environment.NewLine, stringArray);
+        }
+
+        public string UseFileReadAllText() => File.ReadAllText(_sampleFilePath);
+        
+        public string UseFileReadLines()
+        {
+            var stringArray = File.ReadLines(_sampleFilePath);
```
jshergal, line 25: Since ReadLines is returning an IEnumerable, we should probably rename this variable, maybe `lines`

files-csharp/FastestWayToReadATextFileInCsharp/FastestWayToReadATextFileInCsharp/WaysToReadATextFileInCsharp.cs
```diff
@@ -0,0 +1,86 @@
...
+            using var streamReader = new StreamReader(_sampleFilePath);
+            var stringBuilder = new StringBuilder();
+            string fileLine;
+
+            while ((fileLine = streamReader.ReadLine()) != null)
+            {
+                stringBuilder.AppendLine(fileLine);
+            }
+
+            return stringBuilder.ToString().TrimEnd();
```
jshergal, line 41: Can this TrimEnd be avoided? This will cause a second string allocation before the return.

files-csharp/FastestWayToReadATextFileInCsharp/FastestWayToReadATextFileInCsharp/WaysToReadATextFileInCsharp.cs
```diff
@@ -0,0 +1,86 @@
...
+            {
+                string fileLine;
+
+                while ((fileLine = streamReader.ReadLine()) != null)
+                {
+                    stringBuilder.AppendLine(fileLine);
+                }
+            }
+
+            return stringBuilder.ToString().TrimEnd();
```
jshergal, line 82: Can this TrimEnd be avoided? It results in a second string allocation before the return.

files-csharp/FastestWayToReadATextFileInCsharp/FastestWayToReadATextFileInCsharp/WaysToReadATextFileInCsharp.cs
```diff
@@ -0,0 +1,86 @@
...
+            using (var fileStream = new FileStream(_sampleFilePath, FileMode.Open))
+            using (var bufferedStream = new BufferedStream(fileStream))
+            using (var streamReader = new StreamReader(bufferedStream))
+            {
+                string fileLine;
+
+                while ((fileLine = streamReader.ReadLine()) != null)
+                {
+                    stringBuilder.AppendLine(fileLine);
+                }
```
jshergal, line 79: Could we just use StreamReader.ReadToEnd here? I would think it would be more performant.

jshergal:
One thing to note and be careful about is in the code where we are using `StreamReader.ReadLine`. `ReadLine` strips the new line characters from the returned string. In our sample code we are then using `StringBuilder.AppendLine` to add those lines to our final string. This can have the side effect of returning a different string than the one that exists in the file as `AppendLine` is going to use the whatever `Environment.NewLine` uses. If the file, for example contained only '\r' as the end of line character, but our current running system had '\r\n', then we are not returning a true representation of the text. That may be acceptable in some cases, but we should be careful to point that out. This also applies in the `File.ReadAllLines` and `File.ReadLines` code where we are also using `Environment.NewLine`.

files-csharp/FastestWayToReadATextFileInCsharp/FastestWayToReadATextFileInCsharp/WaysToReadATextFileInCsharp.cs
```diff
@@ -0,0 +1,86 @@
...
+
+            return stringBuilder.ToString();
+        }
+
+        public string UseBufferedStreamObject()
+        {
+            var stringBuilder = new StringBuilder();
+
+            using (var fileStream = new FileStream(_sampleFilePath, FileMode.Open))
+            using (var bufferedStream = new BufferedStream(fileStream))
```
jshergal, line 71: Since we are looking at using the BufferedStream here, we should experiment with calling the FileStream constructor that disables buffering. By default FileStream will use an internal buffer of 4096 bytes. So this code ends up "double buffering".

files-csharp/FastestWayToReadATextFileInCsharp/FastestWayToReadATextFileInCsharp/WaysToReadATextFileInCsharp.cs
```diff
@@ -0,0 +1,86 @@
...
+        {
+            using var streamReader = new StreamReader(_sampleFilePath);
+
+            return streamReader.ReadToEnd();
+        }
+
+        public string UseStreamReaderReadBlock()
+        {
+            using var streamReader = new StreamReader(_sampleFilePath);
+            var buffer = new char[1024];
```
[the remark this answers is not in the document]
karimkod: @cjaynjoku For the span suggestion, can you please keep this with the array, and add one more method with the Span, this can affect performance. 
Add also these buffer size detail in the article and justify why we chose 4096 as @jshergal  suggested.

files-csharp/FastestWayToReadATextFileInCsharp/FastestWayToReadATextFileInCsharp/WaysToReadATextFileInCsharp.cs
karimkod: @cjaynjoku same here, we can add another method that doesn't do double buffering by disabled FileStream buffering. and we need to explain it in the article as well.

files-csharp/FastestWayToReadATextFileInCsharp/FastestWayToReadATextFileInCsharp/WaysToReadATextFileInCsharp.cs
karimkod: @jshergal the article is about benchmarking different ways, so it's important to have this one and also, ReadToEnd in another method.

karimkod: @cjaynjoku I think you can skip this one.

files-csharp/FastestWayToReadATextFileInCsharp/FastestWayToReadATextFileInCsharp/WaysToReadATextFileInCsharp.cs
```diff
@@ -0,0 +1,86 @@
...
+        {
+            var stringBuilder = new StringBuilder();
+
+            using (var fileStream = new FileStream(_sampleFilePath, FileMode.Open))
+            using (var bufferedStream = new BufferedStream(fileStream))
+            using (var streamReader = new StreamReader(bufferedStream))
+            {
+                string fileLine;
+
+                while ((fileLine = streamReader.ReadLine()) != null)
```
karimkod, line 76: We should include this in the while definition. 
```suggestion
                while (streamReader.ReadLine() is { } fileLine)
```

files-csharp/FastestWayToReadATextFileInCsharp/FastestWayToReadATextFileInCsharp/WaysToReadATextFileInCsharp.cs
```diff
@@ -0,0 +1,86 @@
...
+            return string.Join(Environment.NewLine, stringArray);
+        }
+
+        public string UseStreamReaderReadLine()
+        {
+            using var streamReader = new StreamReader(_sampleFilePath);
+            var stringBuilder = new StringBuilder();
+            string fileLine;
+
+            while ((fileLine = streamReader.ReadLine()) != null)
```
karimkod, line 36: We should include this in the while definition.

karimkod (changes requested):

files-csharp/FastestWayToReadATextFileInCsharp/FastestWayToReadATextFileInCsharp/WaysToReadATextFileInCsharp.cs
jshergal: @karimkod - If we leave this with the `AppendLine` then we have the potential issue of text fidelity between the line endings in the file being read and the line endings in the result. This may be fine under some use cases, but it for sure needs to be pointed out in the article that there is a potential loss of fidelity.

karimkod: @jshergal - the `StreamReader.ReadLine` [doc ](https://learn.microsoft.com/en-us/dotnet/api/system.io.streamreader.readline?view=net-7.0#remarks) states that : 

> A line is defined as a sequence of characters followed by a line feed ("\n"), a carriage return ("\r"), or a carriage return immediately followed by a line feed ("\r\n"). **The string that is returned does not contain the terminating carriage return or line feed. The returned value is null if the end of the input stream is reached.**

The line endings in the file won't be read. Would the risk still be present?

karimkod:
> One thing to note and be careful about is in the code where we are using `StreamReader.ReadLine`. `ReadLine` strips the new line characters from the returned string. In our sample code we are then using `StringBuilder.AppendLine` to add those lines to our final string. This can have the side effect of returning a different string than the one that exists in the file as `AppendLine` is going to use the whatever `Environment.NewLine` uses. If the file, for example contained only '\r' as the end of line character, but our current running system had '\r\n', then we are not returning a true representation of the text. That may be acceptable in some cases, but we should be careful to point that out. This also applies in the `File.ReadAllLines` and `File.ReadLines` code where we are also using `Environment.NewLine`.

@cjaynjoku please take this in consideration and add a disclaimer when applicable in the article. 
Thanks, @jshergal, for the thorough explanation, you can ignore my latest comments :)

files-csharp/FastestWayToReadATextFileInCsharp/FastestWayToReadATextFileInCsharp/WaysToReadATextFileInCsharp.cs
karimkod: @jshergal, To fix the TrimEnd issue, we can fix use Append and append the NewLine character manually after each line, this will give us more control on when to put the NewLine and we can avoid that extra NewLine that AppendLine does at the end of the file and we subsequently avoid the TrimEnd(); 
Do you see any flaws in this approach? apart from the potential end of line fidelity loss?

cjaynjoku:
> > One thing to note and be careful about is in the code where we are using `StreamReader.ReadLine`. `ReadLine` strips the new line characters from the returned string. In our sample code we are then using `StringBuilder.AppendLine` to add those lines to our final string. This can have the side effect of returning a different string than the one that exists in the file as `AppendLine` is going to use the whatever `Environment.NewLine` uses. If the file, for example contained only '\r' as the end of line character, but our current running system had '\r\n', then we are not returning a true representation of the text. That may be acceptable in some cases, but we should be careful to point that out. This also applies in the `File.ReadAllLines` and `File.ReadLines` code where we are also using `Environment.NewLine`.
> 
> @cjaynjoku please take this in consideration and add a disclaimer when applicable in the article. Thanks, @jshergal, for the thorough explanation, you can ignore my latest comments :)

@karimkod I've added this disclaimer to the first method where the AppendLine is used, should I add it everywhere else?

files-csharp/FastestWayToReadATextFileInCsharp/FastestWayToReadATextFileInCsharp/WaysToReadATextFileInCsharp.cs
jshergal: @karimkod - That approach would work, with the caveat of the potential loss of fidelity. I had posed the question as I didn't run the code to check exactly what we were trimming. Having looked at it a few times now, I am confident that we were doing this to remove the final newline that was added by the AppendLine. With that being the case, we could also take the approach of removing it from the StringBuilder buffer before calling to string. Something like this (I didn't test this and it might have an off by one error as I am just freehanding this - also maybe it could be simplified a little)
`stringBuilder.Remove(stringBuilder.Length - Environment.NewLine.Length - 1, Environment.NewLine.Length);`

files-csharp/FastestWayToReadATextFileInCsharp/FastestWayToReadATextFileInCsharp/WaysToReadATextFileInCsharp.cs
```diff
@@ -1,86 +1,128 @@
...
+        return stringBuilder.ToString();
+    }
 
-            return string.Join(Environment.NewLine, stringArray);
-        }
+    public string UseFileReadAllText() => File.ReadAllText(_sampleFilePath);
+    
+    public string UseFileReadLines()
+    {
+        var lines = File.ReadLines(_sampleFilePath);
```
jshergal, line 32: There is no need for this variable, the `File.ReadLines` call can be made in the `foreach`

files-csharp/FastestWayToReadATextFileInCsharp/FastestWayToReadATextFileInCsharp/WaysToReadATextFileInCsharp.cs
```diff
@@ -1,86 +1,128 @@
...
-                stringBuilder.Append(buffer, 0, numberRead);
-            }
+        return stringBuilder.ToString();
+    }
 
-            return stringBuilder.ToString();
+    public string UseStreamReaderReadBlockWithSpan()
+    {
+        using var streamReader = new StreamReader(_sampleFilePath);
+        var buffer = new char[4096].AsSpan();
```
jshergal, line 83: I should have mentioned this last time, but both here and the method above where we are allocating the buffer, we can gain some potential performance improvement, and certainly memory improvement by switching to an array from the Array pool.

files-csharp/FastestWayToReadATextFileInCsharp/Tests/WaysToReadATextFileInCsharpTests.cs
```diff
@@ -1,81 +1,94 @@
...
-            = """
+    private const string ExpectedText
+        = """
             Integer facilisis ex libero, ut suscipit leo blandit non. Vivamus nec ipsum orci.
             Proin nec mauris dui. Proin at felis et eros commodo aliquet.
             Pellentesque lacinia porta leo, non accumsan turpis sagittis et.
             Vestibulum ante ipsum primis in faucibus orci luctus et ultrices posuere.
             """;
+    
+    private static readonly string _tempFilePath = Path.GetTempFileName();
```
jshergal, line 15: It is good practice to clean up after ourselves, so we should try to ensure we remove this temp file when the test is done, so we can just remove the static here and implement IDisposable to delete the file.

files-csharp/FastestWayToReadATextFileInCsharp/FastestWayToReadATextFileInCsharp/WaysToReadATextFileInCsharp.cs
karimkod: @jshergal From documentation : 

> Using the [ArrayPool<T>](https://learn.microsoft.com/en-us/dotnet/api/system.buffers.arraypool-1?view=net-7.0) class to rent and return buffers (using the [Rent](https://learn.microsoft.com/en-us/dotnet/api/system.buffers.arraypool-1.rent?view=net-7.0) and [Return](https://learn.microsoft.com/en-us/dotnet/api/system.buffers.arraypool-1.return?view=net-7.0) methods) can improve performance in situations where arrays are created and destroyed frequently, resulting in significant memory pressure on the garbage collector.

I think here, we are not allocating and destorying frequently the array, do you think it is a good usage for the ArrayPool?

files-csharp/FastestWayToReadATextFileInCsharp/FastestWayToReadATextFileInCsharp/WaysToReadATextFileInCsharp.cs
```diff
@@ -0,0 +1,128 @@
...
+        var lines = File.ReadLines(_sampleFilePath);
+        var stringBuilder = new StringBuilder();
+
+        foreach(var line in lines)
+        {
+            stringBuilder.AppendLine(line);
+        }
+        stringBuilder.Length -= Environment.NewLine.Length;
+
+        return stringBuilder.ToString();
```
karimkod, line 41: As this is repeated throughout the code many times, can we put it in a separate method and call it when appropriate?

files-csharp/FastestWayToReadATextFileInCsharp/Tests/WaysToReadATextFileInCsharpTests.cs
karimkod: Please also fix the naming convention of the _tempFilePath to be of a static readonly string variable.

files-csharp/FastestWayToReadATextFileInCsharp/FastestWayToReadATextFileInCsharp/WaysToReadATextFileInCsharp.cs
```diff
@@ -0,0 +1,128 @@
...
+        using var streamReader = new StreamReader(bufferedStream);
+
+        while (streamReader.ReadLine() is { } fileLine)
+        {
+            stringBuilder.AppendLine(fileLine);
+        }
+        stringBuilder.Length -= Environment.NewLine.Length;
+
+        return stringBuilder.ToString();
+    }
```
karimkod, line 127: I think this code is silghtly different than the one in the article, can you please update the article and make sure all the article in the code matches the code in the repo?

karimkod (changes requested):

files-csharp/FastestWayToReadATextFileInCsharp/FastestWayToReadATextFileInCsharp/WaysToReadATextFileInCsharp.cs
jshergal: @karimkod - My suggestion was in light of the benchmark and in light of the potential usage of such a block of code. Since what we are talking about in this article is the most "performant" reading of text files, we are already dealing with the use case where we are trying to squeeze out some performance related to the reading of text files. If we want to keep what we already have and add a method that uses the ArrayPool, then we can showcase what performance gain (if any) there is, but more importantly can show the memory overhead improvement. Remeber in the case of a benchmark we are actually creating and allocating arrays frequently due to the nature of a benchmark which is executing this code multiple times to compute an average time and average memory usage.
So, that was what I had in mind with the suggestion. No pressure on the implementing. I was just adding the thought as we are focusing on performance, and this is another way to potentially squeeze out a little bit more.

karimkod: @jshergal thanks for the insight. I see the point clearly now. 
@cjaynjoku could you please add this as another method of the benchmark. You can put it in the same section as the one that uses the array in the article (no need for an extra section). And include it in the benchmark section as well. 
And also, if you could highlight the memory usage as @jshergal suggested above. 
Thanks for both of you!

files-csharp/FastestWayToReadATextFileInCsharp/FastestWayToReadATextFileInCsharp/WaysToReadATextFileInCsharp.cs
cjaynjoku: How do I show this in the article without breaking the flow of the article?

files-csharp/FastestWayToReadATextFileInCsharp/FastestWayToReadATextFileInCsharp/WaysToReadATextFileInCsharpBenchmark.cs
```diff
@@ -0,0 +1,54 @@
...
+using BenchmarkDotNet.Order;
+
+namespace FastestWayToReadATextFileInCsharp;
+
+[MemoryDiagnoser]
+[Orderer(SummaryOrderPolicy.FastestToSlowest)]
+public class WaysToReadATextFileInCsharpBenchmark
+{
+    private static readonly string SampleFilePath
+       = Path.Combine(Directory.GetParent(Directory.GetCurrentDirectory()).Parent.Parent.Parent.Parent.Parent.Parent.FullName, "BenchmarkTextFile.txt");
```
karimkod, line 11: Is this needed, why can't we just put the name of the file "BenchmarkTextFile.txt" in the variable, and include the file in the build folder?

karimkod (changes requested):

files-csharp/FastestWayToReadATextFileInCsharp/FastestWayToReadATextFileInCsharp/WaysToReadATextFileInCsharpBenchmark.cs
cjaynjoku: Whenever I try this, I get the FileNotFoundException with this exception message:

`System.IO.FileNotFoundException: Could not find file 'C:\Users\Januarius\Desktop\CodeMazeGuides\files-csharp\FastestWayToReadATextFileInCsharp\FastestWayToReadATextFileInCsharp\bin\Release\net7.0\29df687e-8207-4cf8-970a-abac7154797d\bin\Release\net7.0\BenchmarkTextFile.txt'`

From this message, the benchmark file should be in that build folder, which I cannot locate on my PC.
It also means I have to push the bin\Release folder too.

I followed this approach because it was suggested by Marinko in a previous article I worked on: https://github.com/CodeMazeBlog/CodeMazeGuides/blob/main/json-csharp/ReadAndParseAJSONFileInCSharp/ReadAndParseAJSONFileInCSharp/ReadAndParseJSONFileMethodsBenchMark.cs

jshergal: @cjaynjoku and @karimkod - another potential option would be to use the `[GlobalSetup]` and `[GlobalCleanup]` methods and write the text out to a temporary file. We could create the file in the setup method and delete it in the cleanup method.

cjaynjoku: I have chosen the option above and it works perfectly. This article has dragged on for over a month. Thank you.

jshergal:
@cjaynjoku - I think you meant to push that change into another branch.

PR #1352, opened 2023-08-17
CM-191: How to Perform Case-Insensitive Substring Search

ahmedryanfaiyaz:


csharp-basic-topics/PerformSubstringSearchWithContainsMethod/PerformSubstringSearchWithContainsMethod/SubstringSearch.cs
```diff
@@ -0,0 +1,111 @@
...
+            // Regular Expression Search
+            RegularExpressionSearchWithNoParameter(sourceString, substringToSearch);
+            RegularExpressionSearchWithRegexOptionsNoneParameter(sourceString, substringToSearch);
+            RegularExpressionSearchWithRegexOptionsIgnoreCaseParameter(sourceString, substringToSearch);
+
+            // Linq With String Equals Method Search
+            LinqWithStringEqualsMethodSearchWithNoParameter(sourceString, substringToSearch, separator);
+            LinqWithStringEqualsMethodSearchWithOrdinalStringComparisonParameter(sourceString, substringToSearch, separator);
+            LinqWithStringEqualsMethodSearchWithOrdinalIgnoreCaseStringComparisonParameter(sourceString, substringToSearch, separator);
+        }
```
jshergal, line 32: This code is basically the console app part and should be split into its own file. We can just add a Program.cs and drop this in there. Since this project targets .NET 7.0, we can drop the `public static void Main` part and just drop in the heart of the method (using top level statements)

csharp-basic-topics/PerformSubstringSearchWithContainsMethod/Tests/StringSearchUnitTest.cs
```diff
@@ -0,0 +1,131 @@
...
+namespace Tests
+{
+    [TestClass]
+    public class StringSearchUnitTest
+    {
+        [TestMethod]
+        public void GivenSourceAndSubstring_WhenUsingStringContainsMethod_ThenShouldReturnContainmentResult()
+        {
+            var sourceString = "Code Maze";
+            var substringToSearch = "maze";
```
jshergal, line 12: Since each test is using the same arguments for sourceString and substringToSearch, rather than local variables in each method, we can make them class level constants.

PR #1356, opened 2023-08-20
CM-317: Basic Authentication with HttpClient

RPM1984:


authorization-dotnet/BasicAuthenticationWithHttpClient/Client/WeatherForecastClient.cs
```diff
@@ -0,0 +1,21 @@
...
+    private readonly HttpClient _client;
+
+    public WeatherForecastClient(HttpClient client, IConfiguration configuration)
+    {
+        _client = client;
+        _client.BaseAddress = new Uri("https://localhost:7003");                
+        var basicAuthenticationUsername = configuration["BasicAuthenticationUsername"];
+        var basicAuthenticationPassword = configuration["BasicAuthenticationPassword"];
+        var basicAuthenticationValue = Convert.ToBase64String(Encoding.ASCII.GetBytes($"{basicAuthenticationUsername}:{basicAuthenticationPassword}"));
+        _client.DefaultRequestHeaders.Authorization = new AuthenticationHeaderValue("Basic", basicAuthenticationValue);
```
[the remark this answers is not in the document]
RPM1984: That was my thinking too. If we are going to create these typed clients, it's better to keep the config close to the client. The only addition i would normally do is use strongly-typed configuration, but as i mentioned in this article it's kind of out of scope. So my suggestion is we leave this, but happy to go with your recommendation.

authorization-dotnet/BasicAuthenticationWithHttpClient/Server/Controllers/WeatherForecastController.cs
```diff
@@ -0,0 +1,72 @@
...
+                            {
+                                return true;
+                            }
+                        }
+                    }
+                }
+            }
+            catch
+            {
+            }
```
CodeMazeBlog, line 67: What's the point of this try-catch? Does it help the flow?

RPM1984: There are some string operations that could throw exceptions, eg param could be invalid base 64 string, substring throwing out of range etc

PR #1358, opened 2023-08-22
CM-801: How to Use StopWatch in C#

martinchege324:
First article and draft.

### DIFF
--- a/dotnet-datetime/StopWatchCSharp/StopWatchCSharp/Program.cs
+++ b/dotnet-datetime/StopWatchCSharp/StopWatchCSharp/Program.cs
@@ -1,0 +1,2 @@
+﻿// See https://aka.ms/new-console-template for more information
+Console.WriteLine("Hello, World!");
--- a/dotnet-datetime/StopWatchCSharp/StopWatchCSharp/Program.cs
+++ b/dotnet-datetime/StopWatchCSharp/StopWatchCSharp/Program.cs
@@ -1,9 +1,11 @@
 ﻿using StopWatchCSharp;
+using System.Diagnostics;
 
-var stopWatchInstance = new StopWatchMethods();
-var stopWatch = stopWatchInstance.CreateRandomArray(1000);
+var stopWatch = StopWatchMethods.CreateRandomArray(1000);
 
 Console.WriteLine($@"Elapsed Time: {stopWatch.Elapsed}
 Elapsed Milliseconds: {stopWatch.ElapsedMilliseconds}
 Elapsed Ticks: {stopWatch.ElapsedTicks}
-Is Running: {stopWatch.IsRunning}");
+Is Running: {stopWatch.IsRunning}
+Frequency: {Stopwatch.Frequency}
+Using High Resolution: {Stopwatch.IsHighResolution}");
--- a/dotnet-datetime/StopWatchCSharp/StopWatchCSharp/Program.cs
+++ b/dotnet-datetime/StopWatchCSharp/StopWatchCSharp/Program.cs
@@ -1,2 +1,9 @@
-﻿// See https://aka.ms/new-console-template for more information
-Console.WriteLine("Hello, World!");
+﻿using StopWatchCSharp;
+
+var stopWatchInstance = new StopWatchMethods();
+var stopWatch = stopWatchInstance.CreateRandomArray(1000);
+
+Console.WriteLine($@"Elapsed Time: {stopWatch.Elapsed}
+Elapsed Milliseconds: {stopWatch.ElapsedMilliseconds}
+Elapsed Ticks: {stopWatch.ElapsedTicks}
+Is Running: {stopWatch.IsRunning}");
--- a/dotnet-datetime/StopWatchCSharp/StopWatchCSharp/StopWatchMethods.cs
+++ b/dotnet-datetime/StopWatchCSharp/StopWatchCSharp/StopWatchMethods.cs
@@ -3,94 +3,19 @@
 namespace StopWatchCSharp
 {
     public class StopWatchMethods
-    {
-        public Stopwatch BubbleSort(int[] numArray, int size)
+    {     
+        public Stopwatch CreateRandomArray(int size)
         {
             var stopWatch = new Stopwatch();
-            
             stopWatch.Start();
-
-            for (int i = 0; i < size - 1; i++)
-                for (int j = 0; j < size - i - 1; j++)
-                    if (numArray[j] > numArray[j + 1])
-                    {
-                        var tempVar = numArray[j];
-                        numArray[j] = numArray[j + 1];
-                        numArray[j + 1] = tempVar;
-                    }
-
-            stopWatch.Stop();
-
-            return stopWatch;
-        }
-
-        public Stopwatch HeapSort(int[] array, int size)
-        {
-            var stopWatch = new Stopwatch();
-
-            stopWatch.Start();
-
-            if (size <= 1) 
-            {
-                stopWatch.Stop();
-
-                return stopWatch;
-            }
-                
-            for (int i = size / 2 - 1; i >= 0; i--)
-            {
-                Heapify(array, size, i);
-            }
-
-            for (int i = size - 1; i >= 0; i--)
-            {
-                var tempVar = array[0];
-                array[0] = array[i];
-                array[i] = tempVar;
-
-                Heapify(array, i, 0);
-            }
-
-            stopWatch.Stop();
-
-            return stopWatch;
-        }
-
-        static void Heapify(int[] array, int size, int index)
-        {
-            var largestIndex = index;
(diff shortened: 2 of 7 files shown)


dotnet-datetime/StopWatchCSharp/StopWatchCSharp/StopWatchMethods.cs
```diff
@@ -0,0 +1,96 @@
...
+            }
+        }
+
+        public int[] CreateRandomArray(int size)
+        {
+            var array = new int[size];
+            var rand = new Random();
+
+            for (int i = 0; i < size; i++)
+                array[i] = rand.Next();
```
jshergal, line 91: We can just use Random.Shared here

dotnet-datetime/StopWatchCSharp/StopWatchCSharpTests/StopWatchUnitTests.cs
```diff
@@ -0,0 +1,72 @@
...
+        }
+
+        [TestMethod]
+        public void GivenBubbleSortAlgorithm_WhenInvoked_VerifyStopwatchSuccess()
+        {
+            var stopWatch = _stopWatchMethods.BubbleSort(_numArray, _numArray.Length);
+
+            Assert.IsInstanceOfType(stopWatch, typeof(Stopwatch));
+            Assert.IsInstanceOfType(stopWatch.Elapsed, typeof(TimeSpan));
+            Assert.IsFalse(stopWatch.IsRunning);
```
jshergal, line 23: The only meaningful assert here is the IsFalse(stopWatch.IsRunning). We could add an additional assert to ensure the stopwatch elapsed time is not zero after calling the sort method. That would be meaningful in the context of the test.

dotnet-datetime/StopWatchCSharp/StopWatchCSharpTests/StopWatchUnitTests.cs
```diff
@@ -0,0 +1,72 @@
...
+
+            stopWatch.Reset();
+
+            Assert.AreEqual(stopWatch.Elapsed, TimeSpan.Zero);
```
jshergal, line 27: This can be removed.

dotnet-datetime/StopWatchCSharp/StopWatchCSharpTests/StopWatchUnitTests.cs
```diff
@@ -0,0 +1,72 @@
...
+        }
+
+        [TestMethod]
+        public void GivenHeapSortAlgorithm_WhenInvoked_VerifyStopwatchSuccess()
+        {
+            var stopWatch = _stopWatchMethods.HeapSort(_numArray, _numArray.Length);
+            
+            Assert.IsInstanceOfType(stopWatch, typeof(Stopwatch));
+            Assert.IsInstanceOfType(stopWatch.Elapsed, typeof(TimeSpan));
```
jshergal, line 36: Can be removed

dotnet-datetime/StopWatchCSharp/StopWatchCSharpTests/StopWatchUnitTests.cs
```diff
@@ -0,0 +1,72 @@
...
+            Assert.IsFalse(stopWatch.IsRunning);
+
+            stopWatch.Reset();
+
+            Assert.AreEqual(stopWatch.Elapsed, TimeSpan.Zero);
+        }
+
+        [TestMethod]
+        public void GivenHeapSortAlgorithm_WhenInvoked_VerifyStopwatchSuccess()
```
jshergal, line 31: Given that the name of this test method is similar to the previous method, one would expect the assertions and operations to be similar, so we can remove the restarting of the stopwatch and other operations. Basically since they both are testing that a stopwatch with some kind of elapsed time, that is no longer running, was returned from the sort method, the assertions should be focused on that in both cases.

dotnet-datetime/StopWatchCSharp/StopWatchCSharpTests/StopWatchUnitTests.cs
```diff
@@ -0,0 +1,72 @@
...
+
+        [TestMethod]
+        public void GivenTwoSortingAlgorithm_WhenInvoked_VerifyStopwatchSuccess()
+        {
+            var heapSort = _stopWatchMethods.HeapSort(_numArray, _numArray.Length);
+            var bubbleSort = _stopWatchMethods.BubbleSort(_numArray, _numArray.Length);
+
+            Assert.IsFalse(heapSort.Equals(bubbleSort));
+            Assert.IsInstanceOfType(heapSort.ElapsedMilliseconds, typeof(long));
+            Assert.IsInstanceOfType(bubbleSort.ElapsedTicks, typeof(long));
```
jshergal, line 57: These can be removed

dotnet-datetime/StopWatchCSharp/StopWatchCSharp/Program.cs
jshergal, line 1: We should call the two sorting methods here and print the elapsed time.

dotnet-datetime/StopWatchCSharp/StopWatchCSharp/StopWatchMethods.cs
```diff
@@ -0,0 +1,96 @@
...
+            {
+                var tempVar = array[index];
+                array[index] = array[largestIndex];
+                array[largestIndex] = tempVar;
+
+                Heapify(array, size, largestIndex);
+            }
+        }
+
+        public int[] CreateRandomArray(int size)
```
jshergal, line 85: This should move to some kind of Utility class rather then being here. It is not a StopWatch method. I would have suggested just moving it to the test class, but we should probably also call it in the Program.cs.

dotnet-datetime/StopWatchCSharp/StopWatchCSharp/StopWatchMethods.cs
```diff
@@ -0,0 +1,96 @@
...
+                array[0] = array[i];
+                array[i] = tempVar;
+
+                Heapify(array, i, 0);
+            }
+
+            stopWatch.Stop();
+
+            return stopWatch;
+        }
```
jshergal: We should create a master method that initializes, starts and stops the `StopWatch` and pass in the method that we want to time. It would probably be best to create a new class called `SortingMethods` that contains these two sorts. Then this class can be called something like `ExecutionTimeMeasurer` or something like that. It would have a single method that takes a sort `Action` and the arguments to call the sort methods. Then internally it would create, start and stop the `StopWatch` timing the execution of the sort. It could be even more generic and we could call it `BenchmarkTimer` or something, which would take just an `Action` to execute. Then the caller could pass in whatever they want as a lambda or whatever.

jshergal:
You can ignore that last comment. I thought I had added it, but then hadn't finished it (got distracted with a phone call). By the time I added it, the code had changed and is much simpler now 😀

dotnet-datetime/StopWatchCSharp/StopWatchCSharp/StopWatchMethods.cs
```diff
@@ -0,0 +1,21 @@
+﻿using System.Diagnostics;
+
+namespace StopWatchCSharp
```
mahrncic, line 3: Use file-scoped namespaces 😄

mahrncic (changes requested):

mahrncic (approved):

dotnet-datetime/StopWatchCSharp/StopWatchCSharp/Program.cs
```diff
@@ -0,0 +1,11 @@
...
+using System.Diagnostics;
+
+var stopWatch = StopWatchMethods.CreateRandomArray(1000);
+
+Console.WriteLine($@"Elapsed Time: {stopWatch.Elapsed}
+Elapsed Milliseconds: {stopWatch.ElapsedMilliseconds}
+Elapsed Ticks: {stopWatch.ElapsedTicks}
+Is Running: {stopWatch.IsRunning}
+Frequency: {Stopwatch.Frequency}
+Using High Resolution: {Stopwatch.IsHighResolution}");
```
mahrncic, line 11: Why are these two `Stopwatch` instances starting with an uppercase letter why all of the previous instances are spelled like `stopWatch`?

martinchege324: @mahrncic Those are static read-only fields that are part of the Stopwatch class: https://learn.microsoft.com/en-us/dotnet/api/system.diagnostics.stopwatch.frequency?view=net-7.0. Therefore, we use Stopwatch.IsHighResolution to fetch its value.

mahrncic (approved):

PR #1366, opened 2023-08-26
CM-794: Volatile vs Interlocked vs lock in C#

lasaetarubia:


csharp-advanced-topics/VolatileInterlockedLockInCsharp/VolatileInterlockedLockInCsharp/Program.cs
```diff
@@ -0,0 +1 @@
+﻿Console.WriteLine("Hello World");
```
jshergal, line 1: We should be exercising the code here, showing some examples in the console app.

csharp-advanced-topics/VolatileInterlockedLockInCsharp/VolatileInterlockedLockInCsharp/Account.cs
```diff
@@ -0,0 +1,70 @@
...
+            lock (_lock)
+            {
+                _balanceLock -= amount;
+            }
+        }
+
+        public void WithdrawInterlocked(int amount)
+        {
+            Interlocked.Exchange(ref _balanceInterlocked, _balanceInterlocked - amount);
+        }
```
jshergal, line 68: While this may be hard to reproduce, especially the way the code is written, this code here has a race condition. The interlocked exchange will guarantee atomic write access to the `ref` value, but we are still reading the `_balanceInterlocked` value without any synchronization. If you look at the JIT output of that call, you will see an `lea` call on the `_balanceInterlocked` value. Followed by `add` and finally the `xchg` call, which is a full fenced atomic operation.
We can solve this by using `Interlocked.Add(ref _balanceInterlocked, -amount)`

jshergal (changes requested):

csharp-advanced-topics/VolatileInterlockedLockInCsharp/VolatileInterlockedLockInCsharp/Account.cs
```diff
@@ -7,6 +7,7 @@ public class Account
         private int _balanceLock;
         private int _balanceInterlocked;
         private object _lock = new object();
+        private readonly int _delay = new Random().Next(0, 100);
```
jshergal, line 10: Use Random.Shared here

jshergal:
I added a couple of comments on the code. I am adding this as it is not something you have to do, but if you take a look at each of the tests, there is a lot of duplicate code. We could probably extract that into a method that we call from within each unit test.

csharp-advanced-topics/VolatileInterlockedLockInCsharp/VolatileInterlockedLockInCsharp/Account.cs
jshergal: I think you are still misunderstanding my statement about the random delay. What I was trying to indicate is that we should compute a different amount of delay in each Task. That is this line here:
`Thread.Sleep(_delay);`
using my suggestion would become:
`Thread.Sleep(Random.Shared.Next(50,300);`
That way we have a little more "real world" wobble in the threads. Each one taking a different amount of time to execute.

csharp-advanced-topics/VolatileInterlockedLockInCsharp/VolatileInterlockedLockInCsharp/Program.cs
jshergal: This still needs to be addressed

jshergal (changes requested):

csharp-advanced-topics/VolatileInterlockedLockInCsharp/VolatileInterlockedLockInCsharp/Account.cs
lasaetarubia: Understood. Will make the necessary change.

csharp-advanced-topics/VolatileInterlockedLockInCsharp/VolatileInterlockedLockInCsharp/Account.cs
```diff
@@ -0,0 +1,70 @@
...
+                {
+                    _balanceLock = value;
+                }
+            }
+        }
+
+        public int BalanceInterlocked
+        {
+            get => _balanceInterlocked;
+            set => _balanceInterlocked = value;
```
jshergal, line 44: For consistency with the other properties and code, we should be using the Interlocke statements here. We are using lock in the BalanceLock case, so we should "mirror" that style of behavior here.
So we would use `CompareExchange()` for the get and `Exchange()` for the set.

jshergal (changes requested):

csharp-advanced-topics/VolatileInterlockedLockInCsharp/VolatileInterlockedLockInCsharp/Account.cs
lasaetarubia: Done

csharp-advanced-topics/VolatileInterlockedLockInCsharp/Tests/Tests.cs
```diff
@@ -0,0 +1,79 @@
+using VolatileInterlockedLockInCsharp;
+
+namespace Tests
```
mahrncic, line 3: Please use file-scoped namespaces. In that way, our code looks cleaner since there is a bit less indentation in it.

csharp-advanced-topics/VolatileInterlockedLockInCsharp/VolatileInterlockedLockInCsharp/Account.cs
```diff
@@ -0,0 +1,70 @@
+﻿namespace VolatileInterlockedLockInCsharp
```
mahrncic, line 1: file-scoped

mahrncic (changes requested):

csharp-advanced-topics/VolatileInterlockedLockInCsharp/Tests/Tests.cs
lasaetarubia: Done

csharp-advanced-topics/VolatileInterlockedLockInCsharp/VolatileInterlockedLockInCsharp/Account.cs
lasaetarubia: Done

mahrncic (approved):

PR #1370, opened 2023-08-27
CM-803: How to Read a String From a .Resx File in C#

JPetronilhoNunes:


jshergal:
Retest this please

strings-csharp/ReadStringFromResourceFile/ReadStringFromResourceFile/Classes/ResourcesManager.cs
```diff
@@ -0,0 +1,57 @@
+﻿using System.Globalization;
+using System.Reflection;
+using System.Resources;
+
+namespace ReadStringFromResourceFile;
+
+public class ResourcesManager
+{
+    private Assembly? assembly;
```
jshergal, line 9: This field is only assigned to and never used, it can be removed

strings-csharp/ReadStringFromResourceFile/ReadStringFromResourceFile/Classes/ResourcesManager.cs
```diff
@@ -0,0 +1,57 @@
...
+    }
+
+    public ResourceItem<T> GetResource<T>(string key)
+    {
+        try
+        {
+            var obj = resources.GetObject(key, culture);
+
+            if (obj is T)
+                return new ResourceItem<T>((T)obj);
```
jshergal, line 48: We can merge the cast with the type check here.

strings-csharp/ReadStringFromResourceFile/ReadStringFromResourceFile/Classes/ResourcesManager.cs
```diff
@@ -0,0 +1,57 @@
...
+            else
```
jshergal, line 49: The else is redundant. We can remove it and reduce the nesting. We just need to add a blank line

strings-csharp/ReadStringFromResourceFile/ReadStringFromResourceFile/Classes/ResourcesManager.cs
```diff
@@ -0,0 +1,57 @@
...
+    public string? GetString(string key)
+    {
+        try
+        {
+            var resource = GetResource<string>(key);
+            
+            if (!resource.IsValid)
+                return string.Empty;
+
+            return resource.Item;
```
jshergal, line 33: We can simplify this with the ternary operator

strings-csharp/ReadStringFromResourceFile/ReadStringFromResourceFile/Program.cs
```diff
@@ -0,0 +1,10 @@
+﻿using ReadStringFromResourceFile;
+
+ResourcesManager rmEnglish = new ResourcesManager(@"ReadStringFromResourceFile.Resources.Texts.English");
+Console.WriteLine(rmEnglish.GetString("GREETINGS_TEXT"));
+
+var logoff = rmEnglish.GetString("LOGOFF_2");
```
jshergal, line 6: We should probably at least display this string, otherwise we can change this to just be a discard

strings-csharp/ReadStringFromResourceFile/ReadStringFromResourceFile/Program.cs
```diff
@@ -0,0 +1,10 @@
+﻿using ReadStringFromResourceFile;
+
+ResourcesManager rmEnglish = new ResourcesManager(@"ReadStringFromResourceFile.Resources.Texts.English");
```
jshergal, line 3: Use var

strings-csharp/ReadStringFromResourceFile/ReadStringFromResourceFile/Program.cs
```diff
@@ -0,0 +1,10 @@
...
+Console.WriteLine(rmEnglish.GetString("GREETINGS_TEXT"));
+
+var logoff = rmEnglish.GetString("LOGOFF_2");
+
+ResourcesManager rmPortuguese = new ResourcesManager(@"ReadStringFromResourceFile.Resources.Texts.Portuguese");
```
jshergal, line 10: Use var

strings-csharp/ReadStringFromResourceFile/ReadStringFromResourceFile/Classes/ResourcesManager.cs
```diff
@@ -0,0 +1,57 @@
...
+    }
+
+    public string? GetString(string key)
+    {
+        try
+        {
+            var resource = GetResource<string>(key);
+            
+            if (!resource.IsValid)
+                return string.Empty;
```
jshergal, line 31: It seems inconsistent that we return empty string if the resource is not valid, but in an exception return null. I might be missing something here, so just want to mention it to think about it, whether this should be the case or not.

strings-csharp/ReadStringFromResourceFile/ReadStringFromResourceFile/Classes/ResourceItem.cs
```diff
@@ -0,0 +1,19 @@
+﻿namespace ReadStringFromResourceFile;
+public class ResourceItem<T>
```
jshergal, line 2: We can seal this class

strings-csharp/ReadStringFromResourceFile/ReadStringFromResourceFile/Classes/ResourceItem.cs
```diff
@@ -0,0 +1,19 @@
...
+{
+    private T? item;
```
jshergal: Can be readonly and naming convention should have leading _

strings-csharp/ReadStringFromResourceFile/ReadStringFromResourceFile/Classes/ResourcesManager.cs
JPetronilhoNunes: Should had been:
 this.assembly = assembly ?? Assembly.GetExecutingAssembly();
on the following line, I forget that.
I´ve renamed all private fields with prefix underscore to get rid of this issue.
Also, in the Program.cs, I changed a little bit, now it shows an error for the LOGOFF_2 resource, which doesn´t exist. I´m going to update the article also.

strings-csharp/ReadStringFromResourceFile/ReadStringFromResourceFile/Classes/ResourceItem.cs
```diff
@@ -0,0 +1,19 @@
...
+
+    public ResourceItem(T? item = default(T))
```
jshergal, line 6: We can simplify to just `T? item = default`

strings-csharp/ReadStringFromResourceFile/ReadStringFromResourceFile/Classes/ResourceItem.cs
```diff
@@ -0,0 +1,19 @@
...
+    {
+        this.item = item;
```
jshergal: No need for this once we correct the naming convention issue

strings-csharp/ReadStringFromResourceFile/ReadStringFromResourceFile/Classes/ResourceItem.cs
```diff
@@ -0,0 +1,19 @@
...
+    }
+
+    public T? Item { get => item; }
```
jshergal: No need for the `get` here, we can just use the expression body syntax like we are doing on IsValid

strings-csharp/ReadStringFromResourceFile/ReadStringFromResourceFileTests/Tests.cs
```diff
@@ -0,0 +1,34 @@
...
+    public class Tests 
+    {
+        ResourcesManager rmEnglish;
+        ResourcesManager rmPortuguese;
+
+        public Tests()
+        {
+            rmEnglish = new ResourcesManager(@"ReadStringFromResourceFile.Resources.Texts.English");
+            rmPortuguese = new ResourcesManager(@"ReadStringFromResourceFile.Resources.Texts.Portuguese");
+        }
```
jshergal, line 16: These are not following naming conventions. Also, this can be simplified by converting these statements into member initializers and removing the constructor altogether

jshergal (changes requested):

jshergal:
The build is failing because the file Test.cs is saved with an Ansi encoding and contains high ansi characters. You will need to fix this in order for the test to pass.

JPetronilhoNunes:
Thank you for your comments, I´m going to review them all.

jshergal:
@JPetronilhoNunes - I checked the Program.cs and it is UTF8, I would assume the other files are as well, but somehow the Test.cs file is saved as Ansi. You just need to convert it to UTF8 and then be sure that the text strings are correct and don't need to be reentered. The resource files are UTF8 so you could safely copy the string from there.

JPetronilhoNunes:
Hi Jeff,

I´ve reviewed all the issues you reported. 
You were right, the Tests.cs was Ansi encoded, must have been when introducing Portuguese chars. Everything is OK now.
Looking forward for your feedback.

strings-csharp/ReadStringFromResourceFile/ReadStringFromResourceFile/Classes/ResourceItem.cs
```diff
@@ -1,15 +1,22 @@
 ﻿namespace ReadStringFromResourceFile;
 public class ResourceItem<T>
 {
-    private T? item;
+    private readonly string? _errorMessage;
+    private readonly T? _item;
 
-    public ResourceItem(T? item = default(T))
+    public ResourceItem(T? item = default)
```
jshergal, line 7: With the addition of the `errorMessage` constructor, maybe it makes sense to not have a default value here and to make calling one of the two constructors a requirement. I am assuming the default was here to indicate the invalid state, but now we have an explicit error state constructor.

jshergal (changes requested):
Once again there were several small things that your IDE should be pointing out to you (unused variables, etc). Let the IDE help you and be sure to investigate when there are squiggles, etc. Whether you are using VS, VS Code, JetBrains Rider, etc, each of those have tools that point out those small things and often have quick key combos that fix it for you.

strings-csharp/ReadStringFromResourceFile/ReadStringFromResourceFile/Program.cs
```diff
@@ -1,12 +1,12 @@
 ﻿using ReadStringFromResourceFile;
 
-ResourcesManager rmEnglish = new ResourcesManager(@"ReadStringFromResourceFile.Resources.Texts.English");
+ResourcesManager rmEnglish = new(@"ReadStringFromResourceFile.Resources.Texts.English");
```
jshergal, line 3: `var` is the preferred way of writing these statements with the explicit constructor.

strings-csharp/ReadStringFromResourceFile/ReadStringFromResourceFileTests/Tests.cs
```diff
@@ -6,8 +6,8 @@ namespace Tests
     [TestClass]
     public class Tests 
     {
-        ResourcesManager rmEnglish = new ResourcesManager(@"ReadStringFromResourceFile.Resources.Texts.English");
-        ResourcesManager rmPortuguese = new ResourcesManager(@"ReadStringFromResourceFile.Resources.Texts.Portuguese");
+        readonly ResourcesManager rmEnglish = new(@"ReadStringFromResourceFile.Resources.Texts.English");
+        readonly ResourcesManager rmPortuguese = new(@"ReadStringFromResourceFile.Resources.Texts.Portuguese");
```
jshergal, line 10: Please follow naming conventions for class member fields.

strings-csharp/ReadStringFromResourceFile/ReadStringFromResourceFileTests/Tests.cs
```diff
@@ -24,5 +24,21 @@ public void GivenPortugueseResourceFileHandler_WhenRunIsCalled_ThenResourceStrin
 
             Assert.AreEqual("Olá, como está?", greetingsText);
         }
+
+        [TestMethod]
+        public void GivenPortugueseResourceFileHandler_WhenInstantiatingInvalidResourceBasePath_ThenSetIsValidAsFalse()
+        {
+            ResourcesManager localResourceFile = new(@"ReadStringFromResourceFile.Resources.Texts.Portuguese_Invalid");
```
jshergal, line 31: var with explicit constructor is preferred way of writing this

jshergal (changes requested):

JPetronilhoNunes:
All changes are done. 
Looking forward for your feedback.

strings-csharp/ReadStringFromResourceFile/ReadStringFromResourceFile/Classes/ResourceItem.cs
```diff
@@ -0,0 +1,26 @@
+﻿namespace ReadStringFromResourceFile;
+public sealed class ResourceItem<T>
```
mahrncic, line 2: I don't think we need this file at all for this article. Adding generics over-complicates it and we should focus on getting data from the resource file in string formats.

strings-csharp/ReadStringFromResourceFile/ReadStringFromResourceFile/Classes/ResourcesManager.cs
```diff
@@ -0,0 +1,62 @@
+﻿using System.Globalization;
+using System.Reflection;
+using System.Resources;
+
+namespace ReadStringFromResourceFile;
+
+public class ResourcesManager
```
mahrncic, line 7: This class is too complicated for the scope of this article. Also, the name is too similar to the built-in `ResourceManager` class. Why not use a different name such as `ResourcesUtils`?

Also, it could be a static class with only two methods, for example:

- `public string GetWithResourceManager(string key)`
- `public string GetDirectly(string key)`

Both methods could be static

mahrncic (changes requested):
Requested some changes for simplifying the approach

JPetronilhoNunes:
Hi Marko,

I agree that the code introduces some concepts such as generics, but it is also part of the way I explained good practices. 
As you might already have seen from the article, I explain that we shouldn´t return a null value, which might happen in cases where the resource requested is invalid.
I don´t want, of course, to overcomplicate the article. If you really think changing it would be better, I´ll proceed.

mahrncic:
Hey @JPetronilhoNunes,
While I can see what you are trying to achieve here, and would probably agree on some other types of article, I would like to go with a straightforward path in this one, since it is a mini-article and the concept is pretty simple, so let's go with the simplest approach 😄

strings-csharp/ReadStringFromResourceFile/ReadStringFromResourceFileTests/Tests.cs
```diff
@@ -0,0 +1,28 @@
+using Microsoft.VisualStudio.TestTools.UnitTesting;
+using System.Resources;
+
+namespace Tests
+{
+    [TestClass]
+    public class Tests 
+    {
+        const string HELLO_TEXT = "Hello, how are you?";
+        readonly ResourceManager _rmEnglish = new ResourceManager(@"ReadStringFromResourceFile.Resources.Texts.English", typeof(ReadStringFromResourceFile.Resources.Texts.English).Assembly);
```
mahrncic, line 10: Formatting doesn't seem fine here.

strings-csharp/ReadStringFromResourceFile/ReadStringFromResourceFileTests/Tests.cs
```diff
@@ -0,0 +1,28 @@
+using Microsoft.VisualStudio.TestTools.UnitTesting;
+using System.Resources;
+
+namespace Tests
+{
```
mahrncic, line 5: We should use file-scoped namespaces.

strings-csharp/ReadStringFromResourceFile/ReadStringFromResourceFileTests/Tests.cs
```diff
@@ -0,0 +1,28 @@
...
+    [TestClass]
+    public class Tests 
+    {
+        const string HELLO_TEXT = "Hello, how are you?";
+        readonly ResourceManager _rmEnglish = new ResourceManager(@"ReadStringFromResourceFile.Resources.Texts.English", typeof(ReadStringFromResourceFile.Resources.Texts.English).Assembly);
+
+        [TestMethod]
+        public void GivenEnglishResourceFileHandler_WhenRunIsCalled_ThenResourceStringFetchedIsCorrect()
```
mahrncic, line 13: Why do we have the `WhenRunIsCalled` part in the test method name? Where are we calling `Run`?

mahrncic (changes requested):

jshergal (approved):
@mahrncic the code looks much cleaner now. I haven't looked at the article text so it is still in your hands here, but as far as addressing the things I had pointed out earlier, all looks good.

mahrncic:
@jshergal sounds good, thanks for letting me know.

mahrncic (approved):

PR #1375, opened 2023-08-28
CM-207: Replace Line Breaks in a String C#

koddek:


strings-csharp/ReplaceLineBreaksInAStringCSharp/ReplaceLineBreaksInAStringCSharp/Program.cs
```diff
@@ -0,0 +1,46 @@
...
+        
+        var stringReplaceLineEndingsResult = new ReplaceLineBrake().ReplaceLineBreaksUsingTheStringReplaceLineEndingsMethod();
+        Console.WriteLine(stringReplaceLineEndingsResult);
+        
+        var regularExpressionReplaceResult = new ReplaceLineBrake().ReplaceLineBreaksUsingTheRegularExpressionReplaceMethod();
+        Console.WriteLine(regularExpressionReplaceResult);
+    }
+}
+
+public class ReplaceLineBrake
```
jshergal, line 20: This should be in its own file. Also, we can make it static which will clean up the Program.cs since we will not have to instantiate the class.

strings-csharp/ReplaceLineBreaksInAStringCSharp/ReplaceLineBreaksInAStringCSharp/Program.cs
```diff
@@ -0,0 +1,46 @@
+﻿using System.Text.RegularExpressions;
+
+namespace ReplaceLineBreaksInAStringCSharp;
+
+class Program
+{
+    static void Main(string[] args)
+    {
```
jshergal, line 8: We can use top level statements to reduce the nesting and simplify this

strings-csharp/ReplaceLineBreaksInAStringCSharp/ReplaceLineBreaksInAStringCSharp/Program.cs
```diff
@@ -0,0 +1,46 @@
...
+}
+
+public class ReplaceLineBrake
+{
+    public string ReplaceLineBreaksUsingTheStringReplaceMethod()
+    {
+        var text = "This is a line.\nThis is another line.";
+        var newText = text.Replace("\n", " ");
+        
+        return newText;
```
jshergal, line 27: We can just return the result directly.

strings-csharp/ReplaceLineBreaksInAStringCSharp/ReplaceLineBreaksInAStringCSharp/Program.cs
```diff
@@ -0,0 +1,46 @@
...
+        var regularExpressionReplaceResult = new ReplaceLineBrake().ReplaceLineBreaksUsingTheRegularExpressionReplaceMethod();
+        Console.WriteLine(regularExpressionReplaceResult);
+    }
+}
+
+public class ReplaceLineBrake
+{
+    public string ReplaceLineBreaksUsingTheStringReplaceMethod()
+    {
+        var text = "This is a line.\nThis is another line.";
```
jshergal, line 24: This comment applies across these methods. We should change the method signature to accept a string and then pass in the strings that we are going to replace the endings on. This will make the sample code clearer as well as the unit tests. We can then, in the unit tests, verify that the method does not modify strings that do not contain newlines.

jshergal (changes requested):
We do not need the comments in the unit tests. Also, the expected strings, etc there, can be made constant.

strings-csharp/ReplaceLineBreaksInAStringCSharp/ReplaceLineBreaksInAStringCSharp/Program.cs
koddek: The methods were just meant as a grouping for the code. Changing the method signature to accept the input text complicates the code.

strings-csharp/ReplaceLineBreaksInAStringCSharp/ReplaceLineBreaksInAStringCSharp/ReplaceLineBrake.cs
```diff
@@ -0,0 +1,27 @@
+using System.Text.RegularExpressions;
+
+namespace ReplaceLineBreaksInAStringCSharp;
+
+public static class ReplaceLineBrake
```
elfico, line 5: @koddek 
I think the name you meant here is `ReplaceLineBreak`. This change applies to both the method name and the file name

elfico (changes requested):

elfico (approved):

jshergal (approved):

PR #1376, opened 2023-08-28
CM805: Techniques for Checking Floating-Point Equality

jshergal:


numbers-csharp/FloatingPointEquality/src/FloatingPointEquality.Library/FloatingPointComparisons.cs
```diff
@@ -0,0 +1,97 @@
...
+        1e-11 * 0.5,
+        1e-12 * 0.5,
+        1e-13 * 0.5,
+        1e-14 * 0.5,
+        1e-15 * 0.5,
+        1e-16 * 0.5,
+        1e-17 * 0.5
+    };
+
+
```
fl0tschi, line 30: One line too much

PR #1380, opened 2023-09-01
CM-798 Array Pool

KonstantinosP7:
Array Pool first commit. Please for your review.

### DIFF
--- a/collections-arrays/ArrayPool/ArrayPool.Benchmark/Benchmark.cs
+++ b/collections-arrays/ArrayPool/ArrayPool.Benchmark/Benchmark.cs
@@ -7,22 +7,22 @@ namespace ArrayPool.Benchmark;
 public class Benchmark
 {
     [Params(100, 1000, 10000, 100000)]
-    private int ArraySize { get; set; }
+    public int ArraySize { get; set; }
 
-    private ArrayPool<int>? arrayPool;
+    private ArrayPool<int>? _arrayPool;
 
     [GlobalSetup]
     public void GlobalSetup()
     {
-        arrayPool = ArrayPool<int>.Shared;
+        _arrayPool = ArrayPool<int>.Shared;
     }
 
     [Benchmark]
     public void CreateArrayWithArrayPool()
     {
-        var array = arrayPool.Rent(ArraySize);
+        var array = _arrayPool.Rent(ArraySize);
 
-        arrayPool.Return(array);
+        _arrayPool.Return(array);
     }
 
     [Benchmark]
--- a/collections-arrays/ArrayPool/ArrayPool.Benchmark/Benchmark.cs
+++ b/collections-arrays/ArrayPool/ArrayPool.Benchmark/Benchmark.cs
@@ -6,10 +6,10 @@ namespace ArrayPool.Benchmark;
 [MemoryDiagnoser]
 public class Benchmark
 {
-    [Params(100, 1000, 10000, 100000)]
-    public int ArraySize { get; set; }
-
     private ArrayPool<int>? _arrayPool;
+
+    [Params(100, 1000, 10000, 100000)]
+    public int ArraySize { get; set; }    
 
     [GlobalSetup]
     public void GlobalSetup()
--- a/collections-arrays/ArrayPool/ArrayPool.Benchmark/Benchmark.cs
+++ b/collections-arrays/ArrayPool/ArrayPool.Benchmark/Benchmark.cs
@@ -1,0 +1,33 @@
+﻿using BenchmarkDotNet.Attributes;
+using System.Buffers;
+
+namespace ArrayPool.Benchmark;
+
+[MemoryDiagnoser]
+public class Benchmark
+{
+    [Params(100, 1000, 10000, 100000)]
+    private int ArraySize { get; set; }
+
+    private ArrayPool<int>? arrayPool;
+
+    [GlobalSetup]
+    public void GlobalSetup()
+    {
+        arrayPool = ArrayPool<int>.Shared;
+    }
+
+    [Benchmark]
+    public void CreateArrayWithArrayPool()
+    {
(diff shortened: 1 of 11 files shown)


collections-arrays/ArrayPool/ArrayPool/Methods/GetArrayPool.cs
```diff
@@ -0,0 +1,13 @@
...
+namespace ArrayPool;
+
+public partial class Methods
+{
+    public static ArrayPool<T> GetDefaultArrayPool<T>()
+    {
+        var arrayPool = ArrayPool<T>.Shared;
+
+        return arrayPool;
+    }
```
jshergal, line 12: This could be an expression body statement

collections-arrays/ArrayPool/ArrayPool/Program.cs
```diff
@@ -0,0 +1,17 @@
...
+
+var array = pool.Rent(arraySize);
+
+for (var i = 0; i < arraySize; i++)
+    array[i] = i * 2;
+
+Console.WriteLine("Array elements:");
+
+foreach (var value in array)
+    Console.Write(value + " ");
```
jshergal, line 15: This code may not be doing what we think it is doing. The array returned from the ArrayPool when we Rent an array it is only guaranteed that it is at least the size we asked for, but may be bigger. So this foreach could be printing out junk. Here is a dotnet fiddle showing, that in this case, the request for an array of length 10, returns an array of length 16.
https://dotnetfiddle.net/g0jyS2

KonstantinosP7: I have explicitly mentioned in the article about the size of the array when we Rent one from the pool. In this specific case, yes it prints the total array. I have written it on purpose to show that when we rent an array from the pool, we get an array of at least the size we request.
Pool returns us the first available array with at least the size we want.

collections-arrays/ArrayPool/ArrayPool/Methods/GetArrayPool.cs
KonstantinosP7: Fixed!

collections-arrays/ArrayPool/ArrayPool/Program.cs
jshergal: @KonstantinosP7 - Got it. I hadn't taken a look at the article text, just commented that it may not be doing what we think, but obviously, it is being used as an example to show the fact that the array may be longer.

jshergal: @KonstantinosP7 - I got to thinking about this more, and this is just my opinion and a thought to consider, but I was wondering if maybe it would be better for us to show the fact that the returned array will be at least as large as we ask and potentially larger, via a simply Console.WriteLine of the requested length and the returned array length, rather than using an example with code that is incorrect or potentially dangerous. This way the potential confusion is avoided. Again, just a thought.

collections-arrays/ArrayPool/ArrayPool.Benchmark/Benchmark.cs
```diff
@@ -0,0 +1,33 @@
+﻿using BenchmarkDotNet.Attributes;
+using System.Buffers;
+
+namespace ArrayPool.Benchmark;
+
+[MemoryDiagnoser]
+public class Benchmark
+{
+    [Params(100, 1000, 10000, 100000)]
+    private int ArraySize { get; set; }
```
mahrncic, line 10: Since this is `private`, why is it a property? Could we use a field instead or? Also, let's follow our naming convention where all private fields should start with the underscore.

collections-arrays/ArrayPool/ArrayPool.Benchmark/Benchmark.cs
```diff
@@ -0,0 +1,33 @@
...
+
+    private ArrayPool<int>? arrayPool;
```
mahrncic, line 12: Private fields names should start with an underscore, so this should be `_arrayPool` instead of `arrayPool`

mahrncic (changes requested):

collections-arrays/ArrayPool/ArrayPool.Benchmark/Benchmark.cs
KonstantinosP7: Fixed!

collections-arrays/ArrayPool/ArrayPool/Program.cs
KonstantinosP7: @jshergal - I agree with you. Instead of printing all the elements of the rented array, I printed the size instead. You will see that in the PR soon. Thanks!

collections-arrays/ArrayPool/ArrayPool.Benchmark/Benchmark.cs
KonstantinosP7: @mahrncic -  The correct type is `public`. That is because the [Params] attribute needs `public` setter, and not `private`. I fixed that in the next PR.

mahrncic (changes requested):

collections-arrays/ArrayPool/ArrayPool.Benchmark/Benchmark.cs
```diff
@@ -0,0 +1,33 @@
...
+
+namespace ArrayPool.Benchmark;
+
+[MemoryDiagnoser]
+public class Benchmark
+{
+    [Params(100, 1000, 10000, 100000)]
+    public int ArraySize { get; set; }
+
+    private ArrayPool<int>? _arrayPool;
```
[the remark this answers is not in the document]
KonstantinosP7: Fixed.

mahrncic (approved):

PR #1382, opened 2023-09-02
CM-559: Handling Dates With Noda Time in .NET

aineaTheSecond:


dotnet-client-libraries/HandlingDatesWithNodaTime/Tests/NodaTimeUnitTest.cs
```diff
@@ -0,0 +1,120 @@
...
+    [Fact]
+    public void WhenGetCurrentInstantMethodIsCalled_ThenReturnCurrentInstant()
+    {
+        var instant = SystemClock.Instance.GetCurrentInstant();
+
+        Assert.True(instant != default(Instant));
+    }
+
+    [Fact]
+    public void WhenFromUnixTimeMethodIsCalled_ThenReturnCurrentInstant()
```
jshergal, line 17: The name of the test is incorrect as the code is not returning the current instant, but rather creating a specific instant based on UTC definition.

dotnet-client-libraries/HandlingDatesWithNodaTime/Tests/NodaTimeUnitTest.cs
```diff
@@ -0,0 +1,120 @@
...
+
+        var expected = new LocalDateTime(2023, 8, 31, 20, 00)
+            .InZoneStrictly(DateTimeZone.Utc)
+            .ToInstant();
+
+        Assert.Equal(expected, instant);
+    }
+
+    [Fact]
+    public void WhenFromUnixTimeTicksMethodIsCalled_ThenReturnCurrentInstant()
```
jshergal, line 29: The name of the test is incorrect as the code is not returning the current instant, but rather creating a specific instant based on unix ticks

dotnet-client-libraries/HandlingDatesWithNodaTime/Tests/NodaTimeUnitTest.cs
```diff
@@ -0,0 +1,120 @@
...
+    public void WhenLocalTimeIsInstantiated_ThenReturnLocalTime()
+    {
+        var localTime = new LocalTime(10, 30);
+
+        Assert.Equal(10, localTime.Hour);
+        Assert.Equal(30, localTime.Minute);
+    }
+
+    [Fact]
+    public void WhenLocalTimeIsConvertedToString_ThenReturnString()
```
jshergal, line 69: This should be renamed a bit also as we are not just checking that it is a string, but that it is being formatted in a specific way.

dotnet-client-libraries/HandlingDatesWithNodaTime/HandlingDatesWithNodaTime/Program.cs
```diff
@@ -0,0 +1 @@
+﻿Console.WriteLine("Application started");
```
jshergal, line 1: We should either show something in the console application, or we should remove it.

dotnet-client-libraries/HandlingDatesWithNodaTime/Tests/NodaTimeUnitTest.cs
aineaTheSecond: This is updated, please review

dotnet-client-libraries/HandlingDatesWithNodaTime/Tests/NodaTimeUnitTest.cs
aineaTheSecond: This is updated, please review

dotnet-client-libraries/HandlingDatesWithNodaTime/Tests/NodaTimeUnitTest.cs
aineaTheSecond: I've made the change. Please take a look.

dotnet-client-libraries/HandlingDatesWithNodaTime/HandlingDatesWithNodaTime/Program.cs
aineaTheSecond: I've opted to remove it entirely. Please review

aineaTheSecond: Actually, the pipeline has failed after removing the statement. Can I show the current time in the console in stead?

jshergal: It wouldn't have built locally for you either. We need to be sure we build locally before we push a commit.
In regards to the question, I had not looked at the article at all yet, so was just proposing that we either do something in the console app OR remove the console app project all together. Now that I have taken a quick look at the article and since we are showing how to create a console app, then we need to keep it. The code samples in the article seem to be something that we would be running in the console app, so we should be demoing each of those things within the console app.

dotnet-client-libraries/HandlingDatesWithNodaTime/HandlingDatesWithNodaTime/Extensions/NodaTimeExtensions.cs
```diff
@@ -0,0 +1,29 @@
...
+
+    public static LocalDate AddDays(this LocalDate date, int days)
+    {
+        return date.PlusDays(days);
+    }
+
+    public static Period CalculateDifference(this LocalDate startDate, DateTime endDate)
+    {   
+        var finalDate = LocalDate.FromDateTime(endDate);
+        return Period.Between(startDate, finalDate, PeriodUnits.Years);
```
karimkod, line 27: Instead of having methods with a single statement just for the sake of the having them in the article, we can create a test case with just the NodaTime method instead of creating a wrapper method like this. 
And in the the article we can have the body of the test in the code snippet for e.g 
the AddDays methods should be a test, something like this : 
```
  [Fact]
  void WhenAddingDaysToDate_ThenHaveCorrectLocalDate() 
   {
        var date = // create LocalDate.
        var threeDaysLater = date.PlusDays(days); // and you can only have this line and the line before in the article and explain it
        // TODO - here assert the resutl
    }
```

dotnet-client-libraries/HandlingDatesWithNodaTime/Tests/NodaTimeUnitTest.cs
```diff
@@ -0,0 +1,228 @@
...
+        var eventTwoTime = new LocalTime(12, 0);
+        var eventThreeTime = new LocalTime(12, 0);
+
+        var eventOneResult = eventOneTime.CompareTo(eventTwoTime);
+        var eventTwoResult = eventTwoTime.CompareTo(eventOneTime);
+        var eventThreeResult = eventTwoTime.CompareTo(eventThreeTime);
+
+        Assert.True(eventOneResult < 0);
+        Assert.True(eventTwoResult > 0);
+        Assert.Equal(0, eventThreeResult);
```
karimkod, line 109: Here, we can assert directly with 1 -1 and 0 instead of greater and Less, it's more precise and more clearer, and you can even have this in the article with the asserts, because in the article we put a paragraph to explain it where in reality the asserts do a very good job at communicating intent.

dotnet-client-libraries/HandlingDatesWithNodaTime/Tests/NodaTimeUnitTest.cs
```diff
@@ -0,0 +1,228 @@
...
+    [Fact]
+    public void WhenLocalDateTimezoneAndDateTimezoneAreCombined_ThenReturnZoneDateTime()
+    {
+        var localDateTime = new LocalDateTime(2023, 8, 31, 15, 30);
+        var timezone = DateTimeZoneProviders.Tzdb["Africa/Cairo"];
+
+        var zoneDateTime = localDateTime.InZoneLeniently(timezone);
+
+        Assert.True(true);
+    }
```
karimkod, line 121: Instead of asserting True, you can assert on the time zone in the zoneDateTime and check if it's equal to Africa/Cairo

dotnet-client-libraries/HandlingDatesWithNodaTime/HandlingDatesWithNodaTime/Extensions/NodaTimeExtensions.cs
karimkod: After you convert those to tests, we can delete this class completely, don't foget to updat the article.

dotnet-client-libraries/HandlingDatesWithNodaTime/Tests/NodaTimeUnitTest.cs
```diff
@@ -0,0 +1,228 @@
...
+
+    [Fact]
+    public void WhenDurationsAreAdded_ThenReturnSumOfDuration()
+    {
+        var durationInDays = Duration.FromDays(1);
+        var durationInHours = Duration.FromHours(3);
+
+        var sumOfDurations = durationInDays + durationInHours;
+
+        Assert.Equal(sumOfDurations, durationInDays + durationInHours);
```
karimkod, line 131: Here the way the expected is created is similar to the calculation, which is redundant and make the assert useless, it's better to  put the expect as a hardcoded duration 
```suggestion
        Assert.Equal(Duration.FromHours(27), durationInDays + durationInHours);
```

dotnet-client-libraries/HandlingDatesWithNodaTime/Tests/NodaTimeUnitTest.cs
```diff
@@ -0,0 +1,228 @@
...
+
+    [Fact]
+    public void WhenDurationsAreSubtracted_ThenReturnDifferenceOfDuration()
+    {
+        var firstDuration = Duration.FromHours(3);
+        var secondDuration = Duration.FromHours(2);
+
+        var sumOfDurations = firstDuration - secondDuration;
+
+        Assert.Equal(sumOfDurations, firstDuration - secondDuration);
```
karimkod, line 142: Same as the addition test.

karimkod (changes requested):
The project should move to dotnet-datetime.

PR #1385, opened 2023-09-05
CM-818:Create Cryptographic Numbers With RandomNumberGenerator in C#

goldytech:
@AndreCarvalho Please review

### DIFF
--- a/dotnet-base-libraries/CryptographyRandomNumberGenerator/App/CryptographicHelpers.cs
+++ b/dotnet-base-libraries/CryptographyRandomNumberGenerator/App/CryptographicHelpers.cs
@@ -1,0 +1,16 @@
+using System.Security.Cryptography;
+
+namespace App;
+
+public static class CryptographicHelpers
+{
+    public static byte[] GenerateRandomKey(int bytesCount)
+    {
+        return RandomNumberGenerator.GetBytes(bytesCount);
+    }
+
+    public static int GenerateRandomInteger(int minValue, int maxValue)
+    {
+        return RandomNumberGenerator.GetInt32(minValue, maxValue);
+    }
+}
--- a/dotnet-base-libraries/CryptographyRandomNumberGenerator/App/Program.cs
+++ b/dotnet-base-libraries/CryptographyRandomNumberGenerator/App/Program.cs
@@ -1,0 +1,3 @@
+﻿// See https://aka.ms/new-console-template for more information
+
+Console.WriteLine("Hello, World!");
--- a/dotnet-base-libraries/CryptographyRandomNumberGenerator/App/Program.cs
+++ b/dotnet-base-libraries/CryptographyRandomNumberGenerator/App/Program.cs
@@ -1,3 +1,7 @@
 ﻿// See https://aka.ms/new-console-template for more information
 
-Console.WriteLine("Hello, World!");
+using App;
+
+Console.WriteLine(Convert.ToBase64String(CryptographicHelpers.GenerateRandomKey(32)));
+Console.WriteLine(CryptographicHelpers.GenerateRandomInteger(1, 100));
+Console.ReadLine();
--- a/dotnet-base-libraries/CryptographyRandomNumberGenerator/Tests/CryptographicHelperUnitTest.cs
+++ b/dotnet-base-libraries/CryptographyRandomNumberGenerator/Tests/CryptographicHelperUnitTest.cs
@@ -1,0 +1,28 @@
+using App;
+
+namespace Tests;
+
+public class CryptographicHelperUnitTest
+{
+    [Theory]
+    [InlineData(32)]
+    public void WhenRandomNumberGeneratorApiIsInvokedEveryTime_ThenItReturnsAUniqueRandomNumber(int bytesCount)
+    {
+        var randomNumberOne = CryptographicHelpers.GenerateRandomKey(bytesCount);
+        var randomNumberTwo = CryptographicHelpers.GenerateRandomKey(bytesCount);
+
+        Assert.NotEmpty(randomNumberOne);
+        Assert.NotEmpty(randomNumberTwo);
+        Assert.NotEqual(randomNumberOne, randomNumberTwo);
+    }
+
+    [Theory]
+    [InlineData(1, 100)]
+    public void WhenRandomNumberRangeIsProvided_ThenRandomNumberGeneratedShouldBeInTheRange(int minValue, int maxValue)
+    {
+        var randomNumber = CryptographicHelpers.GenerateRandomInteger(minValue, maxValue);
+
+        Assert.NotEqual(0, randomNumber);
+        Assert.InRange(randomNumber, minValue, maxValue);
+    }
+}
--- a/dotnet-base-libraries/CryptographyRandomNumberGenerator/App/App.csproj
+++ b/dotnet-base-libraries/CryptographyRandomNumberGenerator/App/App.csproj
@@ -1,0 +1,10 @@
+<Project Sdk="Microsoft.NET.Sdk">
+
+    <PropertyGroup>
+        <OutputType>Exe</OutputType>
+        <TargetFramework>net7.0</TargetFramework>
+        <ImplicitUsings>enable</ImplicitUsings>
+        <Nullable>enable</Nullable>
+    </PropertyGroup>
+
+</Project>
--- a/dotnet-base-libraries/CryptographyRandomNumberGenerator/CryptographyRandomNumberGenerator.sln
+++ b/dotnet-base-libraries/CryptographyRandomNumberGenerator/CryptographyRandomNumberGenerator.sln
@@ -1,0 +1,22 @@
+﻿
+Microsoft Visual Studio Solution File, Format Version 12.00
+Project("{FAE04EC0-301F-11D3-BF4B-00C04F79EFBC}") = "App", "App\App.csproj", "{63CF02DA-36BE-4502-B091-E6A1C22D0E66}"
+EndProject
+Project("{FAE04EC0-301F-11D3-BF4B-00C04F79EFBC}") = "Tests", "Tests\Tests.csproj", "{C383B53B-F476-4B5E-9C5B-F36FD028543D}"
+EndProject
+Global
+	GlobalSection(SolutionConfigurationPlatforms) = preSolution
+		Debug|Any CPU = Debug|Any CPU
+		Release|Any CPU = Release|Any CPU
+	EndGlobalSection
+	GlobalSection(ProjectConfigurationPlatforms) = postSolution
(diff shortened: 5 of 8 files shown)


dotnet-base-libraries/CryptographyRandomNumberGenerator/Tests/CryptographicHelperUnitTest.cs
```diff
@@ -0,0 +1,28 @@
...
+        Assert.NotEqual(randomNumberOne, randomNumberTwo);
+    }
+
+    [Theory]
+    [InlineData(1, 100)]
+    public void WhenRandomNumberRangeIsProvided_ThenRandomNumberGeneratedShouldBeInTheRange(int minValue, int maxValue)
+    {
+        var randomNumber = CryptographicHelpers.GenerateRandomInteger(minValue, maxValue);
+
+        Assert.NotEqual(0, randomNumber);
```
jshergal, line 25: This test is covered by the InRange so is not necessary.

dotnet-base-libraries/CryptographyRandomNumberGenerator/App/CryptographicHelpers.cs
```diff
@@ -0,0 +1,16 @@
+using System.Security.Cryptography;
+
+namespace App;
+
+public static class CryptographicHelpers
+{
+    public static byte[] GenerateRandomKey(int bytesCount)
+    {
+        return RandomNumberGenerator.GetBytes(bytesCount);
```
jshergal, line 9: We can make this an expression body statement

dotnet-base-libraries/CryptographyRandomNumberGenerator/App/CryptographicHelpers.cs
```diff
@@ -0,0 +1,16 @@
...
+    }
+
+    public static int GenerateRandomInteger(int minValue, int maxValue)
+    {
+        return RandomNumberGenerator.GetInt32(minValue, maxValue);
```
jshergal, line 14: We can make this an expression body statement

dotnet-base-libraries/CryptographyRandomNumberGenerator/App/Program.cs
```diff
@@ -0,0 +1,7 @@
+﻿// See https://aka.ms/new-console-template for more information
+
+using App;
+
+Console.WriteLine(Convert.ToBase64String(CryptographicHelpers.GenerateRandomKey(32)));
+Console.WriteLine(CryptographicHelpers.GenerateRandomInteger(1, 100));
+Console.ReadLine();
```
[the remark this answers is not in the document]
goldytech: @jshergal Done

dotnet-base-libraries/CryptographyRandomNumberGenerator/App/CryptographicHelpers.cs
goldytech: @jshergal Done

goldytech:
@jshergal The feedback given is now incorporated. Thanks

dotnet-base-libraries/CryptographyRandomNumberGenerator/App/Program.cs
```diff
@@ -0,0 +1,6 @@
+﻿// See https://aka.ms/new-console-template for more information
```
jshergal, line 1: This boilerplate comment is not needed

dotnet-base-libraries/CryptographyRandomNumberGenerator/Tests/CryptographicHelperUnitTest.cs
```diff
@@ -0,0 +1,27 @@
+using App;
```
jshergal, line 1: Since we are using global usings, this should move there as it applies globally to all unit tests, being that this is the namespace under test

dotnet-base-libraries/CryptographyRandomNumberGenerator/App/CryptographicHelpers.cs
```diff
@@ -0,0 +1,12 @@
+using System.Security.Cryptography;
+
+namespace App;
+
+public static class CryptographicHelpers
+{
+    private static readonly Random psuedoRandomNumberGenerator = new(10);
```
jshergal, line 7: Naming convention: static readonly should be Pascal Case

dotnet-base-libraries/CryptographyRandomNumberGenerator/App/CryptographicHelpers.cs
```diff
@@ -0,0 +1,12 @@
...
+    public static byte[] GenerateRandomKey(int bytesCount) => RandomNumberGenerator.GetBytes(bytesCount);
+    public static int GenerateRandomInteger(int minValue, int maxValue) => RandomNumberGenerator.GetInt32(minValue, maxValue);
+    public static int GeneratePseudoRandom(int minValue, int maxValue) => psuedoRandomNumberGenerator.Next(minValue, maxValue);
+
```
jshergal, line 12: Extra newline

dotnet-base-libraries/CryptographyRandomNumberGenerator/App/CryptographicHelpers.cs
```diff
@@ -0,0 +1,12 @@
+using System.Security.Cryptography;
+
+namespace App;
+
+public static class CryptographicHelpers
+{
+    private static readonly Random psuedoRandomNumberGenerator = new(10);
+    public static byte[] GenerateRandomKey(int bytesCount) => RandomNumberGenerator.GetBytes(bytesCount);
+    public static int GenerateRandomInteger(int minValue, int maxValue) => RandomNumberGenerator.GetInt32(minValue, maxValue);
+    public static int GeneratePseudoRandom(int minValue, int maxValue) => psuedoRandomNumberGenerator.Next(minValue, maxValue);
```
jshergal: For readability we should add newlines between the methods

dotnet-base-libraries/CryptographyRandomNumberGenerator/App/Program.cs
```diff
@@ -0,0 +1,6 @@
+using App;
+using BenchmarkDotNet.Running;
+
+Console.WriteLine(Convert.ToBase64String(CryptographicHelpers.GenerateRandomKey(32)));
+Console.WriteLine(CryptographicHelpers.GenerateRandomInteger(1, 100));
+var summary = BenchmarkRunner.Run<RandomNumberBenchmarks>();
```
jshergal, line 6: No need for the variable here

dotnet-base-libraries/CryptographyRandomNumberGenerator/App/RandomNumberBenchmarks.cs
```diff
@@ -0,0 +1,21 @@
...
+    }
+    
+    [Benchmark]
+    public void UsingPseudoRandomNumberGenerator()
+    {
+        var randomNumber = CryptographicHelpers.GeneratePseudoRandom(1, 100);
+    }
+    
+    
+
```
jshergal: Extra newlines

dotnet-base-libraries/CryptographyRandomNumberGenerator/App/RandomNumberBenchmarks.cs
```diff
@@ -0,0 +1,21 @@
+using BenchmarkDotNet.Attributes;
+
+namespace App;
+
+public class RandomNumberBenchmarks
+{
+    [Benchmark]
+    public void UsingRealRandomNumberGenerator()
+    {
+        var randomNumber = CryptographicHelpers.GenerateRandomInteger(1, 100);
```
jshergal, line 10: We should add the return type and return this value as it will prevent the JIT from changing this to a noop. It is not so much a problem anymore, but is still considered a best practice with benchmarks to ensure the method is not nooped.

dotnet-base-libraries/CryptographyRandomNumberGenerator/App/RandomNumberBenchmarks.cs
```diff
@@ -0,0 +1,21 @@
...
+    }
+    
+    [Benchmark]
+    public void UsingPseudoRandomNumberGenerator()
+    {
+        var randomNumber = CryptographicHelpers.GeneratePseudoRandom(1, 100);
```
jshergal, line 16: Same as above

dotnet-base-libraries/CryptographyRandomNumberGenerator/App/CryptographicHelpers.cs
jshergal: Need a unit test for this method since it has been added.

goldytech:
@jshergal @AndreCarvalho All the feedback changes are done. Thanks

dotnet-base-libraries/CryptographyRandomNumberGenerator/App/CryptographicHelpers.cs
```diff
@@ -0,0 +1,14 @@
...
+
+public static class CryptographicHelpers
+{
+    private static readonly Random PseudoRandomNumberGenerator = new(10);
+    
+    public static byte[] GenerateRandomKey(int bytesCount) => RandomNumberGenerator.GetBytes(bytesCount);
+    
+    public static int GenerateRandomInteger(int minValue, int maxValue) => RandomNumberGenerator.GetInt32(minValue, maxValue);
+    
+    public static int GeneratePseudoRandom(int minValue, int maxValue) => PseudoRandomNumberGenerator.Next(minValue, maxValue);
```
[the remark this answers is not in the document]
jshergal: @AndreCarvalho - the point here is showing cryptographically secure randomness which requires using this class:
https://learn.microsoft.com/en-us/dotnet/api/system.security.cryptography.randomnumbergenerator?view=net-7.0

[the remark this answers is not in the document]
jshergal: @AndreCarvalho - ignore my comment, I read your comment as pertaining to the line above, my bad.

dotnet-base-libraries/CryptographyRandomNumberGenerator/App/CryptographicHelpers.cs
```diff
@@ -0,0 +1,14 @@
+using System.Security.Cryptography;
+
+namespace App;
+
+public static class CryptographicHelpers
```
jshergal, line 5: I like the renaming we did with the benchmarks. I think it would be good to go ahead and do the same here. Also, as was pointed out by @AndreCarvalho, since we are not specifying a seed with `RandomNumberGenerator`, we should probably be consistent and just use `Random.Shared` for the general random number case. With that change we can drop the static member variable as well, which is nice.
Then for the rename we can just do something like `GenerateSecureRandomKey`, `GenerateSecureRandomInteger`, `GenerateGeneralRandomInteger`

PR #1392, opened 2023-09-12
CM-793: Selecting Xml Nodes With Xpath in C#

igorgervasio:


xml-csharp/SelectingXmlNodesWithXpath/SelectingXmlNodesWithXpath/Program.cs
```diff
@@ -0,0 +1,67 @@
...
+                SelectBooks(root);
+
+                Console.WriteLine("\nLet's select a single book using XML namespaces:");
+                SelectBooksUsingNamespaces(doc);
+            }
+            catch (Exception)
+            {
+
+                throw;
+            }
```
jshergal, line 64: If we are just catching and rethrowing, then we can probably remove the try/catch

xml-csharp/SelectingXmlNodesWithXpath/SelectingXmlNodesWithXpathTests/BooksCatalog.xml
```diff
@@ -0,0 +1,27 @@
...
+		<author>King, Stephen</author>
+		<title>IT</title>
+		<genre>Horror</genre>
+		<price>40.00</price>
+	</book>
+	<book id="2">
+		<author>Assis, Machado De</author>
+		<title>Dom Casmurro</title>
+		<genre>Romance</genre>
+		<price>50.00</price>
```
jshergal, line 13: This is a duplicate of the other BooksCatalog.xml, we could put it in a resource folder and share it between the projects. This ensures that they both are dealing with the same data and avoids issues like updating one and not the other.

xml-csharp/SelectingXmlNodesWithXpath/SelectingXmlNodesWithXpathTests/SelectingXmlNodesWithXpathTest.cs
```diff
@@ -0,0 +1,123 @@
+using System.Diagnostics;
+using System.Xml;
+using System.Xml.Linq;
+using Xunit;
```
jshergal, line 4: This should have shown as a warning in the IDE, we have this in the global using so it is not needed here. We could also move the others to global usings

xml-csharp/SelectingXmlNodesWithXpath/SelectingXmlNodesWithXpathTests/SelectingXmlNodesWithXpathTest.cs
```diff
@@ -0,0 +1,123 @@
+using System.Diagnostics;
+using System.Xml;
+using System.Xml.Linq;
+using Xunit;
+
+namespace SelectingXmlNodesWithXpathTests
```
jshergal, line 6: Using file level namespace will reduce nesting

xml-csharp/SelectingXmlNodesWithXpath/SelectingXmlNodesWithXpathTests/SelectingXmlNodesWithXpathTest.cs
```diff
@@ -0,0 +1,123 @@
...
+using System.Xml.Linq;
+using Xunit;
+
+namespace SelectingXmlNodesWithXpathTests
+{
+    public class SelectingXmlNodesWithXpathTest
+    {
+        XmlDocument Document { get; set; }
+        Dictionary<string, string> ExpectedResults { get; set; }
+
```
jshergal, line 12: Extra blank line

xml-csharp/SelectingXmlNodesWithXpath/SelectingXmlNodesWithXpathTests/SelectingXmlNodesWithXpathTest.cs
```diff
@@ -0,0 +1,123 @@
...
+                            "<genre>Scientific</genre>" +
+                            "<price>60.00</price>" +
+                        "</book>"
+                    }
+                };
+            }
+            catch (Exception)
+            {
+                throw;
+            }
```
jshergal, line 64: No need for try/catch if we are just rethrowing and not doing anything with it

xml-csharp/SelectingXmlNodesWithXpath/SelectingXmlNodesWithXpathTests/SelectingXmlNodesWithXpathTest.cs
```diff
@@ -0,0 +1,123 @@
...
+                        "</book>"
+                    },
+                    {
+                        "Book4",
+                        "<book id=\"4\" xmlns=\"urn:example-schema\">" +
+                            "<author>Fowler, Martin; Beck, Kent</author>" +
+                            "<title>Refactoring: Improving the design of existing code</title>" +
+                            "<genre>Scientific</genre>" +
+                            "<price>60.00</price>" +
+                        "</book>"
```
jshergal, line 57: We can use raw string literals here to make this code a little nicer and easier to maintain.

xml-csharp/SelectingXmlNodesWithXpath/SelectingXmlNodesWithXpathTests/SelectingXmlNodesWithXpathTest.cs
```diff
@@ -0,0 +1,123 @@
...
+        [Fact]
+        public void GivenAnXmlFile_WhenSelectingASingleNode_ThenReturnsTheSecondPosition()
+        {
+            //Arrange
+
+            //Act
+            var node = Document.SelectSingleNode("//catalog/book[position()=2]");
+
+            //Assert
+            Assert.Equal(node.OuterXml, ExpectedResults["Book2"]);
```
jshergal, line 76: The //Arrange, //Act and //Assert comments aren't needed in the unit tests

xml-csharp/SelectingXmlNodesWithXpath/SelectingXmlNodesWithXpathTests/SelectingXmlNodesWithXpathTest.cs
```diff
@@ -0,0 +1,123 @@
...
+using System.Xml;
+using System.Xml.Linq;
+using Xunit;
+
+namespace SelectingXmlNodesWithXpathTests
+{
+    public class SelectingXmlNodesWithXpathTest
+    {
+        XmlDocument Document { get; set; }
+        Dictionary<string, string> ExpectedResults { get; set; }
```
jshergal, line 11: We should add `private` to these for consistency and clarity. They both can also be made `get` only since they are only set in the constructor.

xml-csharp/SelectingXmlNodesWithXpath/SelectingXmlNodesWithXpathTests/SelectingXmlNodesWithXpathTest.cs
```diff
@@ -0,0 +1,123 @@
...
+        {
+            //Arrange
+
+            //Act
+            var nodes = Document
+                .SelectNodes("//catalog/book[price<50.00]");
+
+            var outerXmls = nodes
+                .Cast<XmlNode>()
+                .Select(_ => _.OuterXml);
```
jshergal, line 90: We should replace `_` with a variable. `_` is technically valid, but in this context will be read as a discard, but is being used which makes the code confusing.

xml-csharp/SelectingXmlNodesWithXpath/SelectingXmlNodesWithXpathTests/SelectingXmlNodesWithXpathTest.cs
```diff
@@ -0,0 +1,123 @@
...
+            //Act
+            var nsmgr = new XmlNamespaceManager(Document.NameTable);
+            nsmgr.AddNamespace("ex", "urn:example-schema");
+
+            var nodes = Document
+                .SelectNodes("descendant::ex:book", nsmgr);
+
+            var outerXmls = nodes
+                .Cast<XmlNode>()
+                .Select(_ => _.OuterXml);
```
jshergal, line 114: We should replace `_` with a variable. `_` is technically valid, but in this context will be read as a discard, but is being used which makes the code confusing.

jshergal (changes requested):
There are also a few compiler warnings related to nullable which we should try to cleanup as it is good practice. Secondly, there are inconsistent line endings in the some of the files.

xml-csharp/SelectingXmlNodesWithXpath/SelectingXmlNodesWithXpathTests/SelectingXmlNodesWithXpathTest.cs
```diff
@@ -0,0 +1,99 @@
...
+        ExpectedResults = new Dictionary<string, string>()
+        {
+            {
+                "Book1",
+                """<book id="1">""" +
+                    """<author>King, Stephen</author>""" +
+                    """<title>IT</title>""" +
+                    """<genre>Horror</genre>""" +
+                    """<price>40.00</price>""" +
+                """</book>"""
```
jshergal, line 24: Please review the documentation for raw string literals, there is no need for the `+` in this, we can just write the string in a formatted way, including new lines and have only one opening and closing set of `"""`
https://learn.microsoft.com/en-us/dotnet/csharp/language-reference/tokens/raw-string

xml-csharp/SelectingXmlNodesWithXpath/SelectingXmlNodesWithXpathTests/SelectingXmlNodesWithXpathTest.cs
```diff
@@ -0,0 +1,99 @@
+using System.Xml;
+
+namespace SelectingXmlNodesWithXpathTests;
+
+public class SelectingXmlNodesWithXpathTest
```
jshergal, line 5: The unit tests should be testing the code that we are adding to the article, so the class under test here should be the `XmlNodeSelector`. We don't need to write unit tests to validate the framework. I should have caught that in my last review, my bad.

jshergal (changes requested):

xml-csharp/SelectingXmlNodesWithXpath/SelectingXmlNodesWithXpathTests/SelectingXmlNodesWithXpathTest.cs
igorgervasio: @jshergal, the string was split like that because the `OuterXml` property is not formatted. I followed your suggestion, but, to that, I had to add a new method to format the strings. In the end, I think that the final result is better.

I also don't know why I was printing the result as an unformatted XML. I'll update the article with this new printing format.

xml-csharp/SelectingXmlNodesWithXpath/SelectingXmlNodesWithXpathTests/SelectingXmlNodesWithXpathTest.cs
igorgervasio: I totally agree with you about how the tests must be made. For this project, I looked at other projects in the code-maze repository for reference and they are using the initial testing implementation.

I removed the print from inside the methods and changed their returning types from void to `string` and `List<string>`. This way, the data printing was passed to the `Program.cs`.

I liked more the actual result. I will update the article to reflect the changes.

xml-csharp/SelectingXmlNodesWithXpath/SelectingXmlNodesWithXpathTests/XmlNodesSelectorTest.cs
```diff
@@ -0,0 +1,100 @@
+using SelectingXmlNodesWithXpath;
+using System.Xml;
+using System.Xml.Linq;
```
jshergal, line 3: This using is not used

xml-csharp/SelectingXmlNodesWithXpath/SelectingXmlNodesWithXpathTests/XmlNodesSelectorTest.cs
```diff
@@ -0,0 +1,100 @@
+using SelectingXmlNodesWithXpath;
+using System.Xml;
```
jshergal, line 2: These should move to the global usings file, since we have one and these will apply across the board to anything within our unit tests

xml-csharp/SelectingXmlNodesWithXpath/SelectingXmlNodesWithXpathTests/XmlNodesSelectorTest.cs
```diff
@@ -0,0 +1,100 @@
...
+        var result = XmlNodesSelector.SelectSingleBook(Document.DocumentElement!);
+
+        Assert.Equal(result, ExpectedResults["Book2"]);
+    }
+
+    [Fact]
+    public void GivenAnXmlFile_WhenSelectingNodes_ThenReturnBooksWithPriceLowerThan50()
+    {
+        var expected = ExpectedResults
+            .Where(pair => pair.Key == "Book1" || pair.Key == "Book3")
```
jshergal, line 78: We could use pattern matching here to convert this to a logical pattern, it will be more readable:
`.Where(pair => pair.Key is "Book1" or "Book3")`

xml-csharp/SelectingXmlNodesWithXpath/SelectingXmlNodesWithXpath/XmlNodesSelector.cs
```diff
@@ -1,42 +1,45 @@
...
-                Console.WriteLine($"Book: {nodes[i]?.OuterXml}");
-            }
-        }
+
+        var outerXmls = nodes!
+            .Cast<XmlNode>()
+            .Select(x => FormatXml(x.OuterXml))
+            .ToList();
+
+        return outerXmls;
```
jshergal, line 28: We can just return this directly

xml-csharp/SelectingXmlNodesWithXpath/SelectingXmlNodesWithXpath/XmlNodesSelector.cs
```diff
@@ -1,42 +1,45 @@
...
```
jshergal, line 43: We can just return this directly

jshergal (approved):
It definitely looks better and more readable. Nice job. Left some additional comments for minor things.

xml-csharp/SelectingXmlNodesWithXpath/SelectingXmlNodesWithXpathTests/XmlNodesSelectorTest.cs
igorgervasio: Veeery nice tip. Thanks for reminding me of these keywords.

jshergal (approved):

PR #1395, opened 2023-09-13
CM-728:How to Get a Database Row as JSON Using Dapper

Otimkev:


dotnet-dapper/HowtoGetaDatabaseRowasJSONUsingDapper/HowtoGetaDatabaseRowasJSONUsingDapper/Repository/Repository.cs
```diff
@@ -0,0 +1,39 @@
...
+    public async Task<dynamic?> GetById(int id)
+    {
+        const string query = "SELECT * FROM Entities WHERE Id = @Id";
+
+        using (var connection = _context.CreateConnection())
+        {
+            var entity = await connection.QuerySingleOrDefaultAsync(query, new { id });
+            if (entity != null)
+            {
+                string json = JsonConvert.SerializeObject(entity, Formatting.Indented);
```
jshergal, line 28: This should be var

Otimkev: Is it okay if I updated the article to have string instead?

jshergal: @Otimkev they both need to match either way, but I am not sure why we wouldn't just use var here. There is no ambiguity as to what the type of object is. Both the call and the variable name make it clear.

Otimkev: JsonConvert.SerializeObject is returning a dynamic result and I cannot seem to log it.

jshergal: @Otimkev - that is bizarre. I just pulled the code myself and I can't figure out what in the world is happening there. Using var should not be confusing the compiler as the return type of that call is explicitly string, but anyway, you don't need to hold up the article for that. Yes, change the article text to match and use string.

One other note, related to the QuerySingleOrDefault call, is it possible to make that call by passing the id directly without constructing a dynamic object as the argument? I don't use Dapper, so I wanted to check with you on that. From the docs, it looks like it is just a object[] param array, so passing an int should be fine, correct?

PR #1418, opened 2023-09-25
CM-94 | ConcurrentQueue in C#

MichalBrylKaminski:


collections-csharp/ConcurrentQueueInCSharp/ConcurrentQueueInCSharp.Console/Consumer.cs
```diff
@@ -0,0 +1,28 @@
+﻿using ConcurrentQueueInCSharp.Models;
+
+namespace ConcurrentQueueInCSharp
```
jshergal, line 3: File level namespace will reduce nesting

collections-csharp/ConcurrentQueueInCSharp/ConcurrentQueueInCSharp.Console/Consumer.cs
```diff
@@ -0,0 +1,28 @@
...
+        public Consumer(OrderMessageBus messageBus)
+        {
+            _messageBus = messageBus;
+        }
+
+        public Task Process()
+        {
+            return Task.Factory.StartNew(() =>
+            {
+                Order order;
```
jshergal, line 18: Can be moved inline with the `Fetch` statement

collections-csharp/ConcurrentQueueInCSharp/ConcurrentQueueInCSharp.Console/Consumer.cs
```diff
@@ -0,0 +1,28 @@
...
+                string processId = Guid.NewGuid().ToString();
```
jshergal, line 19: This can be var, but better still, we can just move this code (minus the `ToString` into the interpolated string

collections-csharp/ConcurrentQueueInCSharp/ConcurrentQueueInCSharp.Console/Models/Order.cs
```diff
@@ -0,0 +1,7 @@
+﻿namespace ConcurrentQueueInCSharp.Models
+{
+    public class Order
+    {
+        public string Id { get; set; }
```
jshergal, line 5: The `set` can be made `init`. It might be cleaner to even add a default value here, as the only time we are really setting this directly is in the tests, where we want to ensure we are getting back the same element we added. Even that case can be covered by just storing the ID value after creation as our expected result.

collections-csharp/ConcurrentQueueInCSharp/ConcurrentQueueInCSharp.Console/OrderMessageBus.cs
```diff
@@ -0,0 +1,30 @@
+﻿using ConcurrentQueueInCSharp.Models;
+using System.Collections.Concurrent;
+
+namespace ConcurrentQueueInCSharp
```
jshergal, line 4: File scoped namespace here

collections-csharp/ConcurrentQueueInCSharp/ConcurrentQueueInCSharp.Console/OrderMessageBus.cs
```diff
@@ -0,0 +1,30 @@
...
+{
+    public class OrderMessageBus
+    {
+        private readonly ConcurrentQueue<Order> _queue = new ConcurrentQueue<Order>();
+
+        public int Count
```
jshergal, line 10: We can use an expression bodied property here

collections-csharp/ConcurrentQueueInCSharp/ConcurrentQueueInCSharp.Console/OrderMessageBus.cs
```diff
@@ -0,0 +1,30 @@
...
+        {
+            get
+            {
+                return _queue.Count;
+            }
+        }
+
+        public void Add(Order order)
```
jshergal, line 18: Since we are allowing null as an input (for testing purposes), we should mark this as nullable

collections-csharp/ConcurrentQueueInCSharp/ConcurrentQueueInCSharp.Console/OrderMessageBus.cs
```diff
@@ -0,0 +1,30 @@
...
+        {
+            if (order == null) throw new ArgumentNullException(nameof(order));
```
jshergal, line 20: We can clean this up with the `ArgumentNullException.ThrowIfNull(order);`. The throw helper uses the `CallerArgumentExpression` attribute to pull the argument name.

collections-csharp/ConcurrentQueueInCSharp/ConcurrentQueueInCSharp.Console/OrderMessageBus.cs
```diff
@@ -0,0 +1,30 @@
...
+
+            _queue.Enqueue(order);
+        }
+
+        public bool Fetch(out Order order)
```
jshergal, line 25: This should be marked as nullable

collections-csharp/ConcurrentQueueInCSharp/ConcurrentQueueInCSharp.Console/Producer.cs
```diff
@@ -0,0 +1,27 @@
+﻿using ConcurrentQueueInCSharp.Models;
+
+namespace ConcurrentQueueInCSharp
```
jshergal, line 3: File scoped namespace

collections-csharp/ConcurrentQueueInCSharp/ConcurrentQueueInCSharp.Console/Producer.cs
```diff
@@ -0,0 +1,27 @@
...
+            _numberOfMessages = numberOfMessages;
+        }
+
+        public Task Produce()
+        {
+            return Task.Factory.StartNew(() =>
+            {
+                for (int i = 0; i < _numberOfMessages; i++)
+                {
+                    _messageBus.Add(new Order { Id = Guid.NewGuid().ToString() });
```
jshergal, line 22: If we modify order to have create the Id itself (which would often be a typical pattern), we can simplify this line.

collections-csharp/ConcurrentQueueInCSharp/ConcurrentQueueInCSharp.Console/Producer.cs
```diff
@@ -0,0 +1,27 @@
...
+
+        public Producer(OrderMessageBus messageBus, int numberOfMessages)
+        {
+            _messageBus = messageBus;
+            _numberOfMessages = numberOfMessages;
+        }
+
+        public Task Produce()
+        {
+            return Task.Factory.StartNew(() =>
```
jshergal, line 18: We can use `Task.Run` to shorten the line. It is also the more standard practice.

collections-csharp/ConcurrentQueueInCSharp/ConcurrentQueueInCSharp.Console/Consumer.cs
```diff
@@ -0,0 +1,28 @@
...
+        private readonly OrderMessageBus _messageBus;
+
+        public Consumer(OrderMessageBus messageBus)
+        {
+            _messageBus = messageBus;
+        }
+
+        public Task Process()
+        {
+            return Task.Factory.StartNew(() =>
```
jshergal, line 16: We can just use `Task.Run` here

collections-csharp/ConcurrentQueueInCSharp/ConcurrentQueueInCSharp.Console/Program.cs
```diff
@@ -0,0 +1,22 @@
...
+        {
+            OrderMessageBus messageBus = new OrderMessageBus();
+
+            Producer producer1 = new Producer(messageBus, 10);
+            Producer producer2 = new Producer(messageBus, 10);
+            Producer producer3 = new Producer(messageBus, 10);
+
+            Consumer consumer1 = new Consumer(messageBus);
+            Consumer consumer2 = new Consumer(messageBus);
+            Consumer consumer3 = new Consumer(messageBus);
```
jshergal, line 15: Use var here

collections-csharp/ConcurrentQueueInCSharp/ConcurrentQueueInCSharp.Console/Consumer.cs
MichalBrylKaminski: Indeed, although that would create new instances of order with each loop, for that reason I left it in the outer scope.

collections-csharp/ConcurrentQueueInCSharp/ConcurrentQueueInCSharp.Console/Consumer.cs
MichalBrylKaminski: To tell the truth, I'm not the biggest fan of file-level namespaces as I find it less readable than the classic approach.

collections-csharp/ConcurrentQueueInCSharp/ConcurrentQueueInCSharp.Console/Consumer.cs
MichalBrylKaminski: But for the example sake it probably doesn't really matter.

collections-csharp/ConcurrentQueueInCSharp/ConcurrentQueueInCSharp.Console/Consumer.cs
MichalBrylKaminski: Good point.

collections-csharp/ConcurrentQueueInCSharp/ConcurrentQueueInCSharp.Console/Models/Order.cs
MichalBrylKaminski: Good point. I was originally thinking about using `init;` but ultimately settled down with this.

collections-csharp/ConcurrentQueueInCSharp/ConcurrentQueueInCSharp.Console/OrderMessageBus.cs
```diff
@@ -0,0 +1,30 @@
+﻿using ConcurrentQueueInCSharp.Models;
+using System.Collections.Concurrent;
+
+namespace ConcurrentQueueInCSharp
+{
+    public class OrderMessageBus
+    {
+        private readonly ConcurrentQueue<Order> _queue = new ConcurrentQueue<Order>();
```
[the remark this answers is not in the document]
MichalBrylKaminski: Good point.

collections-csharp/ConcurrentQueueInCSharp/ConcurrentQueueInCSharp.Console/OrderMessageBus.cs
MichalBrylKaminski: Good point.

collections-csharp/ConcurrentQueueInCSharp/ConcurrentQueueInCSharp.Console/OrderMessageBus.cs
MichalBrylKaminski: Good point.

collections-csharp/ConcurrentQueueInCSharp/ConcurrentQueueInCSharp.Console/OrderMessageBus.cs
MichalBrylKaminski: Nice one. I didn't know it.

collections-csharp/ConcurrentQueueInCSharp/ConcurrentQueueInCSharp.Console/OrderMessageBus.cs
MichalBrylKaminski: Good point.

collections-csharp/ConcurrentQueueInCSharp/ConcurrentQueueInCSharp.Console/Producer.cs
MichalBrylKaminski: Good point.

collections-csharp/ConcurrentQueueInCSharp/ConcurrentQueueInCSharp.Console/Program.cs
MichalBrylKaminski: Good point.

collections-csharp/ConcurrentQueueInCSharp/ConcurrentQueueInCSharpTests/ProducerTests.cs
```diff
@@ -0,0 +1,54 @@
...
+        public void WhenFetchingOrder_ThenOrderIsReturned()
+        {
+            Order order = new Order
+            {
+                Id = "9425f6b8-e57d-494c-a4c3-9e70849135a9"
+            };
+
+            _messageBus.Add(order);
+
+            Order returnedOrder;
```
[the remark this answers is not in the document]
MichalBrylKaminski: Good point.

collections-csharp/ConcurrentQueueInCSharp/ConcurrentQueueInCSharpTests/ProducerTests.cs
```diff
@@ -0,0 +1,54 @@
...
+        [Test]
+        public void WhenAddingEmptyOrder_ThenArgumentNullExceptionIsThrown()
+        {
+            Assert.Throws<ArgumentNullException>(() => _messageBus.Add(null));
+        }
+
+        [Test]
+        public void WhenFetchingOrder_ThenOrderIsReturned()
+        {
+            Order order = new Order
```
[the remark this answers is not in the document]
MichalBrylKaminski: Good point.

collections-csharp/ConcurrentQueueInCSharp/ConcurrentQueueInCSharp.Console/Producer.cs
MichalBrylKaminski: I personally wouldn't do this. Normally I would create some kind of factory class for `Order` that would take care of it since I could be using the `Order` class to retrieve existing orders from the database and I want to maintain the same order id.

collections-csharp/ConcurrentQueueInCSharp/ConcurrentQueueInCSharp.Console/Consumer.cs
jshergal: There is no change nor additional allocation within the loop vs. outside the loop, just more readable as the variable is localized to its usage. Just take a look at the IL.

collections-csharp/ConcurrentQueueInCSharp/ConcurrentQueueInCSharp.Console/Consumer.cs
jshergal: @MichalBrylKaminski to each his own, but personally I find the reduced nesting to be a big win. We remove the initial 4 space/tab indent without changing any code flow. The only major difference is a removal if a single pair of brackets.

MichalBrylKaminski: I agree @jshergal. It all depends on personal preferences unless the project specifies it as a code standard.

collections-csharp/ConcurrentQueueInCSharp/ConcurrentQueueInCSharp.Console/Consumer.cs
MichalBrylKaminski: Already moved.

MichalBrylKaminski:
Let me know @jshergal if you have any more thoughts.

collections-csharp/ConcurrentQueueInCSharp/ConcurrentQueueInCSharp.Console/Consumer.cs
jshergal: @MichalBrylKaminski - From the Code Maze implementation code guidelines for authors

image

collections-csharp/ConcurrentQueueInCSharp/ConcurrentQueueInCSharp.Console/Consumer.cs
```diff
@@ -13,14 +13,12 @@ public Consumer(OrderMessageBus messageBus)
...
+            return Task.Run(() =>
             {
-                Order order;
-                string processId = Guid.NewGuid().ToString();
-
-                while (_messageBus.Fetch(out order))
+                while (_messageBus.Fetch(out var order))
                 {
-                    Console.WriteLine($"ProcessId {processId} | Processing order {order.Id}");
+                    Console.WriteLine($"ProcessId {Guid.NewGuid()} | Processing order {order.Id}");
```
jshergal, line 20: @MichalBrylKaminski - I just realized that I steered you wrong on this statement, that is my bad. I didn't look at it carefully enough. The ProcessId is something that should be constant within the executing code. So the line creating a processId, should either be put back (less the ToString, as it will be handled in the interpolation):
`var processId = Guid.NewGuid()`
Or we could just use the current executing task ID which is just a static property on the Task class: `Task.CurrentId`
https://learn.microsoft.com/en-us/dotnet/api/system.threading.tasks.task.currentid

collections-csharp/ConcurrentQueueInCSharp/ConcurrentQueueInCSharp.Console/Consumer.cs
MichalBrylKaminski: Right, fair enough. I don't recall this point in the guidelines. I have applied the changes.

collections-csharp/ConcurrentQueueInCSharp/ConcurrentQueueInCSharp.Console/OrderMessageBus.cs
MichalBrylKaminski: Done.

collections-csharp/ConcurrentQueueInCSharp/ConcurrentQueueInCSharp.Console/Producer.cs
MichalBrylKaminski: Done.

collections-csharp/ConcurrentQueueInCSharp/ConcurrentQueueInCSharp.Console/Program.cs
```diff
@@ -0,0 +1,22 @@
+﻿namespace ConcurrentQueueInCSharp
+{
+    internal class Program
+    {
+        static void Main(string[] args)
```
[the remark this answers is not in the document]
MichalBrylKaminski: Done.

collections-csharp/ConcurrentQueueInCSharp/ConcurrentQueueInCSharp.Console/Consumer.cs
MichalBrylKaminski: Indeed and I actually missed the point why I was using it in that way. Fixed.

PR #1427, opened 2023-09-29
CM-720: Initial code for the article

MPrtenjak:
Pull request for the article Merging Multiple PDFs Using the iText Library.


dotnet-client-libraries/ITextMergingPDFs/ITextMergingPDFs/Merger.cs
```diff
@@ -41,17 +42,17 @@ public string[] SplitDocument(string pdfFile)
...
         private static string[] CheckParameters(string[] pdfFiles, string mergedPdfFileName)
         {
             var existingFiles = pdfFiles.Where(File.Exists).ToArray();
 
-            if (existingFiles is null || existingFiles.Length == 0)
+            if (existingFiles.Length == 0)
                 throw new ArgumentOutOfRangeException(nameof(pdfFiles));
 
-            if (string.IsNullOrEmpty(mergedPdfFileName))
+            if (string.IsNullOrWhiteSpace(mergedPdfFileName))
```
jshergal, line 55: This should probably be the first check in the method - the fail fast principle, as before this we are doing work, walking an array and then allocating a new one.

dotnet-client-libraries/ITextMergingPDFs/ITextMergingPDFs/Merger.cs
```diff
@@ -41,17 +42,17 @@ public string[] SplitDocument(string pdfFile)
 
         private string GetFullFileName(string mergedPdfFileName)
         {
-            return Path.Combine(_path, mergedPdfFileName);
+            return System.IO.Path.Combine(_path, mergedPdfFileName);
         }
 
         private static string[] CheckParameters(string[] pdfFiles, string mergedPdfFileName)
```
jshergal, line 48: Does this really need to return an array. It seems like this should just be void and assuming it passed without throwing an exception, then all is well with the array of file names and with the merged file name.  Then the check can be changed to something like this:
`if (pdfFiles.Any(file => !File.Exists(file)))`
`    throw new ArgumentOutOfRangeException(nameof(pdfFiles));`

dotnet-client-libraries/ITextMergingPDFs/ITextMergingPDFs/Merger.cs
MPrtenjak: Repaired

dotnet-client-libraries/ITextMergingPDFs/ITextMergingPDFs/Merger.cs
MPrtenjak: Yes, we need the array as it is the input for the next function, which will loop through it and process each document.

jshergal: @MPrtenjak - but you don't need the new array even in the place where it is being used. Basically you can use the same array that is being passed into this method. There really isn't any reason to copy the array. As far as I could see in the location where it is being used, it is not being modified, so we are basically just making a copy of the array to loop through it.

dotnet-client-libraries/ITextMergingPDFs/ITextMergingPDFs/Merger.cs
```diff
@@ -0,0 +1,97 @@
...
+            var existingFiles = pdfFiles.Where(File.Exists).ToArray();
+            if (existingFiles.Length == 0)
+                throw new ArgumentOutOfRangeException(nameof(pdfFiles));
+
+            return existingFiles;
+        }
+
+        private string DoMerging(string[] pdfFiles, string mergedPdfFileName, CopyPages copyPages)
+        {
+            var existingFiles = CheckParameters(pdfFiles, mergedPdfFileName);
```
jshergal, line 62: @MPrtenjak - with the suggested change above, this line can change to just `CheckParameters` or be renamed to `ValidateParameters` if you feel it would be more clear since it will either pass or throw.

dotnet-client-libraries/ITextMergingPDFs/ITextMergingPDFs/Merger.cs
```diff
@@ -0,0 +1,97 @@
...
+            string fullFileName = GetFullFileName(mergedPdfFileName);
+
+            using var writer = new PdfWriter(fullFileName);
+            using var mergedPdfDocument = new PdfDocument(writer);
+
+            var pdfMerger = new PdfMerger(mergedPdfDocument);
+            foreach (string file in existingFiles)
```
jshergal, line 69: @MPrtenjak - then in this line we change existingFiles to pdfFiles and just iterate the passed in array, as we know it is valid.
We can also change this to a var here

dotnet-client-libraries/ITextMergingPDFs/ITextMergingPDFs/Merger.cs
MPrtenjak: Yes, you are right. 

I would never have done it like that, as this is such a small optimization that it brings nothing to the table. On the other hand, we are now changing the array, and we are making an impure function which, in my book, is way worse than a little optimization.

But anyhow. I have repaired the code yes the code is still correct and a little bit faster now.

dotnet-client-libraries/ITextMergingPDFs/ITextMergingPDFs/Merger.cs
MPrtenjak: Repaired

dotnet-client-libraries/ITextMergingPDFs/ITextMergingPDFs/Merger.cs
MPrtenjak: Repaired

dotnet-client-libraries/ITextMergingPDFs/ITextMergingPDFs/Merger.cs
jshergal: @MPrtenjak - I am not sure what you mean by we are making an "impure function", as by definition an impure function would be one which mutates data outside of its scope. The "check parameters" method is not mutating any data, just validating the input and being done as a separate method to simplify the code within the merge method. The merge method is also not mutating the array, simply iterating it. So I am not sure what you mean by "we are now changing the array". If there is something I am missing from the code that I am not seeing, please let me know. All I see in the "DoMerging" method is a foreach that is looping over the filenames contained within the pdfFiles array. Iteration is not mutation.

MPrtenjak: ☺️ It is ok. I have changed the code. Code is functionally equal, so it does not matter.  

I just wanted to say that I would have never written it like that.

PR #1429, opened 2023-09-29
CM-239 How to build a query string for a URL in C#?

karthikns9:


aspnetcore-webapi/BuildQueryString/BooksAPI/Controllers/BooksController.cs
```diff
@@ -0,0 +1,18 @@
...
+{
+    [ApiController]
+    [Route("api/[controller]")]
+    public class BooksController : ControllerBase
+    {
+
+        [HttpGet]
+        public ActionResult Get(string author, string language)
+        {
+            string bookDetails = $"Author: {author}, Language:{language}";
```
jshergal, line 13: Use var here
Second option would be to drop the variable altogether and put the string directly in the `return Ok()`

aspnetcore-webapi/BuildQueryString/BuildQueryString/BooksApiService.cs
```diff
@@ -0,0 +1,107 @@
...
+        public BooksApiService(IHttpClientWrapper httpClientWrapper)
+        {
+            _httpClientWrapper = httpClientWrapper;
+        }
+
+        public async Task<string> HttpGetAsync(string apiUrl)
+        {
+            try
+            {
+                HttpResponseMessage response = await _httpClientWrapper.GetAsync(apiUrl);
```
jshergal, line 17: Use var

aspnetcore-webapi/BuildQueryString/BuildQueryString/HttpClientWrapper.cs
```diff
@@ -0,0 +1,12 @@
+﻿namespace BuildQueryString
+{
+    public class HttpClientWrapper : IHttpClientWrapper
+    {
+        public async Task<HttpResponseMessage> GetAsync(string requestUri)
+        {
+            using HttpClient _httpClient = new();
+
+            return await _httpClient.GetAsync(requestUri);
```
jshergal, line 9: This block of code is presenting a very bad practice in how the HttpClient is being used. Here is one article discussing it, but in short it can lead to socket exhaustion:
https://www.aspnetmonsters.com/2016/08/2016-08-27-httpclientwrong/

aspnetcore-webapi/BuildQueryString/BooksAPI/Controllers/BooksController.cs
karthikns9: I've put the  string directly in the return Ok()

aspnetcore-webapi/BuildQueryString/BuildQueryString/BooksApiService.cs
karthikns9: Updated

aspnetcore-webapi/BuildQueryString/BuildQueryString/HttpClientWrapper.cs
karthikns9: Thank you for your insights. I've removed the using block.

aspnetcore-webapi/BuildQueryString/BuildQueryString/HttpClientWrapper.cs
```diff
@@ -0,0 +1,12 @@
+﻿namespace BuildQueryString
+{
+    public class HttpClientWrapper : IHttpClientWrapper
+    {
+        private static readonly HttpClient _httpClient = new ();
```
rbychiang, line 5: Remove the space between 'new' and '()'.

karthikns9: Updated

rbychiang: Can you please confirm if you have committed this change?

karthikns9: It should reflect now.

PR #1449, opened 2023-10-06
Local Functions in C#

shayar:


jshergal:
@shayar - It appears you have two open PR's for this JIRA, but that PR #1455 is the correct one. Could you please close the one that is not needed. Thank you.

PR #1450, opened 2023-10-07
CM-340: Set the precision for decimal numbers in c#

yohanym95:


jshergal:
@yohanym95 - Would you please close this PR and create a new one containing only the code necessary for the article (as per author guidelines). For each new article we work on, we should create a new branch that contains only code for the current article. Thanks 🙂

yohanym95:
> @yohanym95 - Would you please close this PR and create a new one containing only the code necessary for the article (as per author guidelines). For each new article we work on, we should create a new branch that contains only code for the current article. Thanks 🙂

I will create new PR. Thanks for the update.

PR #1451, opened 2023-10-07
CM-340: Set the precision for decimal numbers in c#

yohanym95:


WeirdBeardDev (commented):
* None of the code is used in the article. I believe the code in the PR should be the same code as in the article.
* Most of the code is formatted well. In `StringFormat.cs` there should be spaces in and around the `NumberFormatInfo` setup, i.e., `NumberFormatInfo() { NumberDecimalDigits = 2 };`

PR #1453, opened 2023-10-07
Custom naming policy

TomBak5:
Jira no- CM-786


jshergal:
@TomBak5 - Please ensure that the PR only contains code for the article. This PR contains code from multiple folders and projects. The PR should contain only commits related to the article (i.e. only code in the  
json-csharp/CustomNamingPolicy folder)
Also, please note from the author guidelines that the PR should contain the Jira in the title:
image
So this PR should be named something like: CM-786: Custom Naming Policy

PR #1456, opened 2023-10-08
CM-531: Read And Write Windows Registry In C#

bozo-spoljaric:


jshergal:
Registry keys implement IDisposable, we need to be sure that we are disposing them.

PR #1467, opened 2023-10-15
CM-825: Event-Driven Architecture in C#

lasaetarubia:


csharp-advanced-topics/EventsDrivenArchitectureInCsharp/LightControlService/LightSwitchEventSubscriber.cs
```diff
@@ -0,0 +1,45 @@
...
+using MassTransit;
+
+namespace LightControlService;
+
+public class LightSwitchEventSubscriber : IConsumer<LightSwitchEvent>
+{
+    public async Task Consume(ConsumeContext<LightSwitchEvent> context)
+    {
+        var lightEvent = context.Message;
+        bool success = await ControlLightsAsync(lightEvent);
```
mahrncic, line 11: Use `var` where possible, also since we are using this in an `if` statement later on, let's modify the naming here to be in a form of a question, eg. `isSuccessful`

csharp-advanced-topics/EventsDrivenArchitectureInCsharp/SharedLibrary/LightSwitchEvent.cs
```diff
@@ -0,0 +1,7 @@
+﻿namespace Events;
+
+public class LightSwitchEvent
```
mahrncic, line 3: Why not use records?

csharp-advanced-topics/EventsDrivenArchitectureInCsharp/SharedLibrary/ThermostatTempChangeEvent.cs
```diff
@@ -0,0 +1,7 @@
+﻿namespace Events;
+
+public class ThermostatTempChangeEvent
```
mahrncic, line 3: Same as above, should we use records for this one?

csharp-advanced-topics/EventsDrivenArchitectureInCsharp/ThermostatControlService/ThermostatEventSubscriber.cs
```diff
@@ -0,0 +1,39 @@
...
+
+namespace ThermostatControlService;
+
+public class ThermostatEventSubscriber : IConsumer<ThermostatTempChangeEvent>
+{
+    public async Task Consume(ConsumeContext<ThermostatTempChangeEvent> context)
+    {
+        var thermostatEvent = context.Message;
+
+        bool success = await AdjustThermostatAsync(thermostatEvent);
```
mahrncic, line 12: Use `var`

csharp-advanced-topics/EventsDrivenArchitectureInCsharp/SharedLibrary/ThermostatTempChangeEvent.cs
```diff
@@ -0,0 +1,7 @@
...
+{
+    public Guid CorrelationId { get; set; }
+    public decimal NewTemperature { get; set; }
```
mahrncic, line 6: Instead of `NewTemperature`, I would call this property just `Temperature`. Since it is the event and it has the `Change` word in its name, it's implicitly clear that this will correspond to the "new" temperature.

mahrncic (changes requested):

mahrncic (approved):

PR #1484, opened 2023-10-22
CM-188: How to convert string to char in C#?

Deltas-Lee:
* Added code to the Program class to demonstrate various methods for converting strings to characters and manipulating character arrays.
* Added the unit tests to ensure the correctness of these methods.
* Additionally, added AllowUnsafeBclocks in the .csproj files

### DIFF
--- a/strings-csharp/ConvertStringToCharInCSharp/ConvertStringToChar/ConvertStringToChar.cs
+++ b/strings-csharp/ConvertStringToCharInCSharp/ConvertStringToChar/ConvertStringToChar.cs
@@ -1,0 +1,129 @@
+using System;
+using System.Linq;
+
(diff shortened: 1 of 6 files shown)


strings-csharp/ConvertStringToCharInCSharp/ConvertStringToChar/ConvertStringToChar.cs
```diff
@@ -0,0 +1,92 @@
...
+
+
+        }
+
+        public static char ConvertUsingCharParse(string input)
+        {
+            return Char.Parse(input);
+        }
+
+
```
rbychiang, line 44: Please remove the extra newlines

PR #1486, opened 2023-10-23
CM-95 ConcurrentStack<T> in C#

benjaminmishra:
Added example and tests for ConcurrentStack article


jshergal:
We should try to avoid using all of the Console.WriteLine methods in the tests as this will enforce a serialization of the calls, thus defeating the purpose of the parallelization going on in the code, which we are using to show how ConcurrentStack is a thread-safe collection.

benjaminmishra:
Updated the PR as per your feedback @jshergal

PR #1490, opened 2023-10-25
CM-427: How to Rename Files in a Folder with C#

Kaybangz:


csharp-intermediate-topics/RenameFilesWithCSharp/RenameFilesWithCSharp/BatchRenameFileWithErrorHandling.cs
```diff
@@ -0,0 +1,59 @@
+﻿namespace RenameFilesWithCSharp
```
Banovvv, line 1: You can use file-scoped namespaces throughout the project to reduce nesting.

csharp-intermediate-topics/RenameFilesWithCSharp/RenameFilesWithCSharp/BatchRenameFileWithErrorHandling.cs
```diff
@@ -0,0 +1,58 @@
...
+public class BatchRenameFileWithErrorHandling
+{
+    public static void RenamePhotosWithDateTime(string directoryPath)
+    {
+        try
+        {
+
+            string[] allowedExtensions = { ".jpg", ".png", ".jpeg", ".gif" };
+
+            var photoFiles = Directory.GetFiles(directoryPath)
```
jshergal, line 12: We should change this to Directory.EnumerateFiles as that returns an IEnumerable rather than an array.
We could honestly even remove the ToList call since we are just going to iterate over this in a foreach loop

csharp-intermediate-topics/RenameFilesWithCSharp/RenameFilesWithCSharp/BatchRenameFiles.cs
```diff
@@ -0,0 +1,45 @@
...
+            var newFileName = $"Image_{creationTime:yyyy-MM-dd_HHmmss}{fileExtension}";
+
+            // Creates the new file path.
+            var newFilePath = Path.Combine(directoryPath, newFileName);
+
+            // Renames the photo file to the new file path.
+            File.Move(photoFile, newFilePath);
+        }
+    }
+
```
jshergal, line 35: Extra space

csharp-intermediate-topics/RenameFilesWithCSharp/RenameFilesWithCSharp/BatchRenameFiles.cs
```diff
@@ -0,0 +1,45 @@
+﻿namespace RenameFilesWithCSharp;
+
+public class BatchRenameFiles
+{
+    public static void RenamePhotosWithDateTime(string directoryPath)
+    {
+
```
jshergal, line 7: Extra space

csharp-intermediate-topics/RenameFilesWithCSharp/RenameFilesWithCSharp/BatchRenameFiles.cs
```diff
@@ -0,0 +1,45 @@
...
+    }
+
+
+    public static void ExecuteRenamePhotosWithDateTime()
+    {
+        var directoryPath = @"C:\MyDirectory2\Images";
+
+        RenamePhotosWithDateTime(directoryPath);
+    }
+
```
jshergal, line 43: Extra space

csharp-intermediate-topics/RenameFilesWithCSharp/RenameFilesWithCSharp/BatchRenameFileWithErrorHandling.cs
```diff
@@ -0,0 +1,58 @@
...
+                var fileExtension = Path.GetExtension(photoFile);
+
+                var newFileName = $"Image_{creationTime:yyyy-MM-dd_HHmmss}{fileExtension}";
+
+                var newFilePath = Path.Combine(directoryPath, newFileName);
+
+                File.Move(photoFile, newFilePath);
+            }
+        }
+
```
jshergal, line 29: Extra space

csharp-intermediate-topics/RenameFilesWithCSharp/RenameFilesWithCSharp/BatchRenameFileWithErrorHandling.cs
```diff
@@ -0,0 +1,58 @@
...
+        catch (Exception ex)
+        {
+            HandleException(ex);
+        }
+
```
jshergal, line 34: Extra space

csharp-intermediate-topics/RenameFilesWithCSharp/RenameFilesWithCSharp/BatchRenameFileWithErrorHandling.cs
```diff
@@ -0,0 +1,58 @@
+﻿namespace RenameFilesWithCSharp;
+
+public class BatchRenameFileWithErrorHandling
+{
+    public static void RenamePhotosWithDateTime(string directoryPath)
+    {
+        try
+        {
+
```
jshergal, line 9: Extra space

csharp-intermediate-topics/RenameFilesWithCSharp/RenameFilesWithCSharp/FindingFilesInFolder.cs
```diff
@@ -0,0 +1,26 @@
+﻿namespace RenameFilesWithCSharp;
+
+public class FindingFilesInFolder
+{
+    public static List<string> FindFilesInFolder(string directoryPath)
+    {
+
+        var files = Directory.GetFiles(directoryPath, "*", SearchOption.AllDirectories);
```
jshergal, line 8: We should use EnumerateFiles since we are going to just turn this into a list. GetFiles returns a string[]

csharp-intermediate-topics/RenameFilesWithCSharp/RenameFilesWithCSharp/FindingFilesInFolder.cs
```diff
@@ -0,0 +1,26 @@
+﻿namespace RenameFilesWithCSharp;
+
+public class FindingFilesInFolder
+{
+    public static List<string> FindFilesInFolder(string directoryPath)
+    {
+
```
jshergal, line 7: Extra space

csharp-intermediate-topics/RenameFilesWithCSharp/Tests/RenameFilesUnitTests.cs
```diff
@@ -0,0 +1,132 @@
...
+        File.Create(Path.Combine(tempDirectory, "photo1.jpg")).Close();
+        File.Create(Path.Combine(tempDirectory, "photo2.png")).Close();
+
+        // Act
+        BatchRenameFiles.RenamePhotosWithDateTime(tempDirectory);
+
+        // Assert
+        Assert.True(File.Exists(Path.Combine(tempDirectory, $"Image_{DateTime.Now:yyyy-MM-dd_HHmmss}.jpg")));
+        Assert.True(File.Exists(Path.Combine(tempDirectory, $"Image_{DateTime.Now:yyyy-MM-dd_HHmmss}.png")));
+
```
jshergal, line 63: This is a brittle assertion as it is time dependent. The rename method is using the file creation time, so the test should pull the file creation time from the newly created files and use that to build the string. This test could potentially fail if the second ticks over between creation and assertion. DateTime.Now could even be different between the two assertions.

csharp-intermediate-topics/RenameFilesWithCSharp/Tests/RenameFilesUnitTests.cs
```diff
@@ -0,0 +1,132 @@
...
+        File.Create(Path.Combine(tempDirectory, "document.docx")).Close();
+
+        // Act
+        BatchRenameFiles.RenamePhotosWithDateTime(tempDirectory);
+
+        // Assert
+        Assert.False(File.Exists(Path.Combine(tempDirectory, $"Image_{DateTime.Now:yyyy-MM-dd_HHmmss}.jpg")));
+        Assert.False(File.Exists(Path.Combine(tempDirectory, $"Image_{DateTime.Now:yyyy-MM-dd_HHmmss}.png")));
+
```
jshergal, line 84: A better assertion in the context of this test name would be to read the file names from the directory, store them, then read them again after the call and assert that the filename list is the same. We should also, as part of the arrange, assert that there are no files with an allowed image extension as the name of the test is "no allowed extension exists".

csharp-intermediate-topics/RenameFilesWithCSharp/Tests/RenameFilesUnitTests.cs
```diff
@@ -0,0 +1,132 @@
...
+        // Clean up
+        Directory.Delete(tempDirectory, true);
+    }
+
+
+    [Fact]
+    public void WhenRenamingPhotosWithDateTimeAllowedExtensionsExist_ThenPhotosShouldBeRenamed()
```
jshergal, line 91: To match the "WhenBatchRenaming" test earlier, this should probably be "WhenBatchRenamingWithErrorHandling"

csharp-intermediate-topics/RenameFilesWithCSharp/Tests/RenameFilesUnitTests.cs
```diff
@@ -0,0 +1,132 @@
...
+        // Assert
+        Assert.True(File.Exists(Path.Combine(tempDirectory, $"Image_{DateTime.Now:yyyy-MM-dd_HHmmss}.jpg")));
+        Assert.True(File.Exists(Path.Combine(tempDirectory, $"Image_{DateTime.Now:yyyy-MM-dd_HHmmss}.png")));
+
+        // Clean up
+        Directory.Delete(tempDirectory, true);
+    }
+
+    [Fact]
+    public void WhenRenamingPhotosWithDateTimeAndNoAllowedExtensionsExist_ThenNoPhotosShouldBeRenamed()
```
jshergal, line 112: Probably should be BatchRenamingWithErrorHandling

csharp-intermediate-topics/RenameFilesWithCSharp/Tests/RenameFilesUnitTests.cs
```diff
@@ -0,0 +1,132 @@
...
+
+        File.Create(Path.Combine(tempDirectory, "document.docx")).Close();
+
+        // Act
+        BatchRenameFileWithErrorHandling.RenamePhotosWithDateTime(tempDirectory);
+
+        // Assert
+        Assert.False(File.Exists(Path.Combine(tempDirectory, $"Image_{DateTime.Now:yyyy-MM-dd_HHmmss}.jpg")));
+        Assert.False(File.Exists(Path.Combine(tempDirectory, $"Image_{DateTime.Now:yyyy-MM-dd_HHmmss}.png")));
+
```
jshergal, line 126: See earlier comment on similar test, the assertion here should be that no file was renamed and so we should get the list of files, run the rename method, then assert that the files all have the same name as before.

Kaybangz:
I have resolved all of the issues

csharp-intermediate-topics/RenameFilesWithCSharp/RenameFilesWithCSharp/BatchRenameFileWithErrorHandling.cs
```diff
@@ -0,0 +1,55 @@
+﻿namespace RenameFilesWithCSharp;
+
+public class BatchRenameFileWithErrorHandling
```
mahrncic, line 3: Why do we need this class since we already have the `BatchRenameFiles` class with the same logic regarding renaming files? I can see what you are trying to do with explaining error handling. However, in my mind, it is important only on the main topic of the article (renaming files), and not on the error handling. So I would completely remove this class.

csharp-intermediate-topics/RenameFilesWithCSharp/RenameFilesWithCSharp/BatchRenameFileWithErrorHandling.cs
```diff
@@ -0,0 +1,55 @@
...
+        try
+        {
+            string[] allowedExtensions = { ".jpg", ".png", ".jpeg", ".gif" };
+
+            var photoFiles = Directory.EnumerateFiles(directoryPath)
+                                    .Where(file => allowedExtensions.Contains(Path.GetExtension(file), StringComparer.OrdinalIgnoreCase));
+
+            foreach (var photoFile in photoFiles)
+            {
+                DateTime creationTime = File.GetCreationTime(photoFile);
```
mahrncic, line 16: Use `var` wherever possible 😄

csharp-intermediate-topics/RenameFilesWithCSharp/RenameFilesWithCSharp/BatchRenameFiles.cs
```diff
@@ -0,0 +1,40 @@
+﻿namespace RenameFilesWithCSharp;
+
+public class BatchRenameFiles
```
mahrncic, line 3: This whole class could be marked as `static`. Also, the name of the class isn't the best. It is using verb in its name, which is normally applied to methods, not classes. Why not call it something like `FileNameService` or `FileProcessor`?

csharp-intermediate-topics/RenameFilesWithCSharp/RenameFilesWithCSharp/BatchRenameFiles.cs
```diff
@@ -0,0 +1,40 @@
...
+
+        // Gets a list of all the photo files in the specified directory and its subdirectories, filtering out files with other extensions.
+        var photoFiles = Directory.EnumerateFiles(directoryPath)
+                                .Where(file => allowedExtensions.Contains(Path.GetExtension(file), StringComparer.OrdinalIgnoreCase));
+
+        // Iterates over the list of photo files and renames each file using its creation time and the desired file extension.
+        foreach (var photoFile in photoFiles)
+        {
+            // Gets the creation time of the photo file.
+            DateTime creationTime = File.GetCreationTime(photoFile);
```
mahrncic, line 18: Use `var` instead of `DateTime` and wherever possible 😄

csharp-intermediate-topics/RenameFilesWithCSharp/RenameFilesWithCSharp/BatchRenameFiles.cs
```diff
@@ -0,0 +1,40 @@
+﻿namespace RenameFilesWithCSharp;
+
+public class BatchRenameFiles
+{
+    public static void RenamePhotosWithDateTime(string directoryPath)
+    {
+        // An array of allowed file extensions.
```
mahrncic, line 7: No need to comment every single line of code in the class, especially if it is straightforward like this one. Too many comments usually act as noise and can distract the readers. Thus, if the code is clean and self-explanatory, you can remove all the comments.

csharp-intermediate-topics/RenameFilesWithCSharp/RenameFilesWithCSharp/BatchRenameFiles.cs
```diff
@@ -0,0 +1,40 @@
...
+
+            // Creates the new file path.
+            var newFilePath = Path.Combine(directoryPath, newFileName);
+
+            // Renames the photo file to the new file path.
+            File.Move(photoFile, newFilePath);
+        }
+    }
+
+    public static void ExecuteRenamePhotosWithDateTime()
```
mahrncic, line 34: I don't see a valid reason for having this method, why not just directly use the `RenamePhotosWithDateTime()` method? This just acts as a wrapper and adds abstraction where it is not needed :)

csharp-intermediate-topics/RenameFilesWithCSharp/RenameFilesWithCSharp/FindingFilesInFolder.cs
```diff
@@ -0,0 +1,24 @@
+﻿namespace RenameFilesWithCSharp;
+
+public class FindingFilesInFolder
```
mahrncic, line 3: Could be static

csharp-intermediate-topics/RenameFilesWithCSharp/RenameFilesWithCSharp/RenameSingleFileInFolder.cs
```diff
@@ -0,0 +1,20 @@
+﻿namespace RenameFilesWithCSharp;
+
+public class RenameSingleFileInFolder
+{
+    public static void RenameFileInDirectory(string oldFile, string newFile)
+    {
+        // Renames the old file to the new file.
```
mahrncic, line 7: No need for comment like this one, the code itself is pretty clear.

csharp-intermediate-topics/RenameFilesWithCSharp/RenameFilesWithCSharp/RenameSingleFileInFolder.cs
```diff
@@ -0,0 +1,20 @@
...
+        File.Move(oldFile, newFile);
+    }
+
+    public static void ExecuteRenameFileInDirectory()
+    {
+        var oldFile = @"C:\MyDirectory\FileOne.txt";
+
+        var newFile = @"C:\MyDirectory\FileOne.txt";
+
+
```
mahrncic, line 17: Why having two blank lines between the code?

mahrncic (changes requested):

csharp-intermediate-topics/RenameFilesWithCSharp/RenameFilesWithCSharp/FileFinderService.cs
```diff
@@ -0,0 +1,36 @@
+﻿namespace RenameFilesWithCSharp;
+
+public static class FileFinderService
+{
+    public static List<string> FindFilesInFolder(string directoryPath)
+    {
+        return Directory.EnumerateFiles(directoryPath, "*", SearchOption.AllDirectories)
+            .ToList();
+    }
```
mahrncic, line 9: Is there a reason why we are returning a `List` here and using `ToList()` instead of just returning `IEnumerable` directly?

csharp-intermediate-topics/RenameFilesWithCSharp/RenameFilesWithCSharp/FileInfoRenameService.cs
```diff
@@ -0,0 +1,19 @@
...
+        var file = new FileInfo(oldFile);
+
+        file.MoveTo(newFile);
+    }
+
+    public static void ExecuteRenameFileWithFileInfoClass()
+    {
+        var oldFile = @"C:\MyDirectory\File2.txt";
+
+        var newFile = @"C:\MyDirectory\File_v2.txt";
```
mahrncic, line 15: The code is good, but we don't need that empty line between two variable declarations.

csharp-intermediate-topics/RenameFilesWithCSharp/RenameFilesWithCSharp/FileRenameService.cs
```diff
@@ -0,0 +1,18 @@
...
+    public static void RenameFileWithFileClass(string oldFile, string newFile)
+    {
+        File.Move(oldFile, newFile);
+    }
+
+    public static void ExecuteRenameFileWithFileClass()
+    {
+        var oldFile = @"C:\MyDirectory\File1.txt";
+
+        var newFile = @"C:\MyDirectory\File_v1.txt";
```
mahrncic, line 14: Same as in the previous service, let's remove this empty line between two variable declarations.

csharp-intermediate-topics/RenameFilesWithCSharp/RenameFilesWithCSharp/PhotoRenameService.cs
```diff
@@ -0,0 +1,27 @@
+﻿namespace RenameFilesWithCSharp;
+
+public static class PhotoRenameService
+{
+    public static void RenamePhotosWithDateTime(string directoryPath)
+    {
+        var allowedExtensions = new[] { ".jpg", ".png", ".jpeg", ".gif" };
+
+        // Gets a list of all the photo files in the specified directory and its subdirectories, filtering out files with other extensions.
```
mahrncic, line 9: We always try to avoid using comments in our code, so let's remove it from here

mahrncic (changes requested):

mahrncic (approved):

mahrncic (approved):

files-csharp/RenameFilesWithCSharp/Tests/RenameFilesUnitTests.cs
```diff
@@ -1,23 +1,32 @@
...
     }
+
     public void Dispose()
     {
-        Directory.Delete(tempDirectory, true);
+        try
+        {
+            Directory.Delete(_tempDirectory, true);
+        }
+        catch (Exception ex)
```
jshergal, line 19: This should just be an empty catch clause as we are in the Dispose and do not want to throw from Dispose.
Also, just a side note, whenever catching and rethrowing the same exception, we should just say:
`throw;`
rather than:
`throw ex;`
The former (empty throw) will preserve the stack trace.

PR #1499, opened 2023-10-27
CM-821 How to truncate a String in .NET

Sammylaw23:


strings-csharp/StringTruncationInCSharp/StringTruncationInCSharp/StringHelper.cs
```diff
@@ -5,99 +5,111 @@ namespace StringTruncationInCSharp
...
+        {
+            if (string.IsNullOrEmpty(originalString) || maxLength <= 0)
+            {
+                return string.Empty;
+            }
+
+            ReadOnlySpan<char> originalStringAsSpan = originalString.AsSpan();
+            if (maxLength >= originalStringAsSpan.Length)
+            {
+                return new string(originalStringAsSpan);
```
jshergal, line 108: Shouldn't we just return originalString here like in the other methods?

strings-csharp/StringTruncationInCSharp/StringTruncationInCSharp/StringHelper.cs
```diff
@@ -5,99 +5,111 @@ namespace StringTruncationInCSharp
...
+            }
+
+            return new string(originalStringAsSpan.Slice(0, maxLength));
```
jshergal, line 111: This could just be:
`originalStringAsSpan[..maxLength].ToString();`

strings-csharp/StringTruncationInCSharp/StringTruncationInCSharp/StringHelperBenchmark.cs
```diff
@@ -3,46 +3,82 @@
...
+
+            _stringHelper.TruncateStringUsingSpan(_originalString, maxLength);
+        }
+
+        private string GetDummyStringToTestBenchmark()
+        {
+            var dummyText = "Lorem ipsum dolor sit amet, consectetuer adipiscing elit. Aenean commodo ligula eget dolor. Aenean massa. Cum sociis natoque penatibus et magnis dis parturient montes, nascetur ridiculus mus. Donec quam felis, ultricies nec, pellentesque eu, pretium quis, sem. Nulla consequat massa quis enim. Donec pede justo, fringilla vel, aliquet nec, vulputate eget, arcu. In enim justo, rhoncus ut, imperdiet a, venenatis vitae, justo. Nullam dictum felis eu pede mollis pretium. Integer tincidunt. Cras dapibus. Vivamus elementum semper nisi. Aenean vulputate eleifend tellus. Aenean leo ligula, porttitor eu, consequat vitae, eleifend ac, enim. Aliquam lorem ante, dapibus in, viverra quis, feugiat a, tellus. Phasellus viverra nulla ut metus varius laoreet. Quisque rutrum. Aenean imperdiet. Etiam ultricies nisi vel augue. Curabitur ullamcorper ultricies nisi. Nam eget dui. Etiam rhoncus. Maecenas tempus, tellus eget condimentum rhoncus, sem quam semper libero, sit amet adipiscing sem neque sed ipsum. N";
```
jshergal, line 79: No need for a variable here, just return the text directly:
`return "Lorem ipsum.......";

strings-csharp/StringTruncationInCSharp/StringTruncationInCSharp/StringHelper.cs
Sammylaw23: Okay

strings-csharp/StringTruncationInCSharp/StringTruncationInCSharp/StringHelper.cs
Sammylaw23: Yes, I agree

strings-csharp/StringTruncationInCSharp/StringTruncationInCSharp/StringHelperBenchmark.cs
Sammylaw23: Okay. I'll effect this change.

strings-csharp/StringTruncationInCSharp/StringTruncationInCSharp/StringHelper.cs
```diff
@@ -0,0 +1,115 @@
...
+        public string TruncateStringUsingSubstring(string originalString, int maxLength)
+        {
+            if (maxLength <= 0)
+            {
+                return string.Empty;
+            }
+            else if (maxLength >= originalString.Length)
+            {
+                return originalString;
+            }
```
OTybolc, line 17: In lots of methods, we have this check. We need to refactor this and extract these checks to a separate method.
We can achieve this by introducing a new private method:
`private string TruncateString(string originalString, int maxLength, Func<string, int, string> specificTruncateMethod)`
Inside of this method, we will have checks and the call of the specificTruncateMethod.
Next, in all the methods that use these checks, we need to have a local function that corresponds to the Func delegate form the above. For instance TruncateStringUsingSubstring will be like this:
```
           private string TruncateString(string originalString, int maxLength, Func<string, int, string> specificTruncateMethod)
        {
            if (maxLength <= 0)
            {
                return string.Empty;
            }
            else if (maxLength >= originalString.Length)
            {
                return originalString;
            }

            return specificTruncateMethod(originalString, maxLength);
        }
        public string TruncateStringUsingSubstring(string originalString, int maxLength)
        {
            return TruncateString(originalString, maxLength, UsingSubstring);

            string UsingSubstring(string str, int len)
            {
                return str.Substring(0, len);
            }
        }
```

OTybolc, line 17: In lots of methods, we have this check. We need to refactor this and extract these checks to a separate method.
We can achieve this by introducing a new private method:
`private string TruncateString(string originalString, int maxLength, Func<string, int, string> specificTruncateMethod)`
Inside of this method, we will have checks and the call of the specificTruncateMethod.
Next, in all the methods that use these checks, we need to have a local function that corresponds to the Func delegate form the above. For instance TruncateStringUsingSubstring will be like this:
```
public string TruncateStringUsingSubstring(string originalString, int maxLength)
{
    return TruncateString(originalString, maxLength, UsingSubstring);

    string UsingSubstring(string str, int len)
    {
        return str.Substring(0, len);
    }
}

private string TruncateString(string originalString, int maxLength, Func<string, int, string> specificTruncateMethod)
{
    if (maxLength <= 0)
    {
        return string.Empty;
    }
    else if (maxLength >= originalString.Length)
    {
        return originalString;
    }

    return specificTruncateMethod(originalString, maxLength);
}
```

strings-csharp/StringTruncationInCSharp/StringTruncationInCSharp/StringHelperBenchmark.cs
```diff
@@ -0,0 +1,84 @@
...
```
OTybolc, line 79: Since we're not interested in chars in the string at all (we only care about it's length) it's better to use new string('a', 1000). And this should be assigned directly in the `_originalString` variable

strings-csharp/StringTruncationInCSharp/StringTruncationInCSharp/StringHelperBenchmark.cs
```diff
@@ -0,0 +1,84 @@
...
+{
+    [MemoryDiagnoser]
+    [Orderer(BenchmarkDotNet.Order.SummaryOrderPolicy.FastestToSlowest)]
+    [RankColumn]
+    public class StringHelperBenchmark
+    {
+        private string _originalString = "This is a long string.";
+        private static readonly StringHelper _stringHelper = new StringHelper();
+
+        [Params(0.01, 0.10, 0.50, 0.95)]
```
OTybolc, line 13: We know the length of the initial string for benchmarking, so this can be ints.

strings-csharp/StringTruncationInCSharp/StringTruncationInCSharp/StringHelperBenchmark.cs
```diff
@@ -0,0 +1,84 @@
...
+        public void Setup()
+        {
+            _originalString = GetDummyStringToTestBenchmark();
+        }
+
+
+        [Benchmark]
+        public void TruncateStringUsingSubstring()
+        {
+            int maxLength = (int)(_originalString.Length * MaxLengthPercentage);
```
OTybolc, line 26: We don't need this. It's possible to use int param because we have a predefined string length for the benchmark test.

PR #1508, opened 2023-11-05
CM-210: How to remove illegal characters from path and filenames in C#

samhascat:


csharp-basic-topics/RemoveIllegalCharsFromPathAndFilenames/RemoveIllegalCharsFromPathAndFilenames/Program.cs
```diff
@@ -0,0 +1,93 @@
...
+            foreach (var path in paths) { if (invalidPath.IsMatch(path)) {  result.Add(path); } }
+
+            return result;
+        }
+
+        public static string ReplaceInvalidChars(string path, string repChar)
+        {
+            var invalidChars = Path.GetInvalidPathChars().Concat(Path.GetInvalidFileNameChars()).ToArray();
+
+            return invalidChars.Aggregate(path, (a, b) => a.Replace(b.ToString(),repChar));
```
fl0tschi, line 90: missing space between last colon and variable `repChar`

csharp-basic-topics/RemoveIllegalCharsFromPathAndFilenames/RemoveIllegalCharsFromPathAndFilenames/Program.cs
```diff
@@ -0,0 +1,93 @@
...
+        {
+            Regex invalidPath = new Regex("[" + Regex.Escape(new string(Path.GetInvalidPathChars()) + new string(Path.GetInvalidFileNameChars())) + "]");
+            List<string> result = new List<string>();
+
+            foreach (var path in paths) { if (invalidPath.IsMatch(path)) {  result.Add(path); } }
+
+            return result;
+        }
+
+        public static string ReplaceInvalidChars(string path, string repChar)
```
fl0tschi, line 86: Why would we replace the invalid characters with something else then empty? Is there a reason for that?

csharp-basic-topics/RemoveIllegalCharsFromPathAndFilenames/RemoveIllegalCharsFromPathAndFilenames/Program.cs
```diff
@@ -0,0 +1,93 @@
+﻿using System.Text.RegularExpressions;
+
+namespace CodeMazeGuides.CSBasic
+{
+    public class Program
```
fl0tschi, line 5: Can we actually put ALL of the static methods into a separate class? I find it not looking good using `Program` in tests.

csharp-basic-topics/RemoveIllegalCharsFromPathAndFilenames/RemoveIllegalCharsFromPathAndFilenames/Program.cs
```diff
@@ -0,0 +1,93 @@
...
+
+            foreach (var path in validPaths)
+            {
+                Console.WriteLine(path);
+            }
+        }
+
+        public static List<string> GetInvalidPaths(List<string> paths)
+        {
+            var invalidChars = Path.GetInvalidPathChars().Concat(Path.GetInvalidFileNameChars()).ToArray();
```
fl0tschi, line 40: Wouldn't be `Union()` instead of `Concat()` better? It can contain duplicates and would be therefore less performant when we duplicate the lookup.

csharp-basic-topics/RemoveIllegalCharsFromPathAndFilenames/RemoveIllegalCharsFromPathAndFilenames/Program.cs
```diff
@@ -0,0 +1,93 @@
...
+            List<string> pathList = new List<string> 
+            {
+                "C://Thisisavalidpath",
+                "D://Thi*IsINVALID/<<",
+                "E://Users/ValidPath/docs",
+                "F://Users/Photos/Cats",
+                "G://Inv@7***lidPath",
+                "H://Test/This<IS>Ano<>therInvalidPath"
+            };
+            var convCheckResult = GetInvalidPaths(pathList);
```
fl0tschi, line 18: All of these are providing different results, right? We should output all of them and not only the LINQ check result.

csharp-basic-topics/RemoveIllegalCharsFromPathAndFilenames/RemoveIllegalCharsFromPathAndFilenames/Program.cs
```diff
@@ -0,0 +1,93 @@
+﻿using System.Text.RegularExpressions;
+
+namespace CodeMazeGuides.CSBasic
```
fl0tschi, line 3: Namespace name should be the CSPROJ name.

csharp-basic-topics/RemoveIllegalCharsFromPathAndFilenames/Tests/InvalidCharsTests.cs
```diff
@@ -0,0 +1,62 @@
+using CodeMazeGuides.CSBasic;
+using NUnit.Framework;
+
+namespace Tests
```
fl0tschi, line 4: You can use file scoped namespace declaration and get rid of the curly brackets block for the namespace

csharp-basic-topics/RemoveIllegalCharsFromPathAndFilenames/RemoveIllegalCharsFromPathAndFilenames/Program.cs
fl0tschi: You can use here the top level statements approach, so instead of defining the Program class you just start writing the content of the `Main()` method block

csharp-basic-topics/RemoveIllegalCharsFromPathAndFilenames/RemoveIllegalCharsFromPathAndFilenames/Program.cs
jshergal: Better still would be a HashSet<char> that we initialize with the invalid path chars and then call set.UnionWith with the second set. Then the for loop could be changed to something like:
foreach(var path in paths)
{
  if (path.Any(c => invalid.Contains(c)) result.Add();
}

My gut feeling is that would be slightly more performant. We could go a step further in readability/clarity if we just changed the list creation and looping to:

`return paths.Where(p => p.Any(c => invalid.Contains(c)).ToList();`

I don't suspect the performance difference would be that significant (we can always test and be sure)

csharp-basic-topics/RemoveIllegalCharsFromPathAndFilenames/RemoveIllegalCharsFromPathAndFilenames/Program.cs
```diff
@@ -0,0 +1,93 @@
...
+        {
+            List<string> pathList = new List<string> 
+            {
+                "C://Thisisavalidpath",
+                "D://Thi*IsINVALID/<<",
+                "E://Users/ValidPath/docs",
+                "F://Users/Photos/Cats",
+                "G://Inv@7***lidPath",
+                "H://Test/This<IS>Ano<>therInvalidPath"
+            };
```
jshergal, line 17: Just to note if running this in a Unix/Linux environment, all of the above paths have invalid filename characters, and so the code that combines both invalid path and invalid file name chars is going to return this entire list.

We should be sure within the context of the article and our code that we are not "windows specific", since .NET is cross platform and there are overrides of the GetInvalidPath methods that are platform specific within the framework (just have a look at the runtime source code). If we want, in the code we could if block around the specific OSs and provide OS specific examples, which would be cool.

fl0tschi: How would you try to do that? Would you create two different projects one targeting windows framework, the other one linux? Same for tests? Or would you add preprocessor directives in the projects? For different input data.

jshergal:
I would just add some if blocks with the RuntimeInformation.IsOSPlatform()
and construct an appropriate list within the block. I think this part does
require a bit of research into what are valid and invalid characters on
each platform.

samhascat:
All comments actioned, some no longer applicable and others fixed. The scope of the article was updated and the code has been updated to reflect.

fl0tschi:
You have here some changes which shouldn't be there from the folder `csharp-advanced-topics/EventsDrivenArchitectureInCsharp/` please remove it from your PR

csharp-basic-topics/RemoveIllegalCharsFromPathAndFilenames/Tests/InvalidCharsTests.cs
```diff
@@ -0,0 +1,73 @@
+public class Tests
```
fl0tschi, line 1: Rename class to `InvalidCharactersTests`. Same to the filename then

csharp-basic-topics/RemoveIllegalCharsFromPathAndFilenames/Tests/InvalidCharsTests.cs
```diff
@@ -0,0 +1,73 @@
...
+        "C://User/Test/Stuff/Invalid>>>>\r<<<<--Chars",
+        "D://User/iLi\0keToIncludeINVALIDChars/*/     z"
+    };
+    static readonly List<string> TestCaseFilenames = new List<string>
+    {
+        "A://ValidPath/SomeFolder/Invalid??File??Broken.nah",
+        "I://ProgramFiles/SomeApplication/:Invalid:Filename.txt"
+    };
+    HashSet<char> invalidPathChars = new HashSet<char>(Path.GetInvalidPathChars());
+    HashSet<char> invalidFilenameChars = new HashSet<char>(Path.GetInvalidFileNameChars());
```
fl0tschi, line 15: Make this `private static readonly` please

csharp-basic-topics/RemoveIllegalCharsFromPathAndFilenames/Tests/InvalidCharsTests.cs
```diff
@@ -0,0 +1,73 @@
+public class Tests
+{
+    // known invalid paths
+    static readonly List<string> TestCaseSearch = new List<string>
```
fl0tschi, line 4: Make this `private static readonly` please

csharp-basic-topics/RemoveIllegalCharsFromPathAndFilenames/Tests/InvalidCharsTests.cs
```diff
@@ -0,0 +1,73 @@
+public class Tests
+{
+    // known invalid paths
+    static readonly List<string> TestCaseSearch = new List<string> 
+    {
+        "C://User/Test/Stuff/Invalid>>>>\r<<<<--Chars",
+        "D://User/iLi\0keToIncludeINVALIDChars/*/     z"
+    };
+    static readonly List<string> TestCaseFilenames = new List<string>
```
fl0tschi, line 9: Make this `private static readonly` please

csharp-basic-topics/RemoveIllegalCharsFromPathAndFilenames/Tests/InvalidCharsTests.cs
```diff
@@ -0,0 +1,73 @@
...
+    {
+        "A://ValidPath/SomeFolder/Invalid??File??Broken.nah",
+        "I://ProgramFiles/SomeApplication/:Invalid:Filename.txt"
+    };
+    HashSet<char> invalidPathChars = new HashSet<char>(Path.GetInvalidPathChars());
```
fl0tschi, line 14: Make this `private static readonly` please

csharp-basic-topics/RemoveIllegalCharsFromPathAndFilenames/Tests/InvalidCharsTests.cs
```diff
@@ -0,0 +1,73 @@
...
+        List<string> expected = new List<string>() 
+        {
+            "Invalid??File??Broken.nah",
+            ":Invalid:Filename.txt"
+        };
+
+        Assert.That(expected, Is.EqualTo(result));
+    }
+        
+    [Test]
```
fl0tschi, line 58: You should use TestCase here instead probably. Or TestCaseSource. It doesn't make sense to have one test and it can fail in different iterations. This ofc makes only sense if you test something specific with each file name. Otherwise, I would test it only with one filename.

csharp-basic-topics/RemoveIllegalCharsFromPathAndFilenames/RemoveIllegalCharsFromPathAndFilenames/Program.cs
```diff
@@ -0,0 +1,44 @@
...
+    "G://Inv@7**\\*lidPath/",
+    "H://Test/This<IS>Ano<>therInvalidPath\r/",
+    "I://ProgramFiles/SomeApplication/:Invalid:Filename.txt"
+};
+
+// Checking for invalid paths
+Console.WriteLine("Conventional Logic Result:");
+foreach (var str in StringMethods.CheckForInvalid(pathList, invalidPathChars)) { Console.WriteLine(str); }
+
+Console.WriteLine("\nLINQ Result:");
```
fl0tschi, line 21: I would rather do another `Console.WriteLine()` instead of using `\n`. Instead of using \n correct would be to use `Environment.NewLine` if you really want to use it.

csharp-basic-topics/RemoveIllegalCharsFromPathAndFilenames/RemoveIllegalCharsFromPathAndFilenames/StringMethods.cs
```diff
@@ -0,0 +1,62 @@
+﻿public static class StringMethods
+{
+    public static List<string> CheckForInvalid(List<string> strings, HashSet<char> criteria)
+    {
+        List<string> result = new List<string>();
+
+        foreach (var str in strings)
```
fl0tschi, line 7: Isn't this

`return strings.Where(str => criteria.Any(ch => str.Contains(ch));`

You can change `List<string>` return type to `IEnumerable<string>`

csharp-basic-topics/RemoveIllegalCharsFromPathAndFilenames/RemoveIllegalCharsFromPathAndFilenames/StringMethods.cs
```diff
@@ -0,0 +1,62 @@
...
+                    result.Add(str);
+                    break;
+                }
+            }
+        }
+
+        return result;
+    }
+
+    public static List<string> CheckForInvalidLINQ(List<string> strings, HashSet<char> criteria)
```
fl0tschi, line 22: You can change `List<string>` return type to `IEnumerable<string>` and remove the `ToList()`

csharp-basic-topics/RemoveIllegalCharsFromPathAndFilenames/RemoveIllegalCharsFromPathAndFilenames/StringMethods.cs
```diff
@@ -0,0 +1,62 @@
...
+    {
+        return strings.Where(str => str.Any(criteria.Contains)).ToList();
+    }
+
+    public static List<string> CheckForInvalidLINQHeader(List<string> strings, HashSet<char> criteria)
```
fl0tschi, line 27: You can change `List<string>` return type to `IEnumerable<string>` and remove the `ToList()`

csharp-basic-topics/RemoveIllegalCharsFromPathAndFilenames/RemoveIllegalCharsFromPathAndFilenames/StringMethods.cs
```diff
@@ -0,0 +1,62 @@
...
+    {
+        var result = from str in strings
+                     where str.Any(criteria.Contains)
+                     select str;
+
+        return result.ToList();
+    }
+
+    public static List<string> CheckForInvalidRegEx(List<string> strings, HashSet<char> criteria)
```
fl0tschi, line 36: You can change `List<string>` return type to `IEnumerable<string>`.

csharp-basic-topics/RemoveIllegalCharsFromPathAndFilenames/RemoveIllegalCharsFromPathAndFilenames/StringMethods.cs
```diff
@@ -0,0 +1,62 @@
...
+
+        return result.ToList();
+    }
+
+    public static List<string> CheckForInvalidRegEx(List<string> strings, HashSet<char> criteria) 
+    {
+        Regex invalidChars = new Regex("[" + Regex.Escape(string.Join("", criteria)) + "]");
+        List<string> result = new List<string>();
+
+        foreach (var str in strings) { if (invalidChars.IsMatch(str)) { result.Add(str); } }
```
fl0tschi, line 41: This can be

`return strings.Where(str => invalidChars.IsMatch(str));`

csharp-basic-topics/RemoveIllegalCharsFromPathAndFilenames/RemoveIllegalCharsFromPathAndFilenames/StringMethods.cs
```diff
@@ -0,0 +1,62 @@
...
+
+        return result;
+    }
+
+    public static List<string> GetFileNames(List<string> paths)
```
fl0tschi, line 46: You can change `List<string>` return type to `IEnumerable<string>`

csharp-basic-topics/RemoveIllegalCharsFromPathAndFilenames/RemoveIllegalCharsFromPathAndFilenames/StringMethods.cs
```diff
@@ -0,0 +1,62 @@
...
+    {
+        List<string> result = new List<string>();
+
+        foreach (var path in paths)
```
fl0tschi, line 50: this can be 

return paths.Select(path => Path.GetFileName(path)).Where(path => !string.IsNullOrEmpty(path));

fl0tschi (changes requested):

samhascat:
Something hasn't gone right in my most recent commit, I'll close this PR and open a new one. Content will be the same but without any issues.

samhascat:
Fixed PR #1550 (https://github.com/CodeMazeBlog/CodeMazeGuides/pull/1550).

PR #1514, opened 2023-11-06
CM705: How to Create Better Looking Console Applications With Spectre…

cjaynjoku:
….Console


dotnet-projects/CreateBetterLookingConsoleApplicationsWithSpectreConsole/Tests/BasicUsagesTest.cs
karimkod, line 1: We don't need the Given/When/Then comments when we only have one line.

dotnet-projects/CreateBetterLookingConsoleApplicationsWithSpectreConsole/Tests/OtherUsagesTest.cs
karimkod, line 1: Same, when we don't have complicated tests, we can get rid of the comments of Given/When/Then

karimkod (changes requested):

karimkod:
@cjaynjoku, the project name is very long let's change it to something shorter. 
This is preventing the project from being loaded properly on Visual Studio and Jetbrains Rider.

PR #1524, opened 2023-11-14
CM-732: How to Join Collections in MongoDB With .NET Core and an Aggregation Pipeline

housgh:


dotnet-mongodb/JoinCollectionsAggregationPipeline/JoinCollectionsAggregationPipeline/UsersDataAccess.cs
```diff
@@ -0,0 +1,39 @@
...
+        //Unwind
+        var pipelineStageTwo = pipelineStageOne.Unwind("role");
+
+        //Project
+        var projection = Builders<BsonDocument>.Projection
+            .Exclude("roleId");
+        var pipelineStageThree = pipelineStageTwo.Project(projection);
+        
+        //Cast
+        var users = await pipelineStageThree.As<User>().ToListAsync();
```
fl0tschi, line 35: As mentioned in the article review, I think I'm not 100% sure but I thought there is a way to bring all the way through this "result" type, without a need to cast it in the end. I mean the `Project()` should make it possible no?

dotnet-mongodb/JoinCollectionsAggregationPipeline/JoinCollectionsAggregationPipeline.Tests/AggregationPipelineLiveTest.cs
```diff
@@ -0,0 +1,63 @@
+﻿using JoinCollectionsAggregationPipeline.Models;
```
fl0tschi, line 1: Can be added to the usings.cs file

fl0tschi (changes requested):

PR #1546, opened 2023-11-26
CM-257: Check if a path is a file or a directory in C#

alessandro0108:
Console program and tests to check if a path is a file or a directory in C#


jshergal:
We should be sure to clean up the temp files and directories we are creating.

PR #1548, opened 2023-11-26
CM-882 How to Serialize an Object into Query String Format in C#

karthikns9:


aspnetcore-webapi/BuildQueryString/BooksAPI/BooksAPI.csproj
```diff
@@ -1,13 +1,13 @@
 <Project Sdk="Microsoft.NET.Sdk.Web">
```
karthikns9, line 1: Mistakenly, this file changed, but it has no impact anywhere. Kindly ignore.

PR #1557, opened 2023-11-28
CM-748: Execute JavaScript Using Selenium WebDriver

01vadim10:


dotnet-client-libraries/Selenium-JavaScript/Tests/JavaScriptTests.cs
```diff
@@ -0,0 +1,77 @@
...
+
+    public JavaScriptTests(WebFactory factory)
+    {
+        factory.HostUrl = _webUrl;
+        factory.CreateDefaultClient();
+
+        _driver = new ChromeDriver();
+    }
+
+    [Fact]
```
elfico, line 22: @01vadim10 
Can we remove th `LiveTest` suffix on the test names

dotnet-client-libraries/Selenium-JavaScript/Tests/JavaScriptTests.cs
```diff
@@ -0,0 +1,77 @@
...
+    public void WhenFoundButton_ThenClickLiveTest()
+    {
+        _driver.Navigate().GoToUrl(_webUrl + "/Home/Index");
+
+        var button = _driver.FindElement(By.Id("testButton"));
+
+        var js = (IJavaScriptExecutor)_driver;
```
elfico, line 51: @01vadim10 
I think it will be great if we are consistent in the naming of variables.
`jsExecutor` reads better than just the  `js` in the the other 2 test methods.

elfico (changes requested):

dotnet-client-libraries/Selenium-JavaScript/Tests/JavaScriptTests.cs
01vadim10: @elfico 
But according to the Guidelines and initial comment to this PR in Jira I should add `LiveTest` at the end of my test name because I'm using `ChromeDriver`
image
image

PR #1559, opened 2023-11-29
CM-96: Conversions in C Sharp

prospero-apps:


csharp-basic-topics/ConversionsInCSharp/ConversionsInCSharp/Note.cs
```diff
@@ -0,0 +1,97 @@
...
+    }
+
+    public char ToChar(IFormatProvider? provider)
+    {
+        return tone;
+    }
+
+    public DateTime ToDateTime(IFormatProvider? provider)
+    {
+        throw new InvalidCastException();
```
mahrncic, line 35: Why are we throwing exception here?

mahrncic (changes requested):

mahrncic (approved):

mahrncic (approved):

PR #1586, opened 2023-12-11
CM754 Implement Custom Authorization Handler in ASP.NET Core

saraelsa:
Add implementation code for CM754.


authorization-dotnet/AuthorizationPolicyProviders/AuthorizationPolicyProviders/Controllers/HomeController.cs
```diff
@@ -0,0 +1,18 @@
...
+    {
+        return View();
+    }
+
+    [LoyaltyAuthorize("PriorityCheckIn")]
+    public IActionResult PriorityCheckIn()
+    {
+        return View();
+    }
+}
```
elfico, line 18: @saraelsa 
Can we add another method here that shows the use of `[Authorize("Loyalty:PriorityCheckIn")]`

elfico (changes requested):
@saraelsa 
There are a couple of places where we have magic strings. It will be best if we extract these as constants and use them in the code.

elfico (approved):

PR #1605, opened 2023-12-14
CM-439: Secret Management in .NET

acha-r:


aspnetcore-features/SecretsManagement/SecretMgt/Secrets.cs
```diff
@@ -0,0 +1,29 @@
...
+    public class Secrets
+    {
+        public string ApiKey { get; set; }
+    }
+
+    public class SecretManager
+    {
+        private IConfiguration _configuration;
+        private readonly string? _key;
+        public Secrets GetSecrets
```
WeirdBeardDev, line 12: Could you update the PR to include the same spacing as the article.

PR #1618, opened 2023-12-15
CM-813: PeriodicTimer in C#

martinchege324:
First draft


threads-csharp/PeriodicTimerCSharp/PeriodicTimerCSharp/PeriodicTimerCSharp.csproj
```diff
@@ -0,0 +1,10 @@
+<Project Sdk="Microsoft.NET.Sdk">
+
+  <PropertyGroup>
+    <OutputType>Exe</OutputType>
+    <TargetFramework>net7.0</TargetFramework>
```
mahrncic, line 5: We should upgrade everything to .NET 8

threads-csharp/PeriodicTimerCSharp/PeriodicTimerCSharp/PeriodicTimerMethods.cs
```diff
@@ -0,0 +1,49 @@
+﻿using System.Diagnostics;
+
+namespace PeriodicTimerCSharp
```
mahrncic, line 3: Use file-scoped namespaces

threads-csharp/PeriodicTimerCSharp/PeriodicTimerCSharp/PeriodicTimerMethods.cs
```diff
@@ -0,0 +1,49 @@
...
+                    {
+                        Console.WriteLine(item);
+                    }
+
+                    Console.WriteLine($"Elapsed time: {stopWatch.Elapsed.Seconds}s and {numberOfArrays++} arrays created");
+
+                    await Task.Delay(1000);
+
+                    if (stopWatch.Elapsed.TotalSeconds > taskDuration)
+                        return timer;
```
mahrncic, line 42: Always add curly braces to if statements, even if it's only one line

threads-csharp/PeriodicTimerCSharp/PeriodicTimerCSharpTests/PeriodicTimerCSharpTests.csproj
```diff
@@ -0,0 +1,23 @@
+<Project Sdk="Microsoft.NET.Sdk">
+
+  <PropertyGroup>
+    <TargetFramework>net7.0</TargetFramework>
```
mahrncic, line 4: .NET 8

threads-csharp/PeriodicTimerCSharp/PeriodicTimerCSharpTests/PeriodicTimerUnitTests.cs
```diff
@@ -0,0 +1,26 @@
+namespace PeriodicTimerCSharpTests
```
mahrncic, line 1: file-scoped

mahrncic (changes requested):

mahrncic (approved):

threads-csharp/PeriodicTimerCSharp/PeriodicTimerCSharp/PeriodicTimerMethods.cs
```diff
@@ -0,0 +1,70 @@
...
+        {
+            array[i] = Random.Shared.Next();
+        }
+        
+        return array;
+    }
+
+    public void Start() 
+    {
+        _task = RandomArrayTimerAsync();
```
mahrncic, line 30: Shouldn't this be awaited? Since the `RandomArrayTimerAsync()` method is asynchronous,

mahrncic (changes requested):

threads-csharp/PeriodicTimerCSharp/PeriodicTimerCSharp/PeriodicTimerMethods.cs
martinchege324: Hi @mahrncic, we are invoking the Start method synchronously, since we assume that it's the first method we execute and we only use it to initialize our _task variable. On the other hand, the Stop method should be awaited to give time for any pending tasks to be completed. Let me know if it makes sense.

mahrncic (approved):

threads-csharp/PeriodicTimerCSharp/PeriodicTimerCSharp/PeriodicTimerMethods.cs
```diff
@@ -0,0 +1,71 @@
...
+            {
+                var array = CreateRandomArray(_size);
+
+                foreach (var item in array)
+                {
+                    Console.WriteLine(item);
+                }
+
+                Console.WriteLine($"Elapsed time: {stopWatch.Elapsed.Seconds}s" +
+                    $" and {numberOfArrays++} arrays created");
```
mahrncic, line 50: I am aware that you split that into two different rows with a `+` string concatenation but I don't think this is a good solution. We shouldn't compromise our code best practices because of the restrictions of our article. That said, it would be better to make the wording smaller instead of performing this type of concatenation.

Let's do it like this:
`Console.Writeline($"Created {numberOfArrays++} in {stopWatch.Elapsed.Seconds}s")`

In that way, we won't compromise our code best practices and we will have a shorter output 😄

mahrncic (changes requested):
One small change regarding `Console.Writeline` string handling

threads-csharp/PeriodicTimerCSharp/PeriodicTimerCSharp/PeriodicTimerMethods.cs
martinchege324: Hi @mahrncic, that's a great idea, I have edited the code and article.

mahrncic (approved):

PR #1635, opened 2023-12-22
CM-205: Converting string to byte array in C#

Kaybangz:


Kaybangz:
> @Kaybangz there are 2 benchmarks projects. I guess it's not intended

I apologize for that, it was an oversight. I'll try to be more thorough in the future.

PR #1652, opened 2023-12-30
CM-943: How to Check if Items of a List Exist in Another List in C#

martinchege324:
First draft

### DIFF
--- a/collections-lists/ListExistsInAnotherCSharp/ListExistsInAnotherCSharp/CompareListsMethods.cs
+++ b/collections-lists/ListExistsInAnotherCSharp/ListExistsInAnotherCSharp/CompareListsMethods.cs
@@ -23,7 +23,7 @@ public class CompareListsMethods
         {
             GenerateIntegerList(false, false, 0, 1000000),
             GenerateIntegerList(false, false, 1000001, 1000000),
-            "Random"
+            "Exclusive"
         };
 
         yield return new object[]
--- a/collections-lists/ListExistsInAnotherCSharp/ListExistsInAnotherCSharp/CompareListsMethods.cs
+++ b/collections-lists/ListExistsInAnotherCSharp/ListExistsInAnotherCSharp/CompareListsMethods.cs
@@ -12,31 +12,31 @@ public class CompareListsMethods
 {
     public IEnumerable<object[]> SampleLists()
     {
-        yield return new object[] 
-        { 
-            GenerateIntegerList(false, false, true, 10000000), 
-            GenerateIntegerList(false, false, true, 10000000), 
-            "Random" 
+        yield return new object[]
+        {
+            GenerateIntegerList(false, false, 0, 1000000),
+            GenerateIntegerList(true, false, 0, 1000000),
+            "Reversed"
         };
 
-        yield return new object[] 
-        { 
-            GenerateIntegerList(true, false, false, 10000000), 
-            GenerateIntegerList(true, false, false, 10000000), 
-            "Common" 
+        yield return new object[]
+        {
+            GenerateIntegerList(false, false, 0, 1000000),
+            GenerateIntegerList(false, false, 1000001, 1000000),
+            "Random"
         };
 
-        yield return new object[] 
-        { 
-            GenerateIntegerList(false, true, false, 10000000), 
-            GenerateIntegerList(false, false, false, 10000000), 
-            "Equal" 
+        yield return new object[]
+        {
+            GenerateIntegerList(false, false, 0, 1000000),
+            GenerateIntegerList(false, true, 1000001, 1000000),
+            "Middle"
         };
     }
 
     [ArgumentsSource(nameof(SampleLists))]
     [Benchmark]
-    public bool CompareListUsingIntersect(List<int> firstList, List<int> secondList, string listName) 
+    public bool CompareListUsingIntersect(List<int> firstList, List<int> secondList, string listName)
     {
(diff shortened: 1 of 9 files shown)


collections-lists/ListExistsInAnotherCSharp/ListExistsInAnotherCSharp/CompareListsMethods.cs
```diff
@@ -0,0 +1,133 @@
...
+        var difference = secondList.Except(firstList).ToList();
+
+        if (difference.Count == secondList.Count)
+        {
+            return elementExists;
+        }
+        else
+        {
+            elementExists = true;
+            return elementExists;
```
mahrncic, line 94: Always add a blank line before a `return` statement

collections-lists/ListExistsInAnotherCSharp/ListExistsInAnotherCSharp/CompareListsMethods.cs
```diff
@@ -0,0 +1,133 @@
...
+
+            for (int i = size/2; i < size; i++)
+            {
+                integerList.Add(_rand.Next());
+            }
+        }
+
+        return integerList;
+    }
+}
```
mahrncic, line 133: Instead of using `== true` on each `if` statement, let's modify the names of boolean inputs to `useCommonElements`, `useAllElements` and `useRandomElements` and then we can have `if (useCommonElements)` etc. instead of using the `== true`

mahrncic (changes requested):

collections-lists/ListExistsInAnotherCSharp/ListExistsInAnotherCSharp/CompareListsMethods.cs
```diff
@@ -0,0 +1,133 @@
...
+    [Benchmark]
+    public bool CompareListUsingIteration(List<int> firstList, List<int> secondList, string listName) 
+    {
+        var elementExists = false;
+
+        foreach (var item in firstList)
+        {
+            if (secondList.Contains(item))
+            {
+                elementExists = true;
```
jshergal, line 71: Just return true here. No need for the variable or the break statement

collections-lists/ListExistsInAnotherCSharp/ListExistsInAnotherCSharp/CompareListsMethods.cs
```diff
@@ -0,0 +1,133 @@
...
+                break;
+            }
+        }
+
+        return elementExists;
```
jshergal, line 76: If we reach here, we just return false

collections-lists/ListExistsInAnotherCSharp/ListExistsInAnotherCSharp/CompareListsMethods.cs
```diff
@@ -0,0 +1,133 @@
...
+    }
+
+    [ArgumentsSource(nameof(SampleLists))]
+    [Benchmark]
+    public bool CompareListUsingExcept(List<int> firstList, List<int> secondList, string listName)
+    {
+        var elementExists = false;
+
+        var difference = secondList.Except(firstList).ToList();
```
jshergal, line 85: The ToList call here invalidates and skews the benchmark. We can just use count and make this a one line method -> return secondList.Except(firstList).Count() != secondList.Count

collections-lists/ListExistsInAnotherCSharp/ListExistsInAnotherCSharp/CompareListsMethods.cs
```diff
@@ -0,0 +1,133 @@
...
+[RankColumn]
+[Orderer(SummaryOrderPolicy.FastestToSlowest)]
+[GroupBenchmarksBy(BenchmarkLogicalGroupRule.ByParams)]
+public class CompareListsMethods
+{
+    private readonly Random _rand;
+
+    public CompareListsMethods()
+    {
+        _rand = new Random();
```
jshergal, line 20: No need for the constructor, just initialize the member field...although since we are not creating Random with a seed, we should just use Random.Shared

collections-lists/ListExistsInAnotherCSharp/ListExistsInAnotherCSharp/CompareListsMethods.cs
```diff
@@ -0,0 +1,133 @@
...
+        else
+        {
+            elementExists = true;
+            return elementExists;
+        }
+    }
+
+    private List<int> GenerateIntegerList(bool commonElements, bool allElements, bool randomElements, int size)
+    {
+        var integerList = new List<int>();
```
jshergal, line 100: We should set the initial capacity, since we know it

mahrncic (approved):

mahrncic (approved):

collections-lists/ListExistsInAnotherCSharp/ListExistsInAnotherCSharp/CompareStudentsExamples.cs
```diff
@@ -0,0 +1,49 @@
+﻿namespace ListExistsInAnotherCSharp;
+
+public class CompareStudentsExamples
```
mahrncic, line 3: This class could be made static?

mahrncic (changes requested):

mahrncic (approved):

PR #1681, opened 2024-01-14
CM-994 How to Calculate the Number of Days Between Two Dates in C#

elliezub:


dotnet-datetime/CheckNumberOfDaysBetweenTwoDates/CheckNumberOfDaysBetweenTwoDates/Program.cs
```diff
@@ -0,0 +1,18 @@
+﻿using CheckNumberOfDaysBetweenTwoDates;
+class Program
+{
+    static void Main()
```
fl0tschi, line 4: You can do this file with a top-level statement, you don't need to define a class here and the `Main()` method.

PR #1692, opened 2024-01-17
CM-777: Parsing HTML With AngleSharp in C#

vnikolaew:
This PR includes the source code with examples for the article **Parsing HTML With AngleSharp in C#**.


dotnet-client-libraries/ParsingHtmlWithAngleSharp/ParsingHtmlWithAngleSharp/FileEntry.cs
```diff
@@ -0,0 +1,54 @@
...
+    {
+        _fileName = fileName;
+        _type = type;
+        _content = content;
+        _modified = DateTime.Now;
+    }
+
+    public Stream Body => _content;
+
+    public Boolean IsClosed => _content.CanRead == false;
```
elfico, line 22: @vnikolaew 
Let's use the language keywords here. 
`bool` instead of `Boolean`, `string` instead of `String` and `int` instead of `Int32` e.t.c
There are other places in the code where this should also be applied.

dotnet-client-libraries/ParsingHtmlWithAngleSharp/Tests/ParsingHtmlWithAngleSharpUnitTests.cs
```diff
@@ -0,0 +1,290 @@
...
+
+        document.Body.AppendChild(paragraphElement);
+
+        Assert.IsAssignableFrom<IHtmlParagraphElement>(document.Body.Children[^1]);
+
+        var ulElement = document.QuerySelector<IHtmlUnorderedListElement>("ul#list")!;
+        var blueLiElement = ulElement.QuerySelector<IHtmlListItemElement>("li.blue")!;
+
+        blueLiElement.Remove();
+        // OR
```
elfico, line 218: @vnikolaew remove this and other comments from the code

dotnet-client-libraries/ParsingHtmlWithAngleSharp/Tests/ParsingHtmlWithAngleSharpUnitTests.cs
```diff
@@ -0,0 +1,290 @@
...
+
+        Assert.Equal("New article content. <br> Second article row.", article.InnerHtml);
+        Assert.IsAssignableFrom<IHtmlBreakRowElement>(article.Children[0]);
+        Assert.Equal(1, article.ClassList.Length);
+        Assert.Equal("small-article", article.ClassList[0]);
+        Assert.NotEmpty(article.Id);
+        Assert.DoesNotContain(article.Attributes, attr => attr.Name == "data-category");
+
+        var writer = new StringWriter();
+        Console.SetOut(writer);
```
elfico, line 244: If we don't need these, we can remove them. Including the other `Console` related code.

elfico (changes requested):

dotnet-client-libraries/ParsingHtmlWithAngleSharp/Tests/ParsingHtmlWithAngleSharpUnitTests.cs
vnikolaew: Is here the console logging the problem? Because in these tests, I'm logging into an in-memory `StringWriter` and not the actual console ...

vnikolaew:
@elfico I've made the necessary changes that you have outlined, can you confirm that everything is alright now?

dotnet-client-libraries/ParsingHtmlWithAngleSharp/Tests/ParsingHtmlWithAngleSharpUnitTests.cs
elfico: @vnikolaew 
This case might not apply. But there is another instance where the `Console.WriteLine` is redundant.

dotnet-client-libraries/ParsingHtmlWithAngleSharp/Tests/ParsingHtmlWithAngleSharpUnitTests.cs
```diff
@@ -0,0 +1,296 @@
...
+
+        var document = await context.OpenAsync(new Url(booksCatalogUrl));
+
+        var booksSection = document.QuerySelector<IHtmlElement>("div.page_inner section")!;
+        var bookInfoArticles = booksSection
+            .QuerySelectorAll<IHtmlElement>("li > article.product_pod")
+            .ToCollection();
+
+        var books = bookInfoArticles.Select(ToBook).ToList();
+        Console.WriteLine(books.Count()); // 20
```
elfico, line 193: @vnikolaew 
here is the instance I mentioned.

vnikolaew: @elfico Okay, shall I remove this section then?

elfico: @vnikolaew ,
Yes, I think we should remove this ` Console.WriteLine(books.Count()); // 20` as it is redundant in this case. The others look ok

vnikolaew: @elfico I suppose I can remove it from one of the article's code snippets as well, right?

elfico (approved):

dotnet-client-libraries/ParsingHtmlWithAngleSharp/ParsingHtmlWithAngleSharp/FileEntry.cs
elfico, line 1: @vnikolaew 
Are we using this file anywhere in this project? I couldn't find a reference for it.

dotnet-client-libraries/ParsingHtmlWithAngleSharp/Tests/ParsingHtmlWithAngleSharpUnitTests.cs
```diff
@@ -0,0 +1,295 @@
...
+        var context = BrowsingContext.New(config);
+
+        var document = await context.OpenAsync(req => req.Content(Html));
+
+        var section = document.Body!.QuerySelector<IHtmlElement>("section")!;
+
+        var sectionInnerHtml = section.InnerHtml;
+        var sectionTextContent = section.TextContent;
+        var sectionAttributes = section.Attributes;
+        var sectionChildren = section.Children; // IHtmlCollection<IElement>
```
elfico, line 140: @vnikolaew , the comment here is redundant.

elfico (changes requested):

dotnet-client-libraries/ParsingHtmlWithAngleSharp/ParsingHtmlWithAngleSharp/FileEntry.cs
vnikolaew: @elfico You are right, it is not being used anywhere. Shall I remove it in this case?

dotnet-client-libraries/ParsingHtmlWithAngleSharp/Tests/ParsingHtmlWithAngleSharpUnitTests.cs
elfico: @vnikolaew 
this is still pending

elfico (approved):

PR #1698, opened 2024-01-20
CM-954: Comparing for and foreach Loop in C#

martinchege324:
First draft and code


csharp-basic-topics/ForVsForeachInCSharp/ForVsForeachInCSharp/GenerateData.cs
```diff
@@ -0,0 +1,64 @@
+﻿namespace ForVsForeachInCSharp;
+
+public class GenerateData
```
jshergal, line 3: Everything here can be made static

csharp-basic-topics/ForVsForeachInCSharp/ForVsForeachInCSharp/ForVsForeachExamples.cs
```diff
@@ -0,0 +1,126 @@
...
+namespace ForVsForeachInCSharp;
+
+[MemoryDiagnoser]
+[RankColumn]
+[Orderer(SummaryOrderPolicy.FastestToSlowest)]
+[GroupBenchmarksBy(BenchmarkLogicalGroupRule.ByParams)]
+
+public class ForVsForeachExamples
+{
+    private readonly GenerateData _generateData = new();
```
jshergal, line 14: Once the other items are made static, this can go away

csharp-basic-topics/ForVsForeachInCSharp/ForVsForeachInCSharp/ForVsForeachExamples.cs
```diff
@@ -0,0 +1,126 @@
...
+    private readonly int[] _intArray;
+    private readonly List<int> _intList;
+    private readonly Stack<int> _intStack;
+    private readonly Queue<int> _intQueue;
+    private readonly Dictionary<int, int> _intDictionary;
```
jshergal, line 19: Initialize these directly inline from the static GenerateData class

csharp-basic-topics/ForVsForeachInCSharp/ForVsForeachInCSharp/ForVsForeachExamples.cs
```diff
@@ -0,0 +1,126 @@
...
+
+    public ForVsForeachExamples()
+    {
+        _intArray = _generateData.GenerateRandomArray(100000);
+        _intList = _generateData.GenerateRandomList(100000);
+        _intStack = _generateData.GenerateRandomStack(100000);
+        _intQueue = _generateData.GenerateRandomQueue(100000);
+        _intDictionary = _generateData.GenerateRandomDictionary(100000);
+    }
```
jshergal, line 28: Once the other changes are made this can go away. Note, we should also add a constant for the size of the collections to generate, since we are using the same value in each of them (which is needed for the purpose of the benchmark). Then we just use the constant in the calls to the generate methods.

csharp-basic-topics/ForVsForeachInCSharp/ForVsForeachInCSharp/ForVsForeachExamples.cs
```diff
@@ -0,0 +1,126 @@
...
+
+    [Benchmark]
+    public void PrintArrayUsingForLoop()
+    {
+        for (int i = 0; i < _intArray.Length; i++)
+        {
+            Console.WriteLine(_intArray[i]);
+        }
+    }
+
```
jshergal, line 47: This applies to all of the benchmark methods. We should not be using Console.WriteLine in a benchmark. We should change the method signature to return an int or a long if that is better and then change the benchmark methods to compute a sum or an average.

csharp-basic-topics/ForVsForeachInCSharp/ForVsForeachInCSharp/Program.cs
```diff
@@ -0,0 +1,4 @@
+﻿using BenchmarkDotNet.Running;
+using ForVsForeachInCSharp;
+
+var summary = BenchmarkRunner.Run<ForVsForeachExamples>();
```
jshergal, line 4: No need for the unused variable, just call the method

PR #1737, opened 2024-01-29
CM-932: How to Get the Number of Files in a Folder with C#

Kaybangz:


files-csharp/CountFilesInaFolderUsingCSharp/CountFilesInaFolderTests/CountFilesInaFolderTests.csproj
```diff
@@ -0,0 +1,37 @@
...
+    <IsTestProject>true</IsTestProject>
+  </PropertyGroup>
+
+  <PropertyGroup Condition="'$(Configuration)|$(Platform)'=='Debug|AnyCPU'">
+    <DefineConstants>$(DefineConstants);WINDOWS</DefineConstants>
+  </PropertyGroup>
+
+  <PropertyGroup Condition="'$(Configuration)|$(Platform)'=='Release|AnyCPU'">
+    <DefineConstants>$(DefineConstants);WINDOWS</DefineConstants>
+  </PropertyGroup>
```
jshergal, line 18: I don't think this constant define is necessary. The code will build just fine. We just need to check the platform (https://learn.microsoft.com/en-us/dotnet/api/system.operatingsystem.iswindows?view=net-8.0) before trying to execute the windows code. We could also use that check in the unit test itself to just pass the test if it is not on windows.

files-csharp/CountFilesInaFolderUsingCSharp/CountFilesInaFolderBenchmarks/FileCountingBenchmark.cs
```diff
@@ -0,0 +1,76 @@
...
+            File.WriteAllText(filePath, "Test content");
+        }
+
+        return directory;
+    }
+
+    [Benchmark]
+    public int DirectoryGetFiles_InMemory()
+    {
+        var directory = CreateSimulatedDirectory(FileCount);
```
jshergal, line 34: We should create the temp directory and the files within it just once per file count rather than every time we run the benchmark code. Doing it the way we have written here will return inconsistent and invalid benchmark results. We can just use the [GlobalSetup] to generate the directory of files, similar to how we are using [GlobalCleanup] to remove them.

files-csharp/CountFilesInaFolderUsingCSharp/CountFilesInaFolderBenchmarks/Program.cs
```diff
@@ -0,0 +1,4 @@
+﻿using BenchmarkDotNet.Running;
+using CountFilesInaFolderBenchmarks;
+
+var summary = BenchmarkRunner.Run<FileCountingBenchmark>();
```
jshergal, line 4: No need to store the summary since we are not doing anything with it. We can just call Run directly

files-csharp/CountFilesInaFolderUsingCSharp/CountFilesInaFolderBenchmarks/FileCountingBenchmark.cs
```diff
@@ -0,0 +1,76 @@
...
+            foreach (var filePath in Directory.EnumerateFiles(_directoryPath))
+            {
+                File.Delete(filePath);
+            }
+
+            if (Directory.Exists(_directoryPath))
+            {
+                Directory.Delete(_directoryPath, true);
+            }
+        }
```
jshergal, line 74: We should be able to just do Directory.Delete(path, true) here without the first walk to delete all the files (like we are doing in Program.cs).

files-csharp/CountFilesInaFolderUsingCSharp/CountFilesInaFolderBenchmarks/FileCountingBenchmark.cs
```diff
@@ -0,0 +1,76 @@
...
+    public int FileCount;
+
+    public DirectoryInfo CreateSimulatedDirectory(int fileCount)
+    {
+        var directory = Directory.CreateDirectory(_directoryPath);
+
+        for (int i = 0; i < fileCount; i++)
+        {
+            var filePath = Path.Combine(directory.FullName, $"File{i}.txt");
+            File.WriteAllText(filePath, "Test content");
```
jshergal, line 25: This will execute faster and be less disk intensive if we just create empty (0 byte) files. We can do that like this:
`File.Create(filePath).Dispose();`

files-csharp/CountFilesInaFolderUsingCSharp/CountFilesInaFolderTests/CountFilesInaFolderTests.csproj
Kaybangz: I originally defined them to allow me to use the preprocessor directives to conditionally compile the `GivenDirectoryWithFiles_WhenUsingWinAPI_ThenReturnCorrectFileCount()` unit test that is now commented out because it was failing to pass the Jenkins CI build. But I'll get rid of it.

jshergal: This is all you need to do to prevent the unit test failing when running on non-Windows OSes:
image

files-csharp/CountFilesInaFolderUsingCSharp/CountFilesInaFolderTests/FileCounterUnitTests.cs
```diff
@@ -0,0 +1,51 @@
...
+    public FileCounterUnitTests()
+    {
+        _tempDirectory = Path.Combine(Path.GetTempPath(), Path.GetRandomFileName());
+        Directory.CreateDirectory(_tempDirectory);
+
+        for (var i = 0; i < 1000; i++)
+        {
+            File.Create(Path.Combine(_tempDirectory, $"file{i}.txt")).Dispose();
+        }
+    }
```
jshergal, line 17: This along with the Dispose should be moved into a ClassFixture so that we only generate the files once and cleanup once. As currently written we are generating the 1000 files on each test execution.

PR #1747, opened 2024-02-01
CM-1001 : How to Reverse a Number as an Integer and Not as a String in C#

AditiSaxena0289:
Initial Commit for Article **CM-1001 : How to Reverse a Number as an Integer and Not as a String in C#**

Jira Link - [CM-1001](https://content.code-maze.com/browse/CM-1001)

To run the code: 
\CodeMazeGuides\numbers-csharp\ReverseNumberAsInteger\reversenumberasinteger> dotnet run

### DIFF
--- a/numbers-csharp/ReverseNumberAsInteger/ReverseNumberAsInteger/ReverseNumbers.cs
+++ b/numbers-csharp/ReverseNumberAsInteger/ReverseNumberAsInteger/ReverseNumbers.cs
@@ -55,7 +55,7 @@
         public static int ReverseUsingMathPow(int num)
         {
             int maxValueDiv10 = int.MaxValue / 10;
-            bool isNegative = num < 0;
+            var isNegative = num < 0;
             num = isNegative ? -num : num;
 
             if (num == int.MaxValue || num == int.MinValue || num == 0)
@@ -99,7 +99,7 @@
                 return 0;
             }
 
-            bool isNegative = num < 0;
+            var isNegative = num < 0;
             num = isNegative ? -num : num;
             int reversedNumber = 0;
             int totalNumOfDigits = (int)Math.Floor(Math.Log10(num)) + 1;
@@ -148,7 +148,7 @@
                 return 0;
             }
 
-            bool isNegative = num < 0;
+            var isNegative = num < 0;
             ReadOnlySpan<char> numChars = Math.Abs(num).ToString().AsSpan();
             int length = numChars.Length;
             Span<char> reversedChars = stackalloc char[length];            
--- a/numbers-csharp/ReverseNumberAsInteger/ReverseNumberAsInteger/ReverseNumbers.cs
+++ b/numbers-csharp/ReverseNumberAsInteger/ReverseNumberAsInteger/ReverseNumbers.cs
@@ -16,7 +16,7 @@
 
             while (num != 0)
             {
-                int remainder = num % 10;
+                var remainder = num % 10;
 
                 if (reversedNumber > maxValueDiv10 || reversedNumber < minValueDiv10)
                 {
@@ -45,7 +45,7 @@
                 return 0;
             }
 
-            int remainder = num % 10;
+            var remainder = num % 10;
 
             reversedNumber = reversedNumber * 10 + remainder;
 
@@ -75,7 +75,7 @@
 
             for (int i = length - 1; i >= 0; i--)
             {
-                int remainder = num % 10;
+                var remainder = num % 10;
 
                 if ((remainder * powersOf10[i] / 10) >= maxValueDiv10)
                 {
--- a/numbers-csharp/ReverseNumberAsInteger/ReverseNumberAsInteger/ReverseNumbers.cs
+++ b/numbers-csharp/ReverseNumberAsInteger/ReverseNumberAsInteger/ReverseNumbers.cs
@@ -176,13 +176,11 @@ namespace ReverseNumberAsInteger
             int reversedNumber = num.ToString()
                 .Reverse()
                 .Select(digit => digit - '0')
-                       // .Aggregate(0, (result, digit) => result * 10 + digit - '0');
                        .Aggregate(0, (result, digit) =>
                        {
-                           // Check for potential overflow
                            if (CheckForOverflow(result))
                            {
-                               return 0; // Handling overflow
+                               return 0; 
                            }
                            return result * 10 + digit;
                        });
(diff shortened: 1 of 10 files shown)


numbers-csharp/ReverseNumberAsInteger/ReverseNumberAsInteger/ReverseNumbers.cs
```diff
@@ -0,0 +1,226 @@
...
+            Array.Reverse(charArray);
+
+            string reversedStr = new string(charArray);
+
+            if (string.IsNullOrEmpty(reversedStr))
+            {
+                return 0;
+            }
+
+            return Int32.Parse(reversedStr);
```
jshergal, line 207: Rather than create a char array on line 197, we should take a span.

Next, since we know the max string length is 11 characters, we can stackalloc a char array as a span with a length equal to the string span.

We can CopyTo from the string span to the stackalloc span and then reverse that, or we can walk the string span in reverse copying the characters one at a time. I would suspect the CopyTo followed by span.Reverse() would be faster.

Then we can call int.Parse(stacllocSpan) and we have saved both an array allocation and a string allocation. This approach can also be adapted into the BigInteger case.

I would have a question on what the intended behavior here is when the number is negative, as it will have a leading minus sign, which in the reversal is now going to be the last character. This will cause an exception when calling int.Parse. (Note, I did not take a look at the article, so maybe this method is intended only for non-negative values). It does seem that this case is being handled properly in the BigInteger version, so that should be addressed here as well. I also didn't find a unit test for this method, which would be good to have, especially considering we need to handle the negative case properly.

A secondary note regarding lines 200 and 202. The conditional doesn't seem to follow here. We know that we have been given an integer and have converted it to a string. This means that it has at a minimum 1 character. We then created an array and reversed it and have created a new string from the array...so:
1. It doesn't follow that we would end up with an empty string at line 196, so we can't have an empty one at line 202.
2. If there is some edge case I am not seeing that would produce that case, we already know it before we even convert the string to a char array and could bail early.
3. We know that reversedStr is not null at line 202 because we just created it at line 200, worst case it is empty, but I can't come up with how that could occur.

AditiSaxena0289: Thank you for your valuable suggestions! I appreciate your feedback and have carefully considered your points.

Regarding the `ReverseAsString(int num)` method, I understand your perspective. I included it in the project to demonstrate the performance difference between integer and string reversal using benchmarks. As the main focus of the article on **How to Reverse a Number as an Integer and Not as a String in C#**, I haven't incorporated this method within the article but will mention it in the benchmark summary for context. I have also excluded it from the unit test cases due to its being outside the article's core scope.

I completely agree with you about the unnecessary empty string check at lines **200** and **202**. I'll promptly remove it and update the pull request accordingly.

I'm happy to further adjust the code based on your suggestions regarding `stackalloc()` for the `ReverseAsString()` method and adding unit tests. Please let me know if these changes are still necessary, and I'll gladly implement them.

In the meantime, I'm actively incorporating the code details into the article and will keep you updated on its progress. Perhaps reviewing the final article with all code changes will provide us with a clearer picture on whether further elaboration on this method is required.

**Regarding the article** - Beyond covering different integer reversal methods, I also intend to reference the existing CodeMaze article **How to Reverse a String in C#**: https://code-maze.com/csharp-reverse-string/ to help readers grasp the differences between string and integer reversal. This includes emphasizing how underflow and overflow become considerations in integer reversal, whereas string reversal doesn't face such constraints. Additionally, even when using `BigInt` to handle large numbers, reversal as a string might offer better performance compared to integer reversal.

jshergal: @AditiSaxena0289 - no worries, I didn't check an article preview, just looked at the Jira after seeing the PR in my GitHub feed. It's a topic I am interested in and so I looked at the PR and just wanted to offer some suggestions. Also good thought regarding the reference to the string reversal article. There are definitely different edge cases to consider when talking about a number vs. a string.

Since we are benchmarking, I wanted to be sure we were benchmarking a method that was as fine-tuned as we could make it, thus the span suggestions to reduce allocations, copies, etc.

I would suggest that, if the code is being listed for a benchmark comparison, we should at least be unit testing it in the same fashion as the other methods (i.e. the same test cases/data Theory block as we are using for similar methods), so we are assured we are also benchmarking a method that is "equivalent" in outcome (for instance the case of a negative int).

Those are just my suggestions and whether I end up doing a review of the article or not - I look forward to reading it. I can see a lot of effort and research went into the PR and the test cases are really good. I appreciated seeing the thoroughness there, and so I fully expect the article to be thorough as well. Cheers!

AditiSaxena0289: @jshergal - I've implemented the stackalloc and span logic in ReverseAsString() as suggested! I've also added unit tests to ensure everything works as expected.

You can find all the changes reflected in the updated PR. I'd appreciate it if you could take a look and let me know if everything seems good.

On a separate note, I've also finished the first draft of the article for this issue!

numbers-csharp/ReverseNumberAsInteger/ReverseNumberAsInteger/ReverseNumbers.cs
```diff
@@ -0,0 +1,155 @@
...
+            if (num == int.MaxValue || num == int.MinValue || num == 0)
+            {
+                return 0;
+            }
+
+            bool isNegative = num < 0;
+            num = isNegative ? -num : num;
+
+            int length = (int)Math.Floor(Math.Log10(num) + 1);            
+            int powerOf10 = (int)Math.Pow(10, length - 1);
```
jshergal, line 63: In most performance oriented code, where there is a known limit to the powers of ten that will be necessary, the powers of 10 are precomputed into an array so they can be quickly indexed.

numbers-csharp/ReverseNumberAsInteger/ReverseNumberAsInteger/ReverseNumbers.cs
```diff
@@ -0,0 +1,155 @@
...
+            int reversedNumber = 0;
+
+            for (int i = 1; i <= length; i++)
```
jshergal, line 66: we should index from 0 to < length, rather than from 1 to length. less than typically is more optimized than less than equal (it may end up the same after the compiler does some tweaking, but less than is generally preferred.

numbers-csharp/ReverseNumberAsInteger/ReverseNumberAsInteger/ReverseNumbers.cs
```diff
@@ -0,0 +1,155 @@
...
+
+        public static int ReverseAsString(int num)
+        {
+            if (num == int.MaxValue || num == int.MinValue || num == 0)
+            {
+                return 0;
+            }
+
+            bool isNegative = num < 0;
+            char[] numChars = Math.Abs(num).ToString().ToCharArray();
```
jshergal, line 135: We should just take the string as a ReadOnlySpan<char> here so there is no additional array allocation

numbers-csharp/ReverseNumberAsInteger/ReverseNumberAsInteger/ReverseNumbers.cs
```diff
@@ -0,0 +1,155 @@
...
+            }
+
+            bool isNegative = num < 0;
+            num = isNegative ? -num : num;
+
+            int reversedNumber = 0;
+
+            int totalNumOfDigits = (int)Math.Floor(Math.Log10(num)) + 1;
+            int leftPow = (int)Math.Pow(10, totalNumOfDigits - 1);
+            int rightPow = 1;
```
jshergal, line 98: Same indexing into an array of powers of 10 applies here as well.

PR #1759, opened 2024-02-04
CM-846: How to Use StringPool to Reduce String Allocations in C#

asetil:


csharp-advanced-topics/HowToUseStringPool/HowtoUseStringPoolApi/StringPoolHelper.cs
```diff
@@ -0,0 +1,223 @@
...
+        return result;
+    }
+
+    public bool AddUser(ReadOnlySpan<char> nameSpan, ReadOnlySpan<char> emailSpan)
+    {
+        var prefixSpan = "USER_".AsSpan();
+        var prefix = _myPool.GetOrAdd(prefixSpan);
+        var name = _myPool.GetOrAdd(nameSpan);
+
+        Span<char> combined = stackalloc char[prefixSpan.Length + nameSpan.Length];
```
jshergal, line 41: We should not be using stackalloc here, unless we make this a conditional that will stackalloc when the length is small (~ < 128), otherwise will use a pooled array or allocate a new array. Using stackalloc with an unknown/caller provided argument is very dangerous.

csharp-advanced-topics/HowToUseStringPool/HowtoUseStringPoolApi/StringPoolHelper.cs
```diff
@@ -0,0 +1,223 @@
...
+
+        if (request.Headers.ContainsKey(keyValue))
+            return _myPool.GetOrAdd(request.Headers[keyValue]);
+
+        return string.Empty;
+    }
+
+    public bool CheckHeader(HttpRequest request)
+    {
+        string authorization = GetHeaderValue(request, "Authorization".AsSpan());
```
jshergal, line 106: string are implicitly convertable to ReadOnlySpan<char> so no need for the "AsSpan" before passing a string to a method that takes a ReadOnlySpan<char>

Note: I just chose this one, there are several in the PR, so please go through it and address them.

asetil: Corrected

csharp-advanced-topics/HowToUseStringPool/HowtoUseStringPoolApi/StringPoolHelper.cs
asetil: Corrected

PR #1766, opened 2024-02-05
CM-931: How to Get the HTML Content of a WebPage as a String in .NET

emekaemego:


strings-csharp/HTMLContentString/HTMLContentString/HtmlAngle.cs
```diff
@@ -0,0 +1,31 @@
...
+{
+    public async Task<string> GetHtmlAsStringAsync(string url)
+    {
+        var config = Configuration.Default.WithDefaultLoader();
+        var context = BrowsingContext.New(config);
+        var document = await context.OpenAsync(url);
+
+        return document.Body.InnerHtml;
+    }
+    public List<string> GetList(string htmlContent)
```
elfico, line 16: @emekaemego 
There should be a new line before the next function (i.e the GetList() function)

strings-csharp/HTMLContentString/HTMLContentString/HtmlHttp.cs
```diff
@@ -0,0 +1,20 @@
...
+        _httpClient = new HttpClient();
+    }
+
+    public async Task<string> GetHtmlAsStringAsync(string url)
+    {
+        var response = await _httpClient.GetAsync(url);
+        response.EnsureSuccessStatusCode();
+
+        var htmlContent = await response.Content.ReadAsStringAsync();
+        return htmlContent;
```
elfico, line 18: @emekaemego 
Also, there should be a newline before the return statement here

elfico (changes requested):

elfico (approved):

strings-csharp/HTMLContentString/HTMLContentString/HtmlHttp.cs
```diff
@@ -0,0 +1,21 @@
+namespace HTMLContentString;
+
+public class HtmlHttp
+{
+    private readonly HttpClient _httpClient;
+
+    public HtmlHttp()
+    {
+        _httpClient = new HttpClient();
```
elfico, line 9: @emekaemego 
We are instantiating HttpClient by using the new keyword, which is far from best practice. We should be using HttpClientFactory injected into the class to retrieve an instance of HttpClient. 

We only need to do a couple of things to our project:
Add a NuGet reference to Microsoft.Extensions.Hosting + Microsoft.Extensions.Http: https://learn.microsoft.com/en-us/dotnet/core/extensions/dependency-injection-usage#create-a-new-console-application

Update the Program class to use a HostApplicationBuilder and register the HttpClient: https://learn.microsoft.com/en-us/dotnet/core/extensions/dependency-injection-usage#register-services-for-di

elfico (changes requested):

elfico (approved):

PR #1767, opened 2024-02-05
CM-881: Fastest method to Check if a Date is Less Than or Equal to Today's Date in C#

cjaynjoku:


dotnet-datetime/CheckIfDateIsLessThanOrEqualToToday/CheckIfDateIsLessThanOrEqualToToday/CheckDateMethods.cs
```diff
@@ -0,0 +1,31 @@
...
+    public bool CheckWithCompare() => DateTime.Compare(_dateToCheck, DateTime.Today) <= 0;
+
+    public bool CheckWithTimeSpan()
+    {
+        TimeSpan timeDifference = _dateToCheck - DateTime.Today;
+
+        return timeDifference.TotalDays <= 0;
+    }
+
+    public bool CheckWithLINQ() => new List<DateTime> { _dateToCheck }.Any(d => d <= DateTime.Today);
```
jshergal, line 30: This method makes no sense, we should remove it. We are creating a list of one value just to use the `<=` date comparison that we already exercise on line 17. We are not adding any value to the article with this.

dotnet-datetime/CheckIfDateIsLessThanOrEqualToToday/CheckIfDateIsLessThanOrEqualToToday/CheckDateMethods.cs
```diff
@@ -0,0 +1,31 @@
...
+
+namespace CheckIfDateIsLessThanOrEqualToToday;
+
+public class CheckDateMethods
+{
+    public readonly DateTime _dateToCheck;
+
+    public CheckDateMethods(string dateString = "01/01/2024")
+    {
+        if (!DateTime.TryParseExact(dateString, "MM/dd/yyyy", CultureInfo.InvariantCulture, DateTimeStyles.None, out _dateToCheck))
```
jshergal, line 11: Note that the DateTime object that we are creating here, is of an unspecified kind, whereas `DateTime.Today` is `Local`. Whenever performing DateTime comparisons we should ensure that we are comparing like kinds:
https://learn.microsoft.com/en-us/dotnet/api/system.datetimekind?view=net-8.0

dotnet-datetime/CheckIfDateIsLessThanOrEqualToToday/CheckIfDateIsLessThanOrEqualToToday/CheckDateMethods.cs
```diff
@@ -0,0 +1,31 @@
+﻿using System.Globalization;
+
+namespace CheckIfDateIsLessThanOrEqualToToday;
+
+public class CheckDateMethods
```
elfico, line 5: @cjaynjoku 
We should rename this class as it currently reads like a method. We could rename as `DateComparison`, `DateCheck` or something different.
This naming will also be propagated across other class names that reads like methods

dotnet-datetime/CheckIfDateIsLessThanOrEqualToToday/CheckIfDateIsLessThanOrEqualToToday/CheckDateMethods.cs
```diff
@@ -0,0 +1,31 @@
...
+{
+    public readonly DateTime _dateToCheck;
```
elfico, line 7: We can make this private if we are not accessing it outside this class

elfico (changes requested):

dotnet-datetime/CheckIfDateIsLessThanOrEqualToToday/CheckIfDateIsLessThanOrEqualToToday/CheckDateMethods.cs
cjaynjoku: Hi @jshergal, is there any documentation I can read to get more information on this statement, 'Whenever performing `DateTime` comparisons, we should ensure that we are comparing like kinds'?

The link you provided states that the `DateTimeKind` members are not used in comparisons.

dotnet-datetime/CheckIfDateIsLessThanOrEqualToToday/CheckIfDateIsLessThanOrEqualToToday/CheckDateMethods.cs
```diff
@@ -0,0 +1,31 @@
+﻿using System.Globalization;
+
+namespace CheckIfDateIsLessThanOrEqualToToday;
```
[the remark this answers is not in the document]
cjaynjoku: Hi @jshergal, the `DateOnly` struct does not have a definition for `Today`, as discussed in this issue: https://github.com/dotnet/runtime/issues/53498.

So, since we are dealing with today's date, instead of having something as clear as this: `DateTime.Today`, we would have to replace it with `DateOnly.FromDateTime(DateTime.Today)`. Should we go ahead with this?

dotnet-datetime/CheckIfDateIsLessThanOrEqualToToday/CheckIfDateIsLessThanOrEqualToToday/CheckDateMethods.cs
jshergal: Hi @cjaynjoku, you are correct. They are not considered when making comparisons and that is exactly the problem. If for instance I give you a local time in the timezone UTC+10, and you call "DateTime.Today" when the code is running in UTC-10, what is the expected behavior? For example, if in UTC+10 the time is currently 09:00 on 2024.02.08, in the UTC-10 timezone the time would be 13:00 on 2024.02.07, but if we computed the time in UTC they are the same day.
That is why we need to consider the kind when performing comparisons. If we ignore it, then we should explain why we are ignoring it and what the consequences are. It is generally considered a good practice to perform date/time comparisons in UTC for just this exact reason.

dotnet-datetime/CheckIfDateIsLessThanOrEqualToToday/CheckIfDateIsLessThanOrEqualToToday/CheckDateMethods.cs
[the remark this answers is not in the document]
jshergal: @cjaynjoku yes, you create a DateOnly from a DateTime, you can use DateTime.Now, UtcNow, Today, etc.

The interesting part in the discussion with the DotNet link you shared is the comments there regarding the fact that DateOnly should not care about the timezone, so that may actually be a greater motivation to focus on DateOnly rather than DateTime, and since in the code we are creating a date from a string, we can then create today's date using `DateOnly.FromDateTime(DateTime.Today)`, and use it without having to think about kinds and timezones at all.

dotnet-datetime/CheckIfDateIsLessThanOrEqualToToday/CheckIfDateIsLessThanOrEqualToToday/CheckDateMethods.cs
cjaynjoku: Alright @jshergal. I get your point now. Thanks for pointing this out.

dotnet-datetime/CheckIfDateIsLessThanOrEqualToToday/CheckIfDateIsLessThanOrEqualToToday/CheckDateMethods.cs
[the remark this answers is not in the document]
cjaynjoku: Okay @jshergal, let's focus on just the DateOnly struct.

elfico (approved):

PR #1772, opened 2024-02-07
Funcactiondelegat

mohamedtayel1980:
Fixed Jenkins CI build error by ensuring the build command targets the correct directory containing the project or solution file. Updated build configuration for consistency across all projects.


mohamedtayel1980:
moh.moh701/moh.moh701@gmail.com 
apply feedback
1 - No need for these comments.
2 - The int can be declared as var.
3 - We should have a space between the return statement and the rest of the code.
 Same applies to code comments here:

PR #1776, opened 2024-02-08
CM-904: Fastest Way to Generate a Random Boolean in C#

MPrtenjak:
Initial code for the article


numbers-csharp/GenerateRandomBooleans/GenerateRandomBooleans/AlgorithmBenchmarks/BenchmarkBase.cs
```diff
@@ -0,0 +1,33 @@
+﻿using GenerateRandomBooleans.BooleanGenerators;
+using GenerateRandomBooleans.RandomGenerators;
+
+namespace GenerateRandomBooleans.AlgorithmBenchmarks
```
Banovvv, line 4: You can use file-scoped namespaces throughout to reduce nesting

PR #1792, opened 2024-02-15
CM-915: Fastest Way to Get the First N Characters of a String in C#

cjaynjoku:


strings-csharp/GetFirstNCharactersOfAString/GetFirstNCharactersOfAString/FirstNCharactersOfStringGetter.cs
```diff
@@ -0,0 +1,54 @@
+﻿namespace GetFirstNCharactersOfAString;
+public class FirstNCharactersOfStringGetter(string inputString = "CodeMaze", int numberOfCharacters = 4)
+{
+    private readonly string _inputString = inputString;
+    private readonly int _numberOfCharacters = numberOfCharacters;
+
+    private bool IsInputStringNullOrEmptyOrShorter()
+        => string.IsNullOrEmpty(_inputString) || _inputString.Length < _numberOfCharacters;
```
jshergal, line 8: Based on nullables, `_inputString` is never null and so this check can be shortened to simply the length check.

strings-csharp/GetFirstNCharactersOfAString/GetFirstNCharactersOfAString/FirstNCharactersOfStringGetter.cs
```diff
@@ -0,0 +1,54 @@
...
+
+    public string? UseRemove()
```
jshergal, line 10: This applies throughout the class, since we are guaranteed that `_inputString` is not null, then the return value of these methods is also guaranteed to never be null. No need then for `string?` as the return type.

strings-csharp/GetFirstNCharactersOfAString/GetFirstNCharactersOfAString/FirstNCharactersOfStringGetter.cs
```diff
@@ -0,0 +1,70 @@
+﻿using System.Text;
+
+namespace GetFirstNCharactersOfAString;
+public class FirstNCharactersOfStringGetter(string inputString = "CodeMaze", int numberOfCharacters = 4)
```
jshergal, line 4: A blank line is needed here

strings-csharp/GetFirstNCharactersOfAString/GetFirstNCharactersOfAString/FirstNCharactersOfStringGetter.cs
```diff
@@ -0,0 +1,70 @@
...
+
+    private bool IsInputStringShorter()
+        => _inputString.Length < _numberOfCharacters;
+
+    public ReadOnlySpan<char> UseForLoop()
+    {
+        if (IsInputStringShorter())
+            return _inputString;
+
+        var strBuilder = new StringBuilder();
```
jshergal, line 17: This won't make a huge difference because this method will still be one of the slowest, but we should at least use the `StringBuilder` overload that takes an initial capacity argument. We already know how long our final string will be, so this will reduce reallocations in the StringBuilder

strings-csharp/GetFirstNCharactersOfAString/GetFirstNCharactersOfAString/FirstNCharactersOfStringGetter.cs
```diff
@@ -0,0 +1,70 @@
+﻿using System.Text;
+
+namespace GetFirstNCharactersOfAString;
+public class FirstNCharactersOfStringGetter(string inputString = "CodeMaze", int numberOfCharacters = 4)
+{
+    private readonly string _inputString = inputString;
+    private readonly int _numberOfCharacters = numberOfCharacters;
```
jshergal, line 7: Since we are using a primary constructor, we don't need these member variables, we can just use the parameters from the constructor

strings-csharp/GetFirstNCharactersOfAString/GetFirstNCharactersOfAString/FirstNCharactersOfStringGetter.cs
```diff
@@ -0,0 +1,63 @@
...
+    public ReadOnlySpan<char> UseRemove()
+    {
+        if (IsInputStringShorter())
+            return inputString;
+
+        return inputString.Remove(numberOfCharacters);
+    }
+
+    public ReadOnlySpan<char> UseLINQ()
+        => string.Join("", inputString.Take(numberOfCharacters));
```
elfico, line 33: @cjaynjoku 
This method takes the N character and converts it to `string` before returning as `ReadOnlySpan<>`.
I think we can get better performance by just returning the `IEnumerable` since we are focusing on returning number of characters.
I made the change and ran the benchmark and got:

Screenshot_1
The new method is the `UseLINQEnumerable` method. Though it uses more memory than the next 2 methods
@jshergal what do you think?

elfico (changes requested):

strings-csharp/GetFirstNCharactersOfAString/GetFirstNCharactersOfAString/FirstNCharactersOfStringGetter.cs
jshergal: @elfico - I agree with you here. Since the article is focused on getting the first N characters, we don't have to return them as strings, or spans. It could be an array, it could be an IEnumerable, etc. The key is that we return "N" characters.

Note, that regarding an IEnumerable, in the benchmark we need to be sure to actually consume it:
https://mawosoft.github.io/BenchmarkDotNet/api/BenchmarkDotNet.Engines.ConsumerExtensions.html

elfico: @cjaynjoku 
Like Jeff mentioned please make a note to use `Consume`. When I added the `Consume` construct, the `UseLINQEnumerable` performed worse.
Here is a link to the documentation: https://benchmarkdotnet.org/articles/samples/IntroDeferredExecution.html

cjaynjoku: I have already done this @elfico. Please, check the latest commit. Thank you.

elfico (approved):

PR #1800, opened 2024-02-17
CM-1015: How to Add Unique Constraints to a Property in EF Core Code First

Banovvv:


dotnet-efcore/UniqueConstraintsInEFCore/UniqueConstraintsInEFCore.Tests/SolarSystemApiApplicationFactory.cs
```diff
@@ -0,0 +1,22 @@
...
+            services.RemoveAll(typeof(DbContextOptions<SolarSystemDbContext>));
+            services.AddDbContext<SolarSystemDbContext>(options =>
+                options.UseSqlServer(_msSqlContainer.GetConnectionString()));
+        });
+    }
+
+    public Task InitializeAsync()
+         => _msSqlContainer.StartAsync();
+
+    public new Task DisposeAsync()
```
fl0tschi, line 20: The `DisposeAsync()` method is virtual at least from the documentation perspective. And we should also consider calling the `base.DisposeAsync()` then, since we don't know what is happening under the hood.

fl0tschi: Ok apparently I didn't see the `IAsyncLifetime `usage, but do we need that when we can override the one from `WebApplicationFactory`?

fl0tschi:
```
public override async ValueTask DisposeAsync() {
    await _msSqlContainer.DisposeAsync();

    await base.DisposeAsync();
}
```

Banovvv: This wouldn't work, but I implemented the `DisposeAsync()` from `IAsyncLifetime` explicitly.

PR #1805, opened 2024-02-19
CM-989 Fastest Way to Find and Get a Number From a String in C#

nitinkale82:
Different ways to find and extract number from string 
-Find and extract a number from a string Using RegEx 
-Find and extract a number from a string Using LINQ and Char.IsDigit 
-Find and extract a number from a string Using StringBuilder and Char.IsDigit

### DIFF
--- a/strings-csharp/CheckNumberInString/CheckNumberInString/ExtractNumber.cs
+++ b/strings-csharp/CheckNumberInString/CheckNumberInString/ExtractNumber.cs
@@ -1,49 +1,107 @@
-﻿using System.Text;
+﻿using System.Globalization;
+using System.Text;
 using System.Text.RegularExpressions;
 
 namespace CheckNumberInString
 {
-    public static class ExtractNumber
+    public static partial class ExtractNumber
     {
-        public static double ExtractNumberUsingRegEx(string inputString)
+        // Define valid numerical characters including digits, minus sign, and decimal point
+        private static readonly char[] validNumericalValues = { '0', '1', '2', '3', '4', '5', '6', '7', '8', '9', '-', '.' };
+
+        [GeneratedRegex(@"-?\d+(\.\d+)?", RegexOptions.Compiled)]
+        private static partial Regex NumberRegex();
+        public static string ExtractNumberUsingRegEx(string inputString)
         {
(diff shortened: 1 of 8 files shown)


strings-csharp/CheckNumberInString/CheckNumberInStringTests/GlobalUsings.cs
```diff
@@ -0,0 +1,9 @@
+global using Microsoft.VisualStudio.TestTools.UnitTesting;
+global using CheckNumberInString;
+using BenchmarkDotNet.Running;
```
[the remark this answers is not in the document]
nitinkale82: Removed all the unwanted directives

strings-csharp/CheckNumberInString/CheckNumberInString/ExtractNumber.cs
```diff
@@ -0,0 +1,107 @@
...
+
+        [GeneratedRegex(@"-?\d+(\.\d+)?", RegexOptions.Compiled)]
+        private static partial Regex NumberRegex();
+        public static string ExtractNumberUsingRegEx(string inputString)
+        {
+            var extractedNumbers = new List<double>();
+
+            var matches = NumberRegex().Matches(inputString);
+
+            foreach (Match match in matches)
```
grantwinney, line 20: One more here that can change to `var`, unless you think it's less clear that way.

strings-csharp/CheckNumberInString/CheckNumberInStringTests/GlobalUsings.cs
```diff
@@ -0,0 +1,3 @@
+global using CheckNumberInString;
+
```
grantwinney, line 2: These blank lines can be removed. Probably wouldn't have bothered mentioning it, but since there's the other `var` keyword to change anyway...

strings-csharp/CheckNumberInString/CheckNumberInString/ExtractNumber.cs
nitinkale82: This does not need to change since match.value is used in the loop and if used var code will through error.

PR #1816, opened 2024-02-20
Action and Func delegates code maze evaluation article.

toc-dev:
Example code with unit tests (using xUnit).


toc-dev (commented):
Updates look good

PR #1826, opened 2024-02-22
[CM-418] How to Convert ReadOnlyMemory to Byte Array in C#

koddek:


csharp-basic-topics/ConvertingReadOnlyMemoryToByteArrayInCSharp/ConvertingReadOnlyMemoryToByteArrayInCSharp/Program.cs
```diff
@@ -0,0 +1,57 @@
...
+
+    public static byte[] HashPassword(string password)
+    {
+        ReadOnlyMemory<byte> passwordMemory = Encoding.UTF8.GetBytes(password);
+        using (var hashAlgorithm = SHA256.Create())
+        {
+            byte[] byteArray = passwordMemory.ToArray();
+            return hashAlgorithm.ComputeHash(byteArray);
+        }
+    }
```
jshergal, line 56: The direction of these examples is good, but some users may comment that since Encoding.UTF8.GetBytes() is already returning a byte[], we are in essence just copying one array to another array in order to pass the copied array into the methods.
So the idea and technique/thought here is good. Maybe for the file writing example we could make this a generic method that takes a ReadOnlyMemory<T> (which also means that we can make this method async since we can't use ReadOnlySpan as an argument to an async). We could then use the Marshal class to marshall the memory to bytes (as we did previously) and then call the WriteAsync method to write the bytes. We can use the unmanaged constraint to ensure that T is bittable.

csharp-basic-topics/ConvertingReadOnlyMemoryToByteArrayInCSharp/ConvertingReadOnlyMemoryToByteArrayInCSharpUnitTests/ProgramTests.cs
```diff
@@ -0,0 +1,61 @@
...
+        // Assert
+        Assert.Equal(expected, Convert.ToBase64String(result));
+    }
+
+    [Fact]
+    public void GivenSavedToFile_WhenValidPathAndContent_ThenFileExists()
+    {
+        // Arrange
+        var documentsPath = Environment.GetFolderPath(Environment.SpecialFolder.Personal);
+        var filePath = Path.Combine(documentsPath, "hello.txt");
```
jshergal, line 49: This should be using a temp file (Path.GetTempFileName), not something in the user folder.

csharp-basic-topics/ConvertingReadOnlyMemoryToByteArrayInCSharp/ConvertingReadOnlyMemoryToByteArrayInCSharpUnitTests/ProgramTests.cs
```diff
@@ -0,0 +1,61 @@
...
+        var content = "Hello, World!";
+
+        // Act
+        Program.SaveFile(filePath, content);
+
+        // Assert
+        Assert.True(File.Exists(filePath));
+
+        // Cleanup
+        File.Delete(filePath);
```
jshergal, line 59: Nice job on remembering to cleanup! 😀

csharp-basic-topics/HowToConvertReadOnlyMemoryToByteArrayInCSharp/HowToConvertReadOnlyMemoryToByteArrayInCSharp/Program.cs
```diff
@@ -0,0 +1,39 @@
...
+
+    public static byte[] ReadOnlyMemoryToByteArray<T>(ReadOnlyMemory<T> input)
+        where T : struct
+    {
+        return MemoryMarshal.AsBytes(input.Span).ToArray();
+    }
+
+    public static byte[] SaveText(string path, ReadOnlyMemory<char> text)
+    {
+        using var hashAlgorithm = SHA256.Create();
```
jshergal, line 31: No need to create an instance of the hasher, we can just use the static SHA256.HashData, HashDataAsync

PR #1830, opened 2024-02-23
CM-979: Fastest Way to Check if a List is Ordered in C#

vnikolaew:
This PR includes the source code with examples for the article "Fastest Way to Check if a List is Ordered in C#".


collections-lists/FastestWayToCheckIfListIsOrdered/FastestWayToCheckIfListIsOrdered/Benchmarks.cs
```diff
@@ -0,0 +1,102 @@
...
+    {
+        var isOrdered = ListOrderValidator.IsOrderedUsingSpans(List);
+    }
+    
+    [Benchmark]
+    public void IsOrderedUsingEnumerator()
+    {
+        var isOrdered = ListOrderValidator.IsOrderedUsingEnumerator(List);
+    }
+
```
jshergal, line 60: These could all be changed from `void` to `bool` and then just be expression body methods

collections-lists/FastestWayToCheckIfListIsOrdered/FastestWayToCheckIfListIsOrdered/Benchmarks.cs
```diff
@@ -0,0 +1,102 @@
...
+        _pool = ArrayPool<int>.Shared;
+        Array = _pool.Rent(Length);
+        
+        for (var i = 0; i < Length; i++)
+        {
+            Array[i] = i;
+        }
+        
+        List = Array.ToList();
+    }
```
jshergal, line 32: There isn't really a need for the ArrayPool here (although kudos for thinking about using it). In a benchmark the GlobalSetup will only be run once, so it does not affect the benchmarking results. In this context, clarity is to be favored over performance. We should just use a line like:
`List = Enumerable.Range(0, Length).ToList()`
Then we can also drop the GlobalCleanup.

collections-lists/FastestWayToCheckIfListIsOrdered/FastestWayToCheckIfListIsOrdered/ListOrderValidator.cs
```diff
@@ -0,0 +1,202 @@
...
+		list.CopyTo(array, 0);
+		Array.Sort(array, 0, length, comparer);
+
+		for (var i = 0; i < length; i++)
+		{
+			if (comparer.Compare(list[i], array[i]) != 0) return false;
+		}
+
+		return true;
+	}
```
jshergal, line 38: The array creation should take place within this method. I see in the Program.cs that we are using a pooled array, which is exactly what we should be doing, but that is an implementation detail that should be hidden from the caller. The way the method is written now, we should be validating the length of the provided array, etc.
This would also have the added benefit of making the method signature the same as the other methods.

collections-lists/FastestWayToCheckIfListIsOrdered/FastestWayToCheckIfListIsOrdered/ListOrderValidator.cs
```diff
@@ -0,0 +1,202 @@
...
+
+	public static bool IsOrderedUsingSpans<T>(List<T> list, IComparer<T>? comparer = default)
+	{
+		comparer ??= Comparer<T>.Default;
+		var span = CollectionsMarshal.AsSpan(list);
+
+		var length = span.Length;
```
jshergal, line 45: For consistency, we should also have the "bail early" check of `list.Count <= 1` here.

collections-lists/FastestWayToCheckIfListIsOrdered/FastestWayToCheckIfListIsOrdered/ListOrderValidator.cs
```diff
@@ -0,0 +1,202 @@
...
+
+		using var enumerator = orderedList.GetEnumerator();
+		enumerator.MoveNext();
+
+		var length = list.Count;
+		for (var i = 0; i < length; i++)
+		{
+			if (comparer.Compare(list[i], enumerator.Current) != 0) return false;
+			enumerator.MoveNext();
+		}
```
jshergal, line 100: Maybe the enumerator is faster than a foreach here, not sure, but if they are basically the same performance, we should favor the foreach since the code will be shorter and easy to follow.

CodeMazeBlog:
retest this please

PR #1851, opened 2024-02-27
CM-1000: How to Convert IAsyncEnumerable to List

Kaybangz:


collections-lists/ConvertIAsyncEnumerableToList/ConvertIAsyncEnumerableToListLibrary/AsyncEnumerableExtensions.cs
```diff
@@ -0,0 +1,18 @@
+﻿namespace ConvertIAsyncEnumerableToListLibrary;
+public static class AsyncEnumerableExtensions
+{
+    public static async Task<List<T>> ToListAsync<T>
+    (this IAsyncEnumerable<T> source, CancellationToken cancellationToken = default)
+    {
+        var userList = new List<T>();
+
+        await foreach (var item in source
+                .ConfigureAwait(false)
```
jshergal, line 10: the ConfigureAwait should be the last call in the chain

collections-lists/ConvertIAsyncEnumerableToList/ConvertIAsyncEnumerableToListLibrary/AsyncEnumerableExtensions.cs
```diff
@@ -0,0 +1,18 @@
+﻿namespace ConvertIAsyncEnumerableToListLibrary;
+public static class AsyncEnumerableExtensions
+{
+    public static async Task<List<T>> ToListAsync<T>
```
jshergal, line 4: This can be a value task

collections-lists/ConvertIAsyncEnumerableToList/ConvertIAsyncEnumerableToListLibrary/AsyncUserListConverter.cs
```diff
@@ -0,0 +1,24 @@
...
+namespace ConvertIAsyncEnumerableToListLibrary;
+public static class AsyncUserListConverter
+{
+    public static async Task<List<User>> GetUsersAsListAsync()
+    {
+        var usersAsync = UserPageFetcher.GetUsersAsync();
+
+        var usersList = await usersAsync.ToListAsync();
+
+        return usersList;
```
jshergal, line 12: Just return directly

collections-lists/ConvertIAsyncEnumerableToList/ConvertIAsyncEnumerableToListLibrary/AsyncUserListConverter.cs
```diff
@@ -0,0 +1,24 @@
...
+    }
+
+    public static async Task<List<User>> GetUsersAsListAsyncWithCancellationToken
+    (CancellationToken cancellationToken = default)
+    {
+        var usersAsync = UserPageFetcher.GetUsersAsyncWithCancellationToken(cancellationToken);
+
+        var usersList = await usersAsync.ToListAsync();
+
+        return usersList;
```
jshergal, line 22: Just return directly

collections-lists/ConvertIAsyncEnumerableToList/ConvertIAsyncEnumerableToListLibrary/Model/User.cs
```diff
@@ -0,0 +1,6 @@
+﻿namespace ConvertIAsyncEnumerableToListLibrary.Model;
+public class User
+{
+    public int Id { get; set; }
+    public string Name { get; set; } = null!;
+}
```
jshergal, line 6: This could be a single line record

collections-lists/ConvertIAsyncEnumerableToList/ConvertIAsyncEnumerableToListLibrary/UserPageFetcher.cs
```diff
@@ -0,0 +1,49 @@
...
+    {
+        for (int page = 1; page <= 3; page++)
+        {
+            await Task.Delay(500);
+
+            var users = new List<User>
+            {
+                new User { Id = 1, Name = "Alice" },
+                new User { Id = 2, Name = "Bob" },
+                new User { Id = 3, Name = "John" },
```
jshergal, line 17: Maybe we should adjust these Id values to be related to the page value from the for loop, so that we don't have the same Id each time, even if we have repeated names.

collections-lists/ConvertIAsyncEnumerableToList/ConvertIAsyncEnumerableToListTests/IAsyncEnumerableToListTests.cs
```diff
@@ -0,0 +1,78 @@
+namespace ConvertIAsyncEnumerableToListTests;
+
+public class IAsyncEnumerableToListTests
+{
+    private static readonly CancellationToken CancellationToken = new CancellationToken();
+    private static readonly CancellationTokenSource CancellationTokenSource = new CancellationTokenSource();
```
jshergal, line 6: We should move these to be local to the tests that will use them. That way there is no issue of calling cancel in one test and it being respected in another test where we don't actually want a cancellation. It also makes the tests cleaner and more straightforward.

collections-lists/ConvertIAsyncEnumerableToList/ConvertIAsyncEnumerableToListLibrary/UserPageFetcher.cs
```diff
@@ -0,0 +1,49 @@
+﻿using ConvertIAsyncEnumerableToListLibrary.Model;
+using System.Runtime.CompilerServices;
+
+namespace ConvertIAsyncEnumerableToListLibrary;
+public static class UserPageFetcher
+{
+    public static async IAsyncEnumerable<User> GetUsersAsync(int delayMilliseconds = 10)
```
grantwinney, line 7: (1 of 3) I think you can leave this as 500, in both methods.
```suggestion
    public static async IAsyncEnumerable<User> GetUsersAsync(int delayMilliseconds = 500)
```

collections-lists/ConvertIAsyncEnumerableToList/ConvertIAsyncEnumerableToListLibrary/AsyncUserListConverter.cs
```diff
@@ -0,0 +1,21 @@
+﻿using ConvertIAsyncEnumerableToListLibrary.Model;
+
+namespace ConvertIAsyncEnumerableToListLibrary;
+public static class AsyncUserListConverter
+{
+    public static async Task<List<User>> GetUsersAsListAsync()
```
grantwinney, line 6: (2 of 3) And then change these methods to accept a delay too.
```suggestion
    public static async Task<List<User>> GetUsersAsListAsync(int delayMilliseconds = 500)
```

collections-lists/ConvertIAsyncEnumerableToList/ConvertIAsyncEnumerableToListTests/IAsyncEnumerableToListTests.cs
```diff
@@ -0,0 +1,46 @@
+namespace ConvertIAsyncEnumerableToListTests;
+
+public class IAsyncEnumerableToListTests
+{
+    [Fact]
+    public async Task GivenGetUsersAsListAsync_WhenCalled_ThenReturnsListOfUsers()
+    {
+        // Act
+        var usersList = await AsyncUserListConverter
+            .GetUsersAsListAsync();
```
grantwinney, line 10: (3 of 3) And then pass in the much lower value from the tests, so they run as fast as possible.

```suggestion
            .GetUsersAsListAsync(10);
```

grantwinney (commented):
Just for the sake of clarity, regarding what I typed in the jira thread, here's what I meant.

collections-lists/ConvertIAsyncEnumerableToList/ConvertIAsyncEnumerableToListLibrary/UserListConverterAndChecker.cs
```diff
@@ -0,0 +1,13 @@
+﻿
```
grantwinney, line 1: This blank line should be removed.

collections-lists/ConvertIAsyncEnumerableToList/ConvertIAsyncEnumerableToListExe/Program.cs
```diff
@@ -0,0 +1,4 @@
+﻿// See https://aka.ms/new-console-template for more information
```
grantwinney, line 1: This default comment can be removed.

collections-lists/ConvertIAsyncEnumerableToList/ConvertIAsyncEnumerableToListLibrary/UserPageFetcher.cs
```diff
@@ -0,0 +1,49 @@
...
+            var users = new List<User>
+            {
+                new User((page - 1) * 3 + 1, "Alice"),
+                new User((page - 1) * 3 + 2, "Bob"),
+                new User((page - 1) * 3 + 3, "John")
+            };
+
+            foreach (var user in users)
+            {
+                yield return user;
```
grantwinney, line 45: In the article, underneath where this snippet is located, you have this text:

> In this modified version, we pass the `CancellationToken` to `Task.Delay` to allow the delay to be cancelable. Additionally, we call `cancellationToken.ThrowIfCancellationRequested()` before yielding each user, as well as before the delay to ensure that the operation can be canceled at any point.

It looks like you meant to call `cancellationToken.ThrowIfCancellationRequested()` before yielding, but forgot to add that part to the code.

Kaybangz:
I added this line of code `cancellationToken.ThrowIfCancellationRequested();` twice in the method. While it might appear redundant, I think it's useful to check the cancellation token at different points to respond promptly to cancellation requests.

PR #1855, opened 2024-03-02
CM-1017 using EF migrations in prod

GergoVandor:


dotnet-efcore/EfCoreCodeFirstMigrationsInProd/Tests/EfCoreCodeFirstMigrationsInProdTests/StartupFactoryTests.cs
```diff
@@ -0,0 +1,50 @@
...
+        
+        var startup = factory.GetStartup(args);
+        
+        Assert.That(startup, Is.InstanceOf<WebApiStartup>());
+    }
+    
+    [Test]
+    public void GivenStartupFactory_WhenNoMigrationArgIsPresent_ThenReturnWebApiStartup()
+    {
+        var args = new[] {"arg1", "--arg2", "migrate", "--migration"};
```
grantwinney, line 32: Would you mind just renaming the last array item here to something else? Since the test has "NoMigration" in the name, but the last item in the array is `--migration`, it might be a bit confusing what the test is doing at first glance.

Could just remove the last item too, and rename the test `GivenStartupFactory_WhenMultipleArgsArePresentWithNoMigrationArg_ThenReturnWebApiStartup`, to contrast it with the test below this one.

GergoVandor: Last item removed, and test renamed

PR #1856, opened 2024-03-02
CM-1019: Why Do We Use the Virtual Keyword for Class Properties in EF Core

DLA-Webagency:


dotnet-efcore/VirtualKeywordInEFCore/VirtualKeywordInEFCore/Models/Author_lazy.cs
```diff
@@ -0,0 +1,12 @@
+﻿using System.ComponentModel.DataAnnotations;
+
+namespace VirtualKeywordInEFCore.Models
+{
+    public class Author_lazy
```
grantwinney, line 5: This should be renamed using pascal case instead of an underscore. Same goes for the `Book_lazy` class, variables like `authors_lazy`, elements in the test cases like `colleenHOOVER_author`, etc.

```suggestion
    public class AuthorLazy
```

dotnet-efcore/VirtualKeywordInEFCore/Tests/VirtualKeywordInEFCoreTests.cs
```diff
@@ -0,0 +1,61 @@
+using Microsoft.EntityFrameworkCore;
+using VirtualKeywordInEFCore.Models;
+
+namespace Tests
+{
+    public class VirtualKeywordInEFCoreTests
+    {
+        private readonly DataContextWithoutLazyLoading _context;
```
grantwinney, line 8: These can be instantiated where they're defined, and then taken out of the constructor.

```suggestion
        private readonly DataContextWithoutLazyLoading _context = new();
```

dotnet-efcore/VirtualKeywordInEFCore/VirtualKeywordInEFCore/Models/DataContextWihoutLazyLoading.cs
grantwinney, line 1: I missed this before, but this file name has a typo in it, although the class itself looks okay.

PR #1867, opened 2024-03-06
CM-995: How to Retrieve the Number of CPU Cores With C#

lennart02:


dotnet-platform-specific/RetrievalOfCpuCores/RetrievalOfCpuCores/CpuInformation.cs
```diff
@@ -0,0 +1,121 @@
...
+        return Environment.ProcessorCount;
+    }
+
+    /// <summary>
+    /// Retrieves the number of logical processors of the CPU.
+    /// </summary>
+    /// <returns>number of logical processors</returns>
+    public int GetLogicalProcessors()
+    {
+        foreach (var item in new System.Management.ManagementObjectSearcher("Select * from Win32_ComputerSystem").Get())
```
elfico, line 22: @lennart02 
These strings should be constants instead of magic strings. It helps us to avoid errors and reuse the code. There are other occurrences of this

dotnet-platform-specific/RetrievalOfCpuCores/RetrievalOfCpuCores/CpuInformation.cs
```diff
@@ -0,0 +1,121 @@
...
+        finally
+        {
+            if (deviceList != IntPtr.Zero) { SetupDiDestroyDeviceInfoList(deviceList); }
+        }
+        return deviceCount;
+    }
+
+    [DllImport("setupapi.dll", SetLastError = true)]
+    private static extern IntPtr SetupDiGetClassDevs(ref Guid ClassGuid,
+        [MarshalAs(UnmanagedType.LPStr)] String enumerator,
```
elfico, line 92: @lennart02 
Let's use `string` instead of `String` and `int` instead of `Int32`

dotnet-platform-specific/RetrievalOfCpuCores/Tests/RetrievalOfCpuCoresTests.cs
```diff
@@ -0,0 +1,69 @@
...
+        public void WhenGetLogicalProcessors_LogicalProcessorEqualToEnvironmentProcessorCount()
+        {
+            if (!OperatingSystem.IsWindows())
+            {
+                return;
+            }
+            var cpuInformation = new CpuInformation();
+            var logicalProcessors = cpuInformation.GetLogicalProcessors();
+            var processorsCount = cpuInformation.GetProcessorsCount();
+            Assert.AreEqual(logicalProcessors, processorsCount);
```
elfico, line 16: @lennart02 
the `Assert` statements are to be separated by a new line

dotnet-platform-specific/RetrievalOfCpuCores/Tests/RetrievalOfCpuCoresTests.cs
```diff
@@ -0,0 +1,69 @@
...
+        }
+
+        [TestMethod]
+        public void WhenGetLogicalProcessors_ResultIsGreaterThanZero()
+        {
+            if (!OperatingSystem.IsWindows())
```
elfico, line 22: @lennart02 
would it be better if we add this logic on the `SUT` instead of in the tests

elfico (changes requested):

elfico:
@lennart02 
The build is also failing for this PR

dotnet-platform-specific/RetrievalOfCpuCores/Tests/RetrievalOfCpuCoresTests.cs
lennart02: I added an Exception in the SUT. However, this check is still necessary in the tests, as we cannot perform it on non-Windows computers.

dotnet-platform-specific/RetrievalOfCpuCores/RetrievalOfCpuCores/CpuInformation.cs
elfico: @lennart02 
We should declare these strings as const and reuse them. From the code I can see 2 places where we used `Select * from Win32_ComputerSystem` 
If we declare it at the top of the class e.g. `const string Win32 = Select * from Win32_ComputerSystem`, we can then reuse it. This applies to the other strings too including `NumberOfLogicalProcessors` e.t.c

dotnet-platform-specific/RetrievalOfCpuCores/RetrievalOfCpuCores/CpuInformation.cs
```diff
@@ -0,0 +1,144 @@
...
+                    deviceCount = deviceNumber;
+                    break;
+                }
+            }
+        }
+        finally
+        {
+            if (deviceList != IntPtr.Zero) { SetupDiDestroyDeviceInfoList(deviceList); }
+        }
+        return deviceCount;
```
elfico, line 110: @lennart02 
There should be a newline before `return` and `assert` statements

elfico (changes requested):

dotnet-platform-specific/RetrievalOfCpuCores/RetrievalOfCpuCores/CpuInformation.cs
```diff
@@ -0,0 +1,148 @@
...
+    /// </summary>
+    /// <returns>number of logical processors</returns>
+    public int GetLogicalProcessors()
+    {
+        if (!OperatingSystem.IsWindows())
+        {
+            throw new PlatformNotSupportedException();
+        }
+
+        var wmiQuery = "Select * from Win32_ComputerSystem";
```
elfico, line 27: @lennart02 
Can we make this a `const` string and use it in other places in the code. 
This string `"Select * from Win32_ComputerSystem";` is also in the other method.

dotnet-platform-specific/RetrievalOfCpuCores/RetrievalOfCpuCores/CpuInformation.cs
```diff
@@ -0,0 +1,148 @@
...
+    /// <returns>number of cores of the CPU</returns>
+    public int GetNumberOfCores()
+    {
+        if (!OperatingSystem.IsWindows())
+        {
+            throw new PlatformNotSupportedException();
+        }
+
+        var coreCount = 0;
+        var wmiQuery = "Select * from Win32_Processor";
```
elfico, line 49: @lennart02 
We should make this a const string too.

dotnet-platform-specific/RetrievalOfCpuCores/RetrievalOfCpuCores/CpuInformation.cs
elfico, line 27: @lennart02 
Can we make this a `const` string and use it in other places in the code. 
This string `"Select * from Win32_ComputerSystem";` is also in the other method.

dotnet-platform-specific/RetrievalOfCpuCores/RetrievalOfCpuCores/CpuInformation.cs
elfico, line 49: @lennart02 
We should make this a const string too.

dotnet-platform-specific/RetrievalOfCpuCores/RetrievalOfCpuCores/CpuInformation.cs
```diff
@@ -0,0 +1,148 @@
...
+    /// </summary>
+    /// <returns>number of physical processors</returns>
+    public int GetPhysicalProcessors()
+    {
+        if (!OperatingSystem.IsWindows())
+        {
+            throw new PlatformNotSupportedException();
+        }
+
+        var wmiQuery = "Select * from Win32_ComputerSystem";
```
elfico, line 70: @lennart02 
here too.

elfico (changes requested):

elfico (approved):

PR #1877, opened 2024-03-09
CM-835: How to Benchmark a Method's Performance Across Different .NET…

cjaynjoku:
… Versions

### DIFF
--- a/dotnet-performance/BenchmarkAcrossDifferentDotNETVersions/BenchmarkAcrossDifferentDotNETVersions/GetBooksBenchmark.cs
+++ b/dotnet-performance/BenchmarkAcrossDifferentDotNETVersions/BenchmarkAcrossDifferentDotNETVersions/GetBooksBenchmark.cs
@@ -1,0 +1,23 @@
+﻿using BenchmarkDotNet.Attributes;
+using BenchmarkDotNet.Columns;
+using BenchmarkDotNet.Configs;
+using BenchmarkDotNet.Jobs;
+using BenchmarkDotNet.Reports;
+
+namespace BenchmarkAcrossDifferentDotNETVersions;
+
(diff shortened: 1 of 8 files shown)


dotnet-performance/BenchmarkAcrossDifferentDotNETVersions/BenchmarkAcrossDifferentDotNETVersions/GetBooksBenchmark.cs
```diff
@@ -0,0 +1,23 @@
...
+
+namespace BenchmarkAcrossDifferentDotNETVersions;
+
+[SimpleJob(RuntimeMoniker.Net60, baseline: true)]
+[SimpleJob(RuntimeMoniker.Net70)]
+[SimpleJob(RuntimeMoniker.Net80)]
+[Config(typeof(StyleConfig))]
+public class GetBooksBenchmark
+{
+    private class StyleConfig : ManualConfig
```
elfico, line 15: @cjaynjoku 
We should move the `StyleConfig` to a different file. Except there is a reason why we want to have it here in which case, we should add an explanation for it in the article.

elfico (changes requested):

dotnet-performance/BenchmarkAcrossDifferentDotNETVersions/BenchmarkAcrossDifferentDotNETVersions/GetBooksBenchmark.cs
cjaynjoku: @elfico, since we have just one benchmark class (`GetBooksBenchmark`) using the configuration defined in the `StyleConfig` class, we should leave it as a private class. I'll add this explanation to the article.

PR #1912, opened 2024-03-18
CM-874: How to Update an Item in a List in C#

KaiWesterschwiensterdt:


collections-csharp/UpdateItemInListCSharp/UpdateItemInListCSharp.sln
sahebagarwal-codemaze, line 1: Remove the “src” folder and move the main project folder and test project folder to the same level as the solution file.

collections-csharp/UpdateItemInListCSharp/src/UpdateItemInListCSharp/Models/Book.cs
```diff
@@ -0,0 +1,13 @@
...
+{
+    public class Book(string author, string isbn, bool isCheckedOut, string title)
+    {
+        public string Author { get; set; } = author;
+
+        public string ISBN { get; set; } = isbn;
+
+        public bool IsCheckedOut { get; set; } = isCheckedOut;
+
+        public string Title { get; set; } = title;
```
sahebagarwal-codemaze, line 11: Better to have Title as the first property of book, I feel it will add more readability

collections-csharp/UpdateItemInListCSharp/src/UpdateItemInListCSharp/Models/Library.cs
```diff
@@ -0,0 +1,81 @@
...
+                new("J.K. Rowling", "978-0439064873", false, "Harry Potter and the Chamber of Secrets"),
+                new("J.K. Rowling", "978-0439139601", false, "Harry Potter and the Goblet of Fire"),
+                new("J.K. Rowling", "978-0439358071", false, "Harry Potter and the Order of the Phoenix"),
+                new("J.K. Rowling", "978-0439785969", false, "Harry Potter and the Half-Blood Prince")
+            ];
+        }
+
+        public Book AddBook(string author, string isbn, bool isCheckedOut, string title)
+        {
+            var book = new Book(author, isbn, isCheckedOut, title); Books.Add(book);
```
sahebagarwal-codemaze, line 21: Books.Add(book); must be in the next line

collections-csharp/UpdateItemInListCSharp/src/UpdateItemInListCSharp/Models/Library.cs
```diff
@@ -0,0 +1,81 @@
...
+            return book;
+        }
+
+        public void CheckoutBookFind(string isbn)
```
sahebagarwal-codemaze, line 25: Please rename the method from "CheckoutBookFind" to "CheckoutBookUsingFind". Use the work "Using" before method name, applicable for all the methods in this class

collections-csharp/UpdateItemInListCSharp/src/UpdateItemInListCSharp/Models/Library.cs
```diff
@@ -0,0 +1,81 @@
...
+            var index = Books.FindIndex(b => b.ISBN == isbn);
+            if (index != -1)
+            {
+                Books[index].IsCheckedOut = true;
+            }
+        }
+
+        public void CheckoutBookFirstOrDefault(string isbn)
+        {
+            var bookToCheckoutFirstOrDefault = Books.FirstOrDefault(b => b.ISBN == isbn);
```
sahebagarwal-codemaze, line 45: variable name can be "bookToCheckout" instead of "bookToCheckoutFirstOrDefault"

collections-csharp/UpdateItemInListCSharp/src/UpdateItemInListCSharp/Models/Library.cs
```diff
@@ -0,0 +1,81 @@
...
+            if (bookToCheckoutFirstOrDefault != null)
+            {
+                bookToCheckoutFirstOrDefault.IsCheckedOut = true;
+            }
+        }
+
+        public void CheckoutBookForeach(string isbn)
+        {
+            foreach (var bookItem in Books)
```
sahebagarwal-codemaze, line 54: We can rename "bookItem" to "book"

collections-csharp/UpdateItemInListCSharp/src/UpdateItemInListCSharp/UpdateItemInListCSharp.csproj
```diff
@@ -0,0 +1,16 @@
+﻿<Project Sdk="Microsoft.NET.Sdk">
+
+  <PropertyGroup>
+    <OutputType>Exe</OutputType>
+    <TargetFramework>net8.0</TargetFramework>
+    <ImplicitUsings>enable</ImplicitUsings>
+    <Nullable>enable</Nullable>
+    <InvariantGlobalization>true</InvariantGlobalization>
+	<Optimize>true</Optimize>
```
sahebagarwal-codemaze, line 9: What is the reason of having <Optimize>true</Optimize> in this project file?

collections-csharp/UpdateItemInListCSharp/src/UpdateItemInListCSharp/Models/Library.cs
```diff
@@ -0,0 +1,81 @@
...
+            {
+                if (bookItem.ISBN == isbn)
+                {
+                    bookItem.IsCheckedOut = true;
+                }
+            }
+        }
+
+        public void CheckoutBookIndexOf(Book book)
```
sahebagarwal-codemaze, line 63: This is not the correct way to get the index of the matching book object. Object is by reference, since you are adding the same object in the list and pass the same book object in the list, it is working for you.

collections-csharp/UpdateItemInListCSharp/src/UpdateItemInListCSharp.Test/LibraryTest.cs
```diff
@@ -0,0 +1,87 @@
...
+
+namespace UpdateItemInListCSharp.Test
+{
+    public class LibraryTest
+    {
+        [Fact]
+        public void WhenCheckoutBookFind_ThenBookIsCheckedOut()
+        {
+            // Arrange
+            var library = new Library();
```
sahebagarwal-codemaze, line 11: Library can be created as a private readonly property of a class. 
In each method we can create one new Book object with IsCheckedOut property set to false and add that object in the library collection. Then we can checkout that book using different methods and finally verify the IsCheckedOut property of the new book object

collections-csharp/UpdateItemInListCSharp/src/UpdateItemInListCSharp/UpdateItemInListCSharp.csproj
KaiWesterschwiensterdt: The InvariantGlobalization and Optimize Tags both increase the performance of the C# compiler.
InvariantGlobalization increasing performance due to a "universal language".
Optimize enables the optimization feature of the C# compiler, which forces the compiler to optimize the code for maximum speed and minimum size. This also improves the runtime performance.

collections-csharp/UpdateItemInListCSharp/src/UpdateItemInListCSharp/Models/Book.cs
KaiWesterschwiensterdt: Should I only adjust the parameters in the constructor or restructure the order of the properties?
Personally, I always have the properties sorted alphabetically so that I work in a standardised way and don't always have to think about it.

collections-csharp/UpdateItemInListCSharp/src/UpdateItemInListCSharp.Test/LibraryTest.cs
KaiWesterschwiensterdt: I had always taken the first book from the list because I had not implemented a GetBookByIsbn() method. To identify the IsCheckedOut property of the books, should I implement this method in the Library class in each Assert section of the tests? Then we can always create new books and compare them at the end of each test.

sahebagarwal-codemaze: What I was suggesting is as follows:
1. Create new Book object in each test case with IsCheckedOut=false.
2. Then add that book object in the list
3. Next checkout the same book which we created in step 1
4. Finally in Assert verify IsCheckedOut=true or not.

collections-csharp/UpdateItemInListCSharp/src/UpdateItemInListCSharp/Models/Book.cs
sahebagarwal-codemaze: I prefer to have most important to least important properties in the class. I am expecting both, to re-arrange properties and order of the parameters.

collections-csharp/UpdateItemInListCSharp/src/UpdateItemInListCSharp/UpdateItemInListCSharp.csproj
sahebagarwal-codemaze: Because of this settings, when I was running the application from VS in Debug mode, it was not allowing me to run, I am getting a warning prompt. I reviewed other projects in our repo, we are not using this settings, I feel we can remove these settings.

collections-csharp/UpdateItemInListCSharp/src/UpdateItemInListCSharp/Models/Library.cs
KaiWesterschwiensterdt: I don't quite understand this point. I add a book and then check out the same book using the CheckoutBookUsingIndexOf method. In my test method for this method, I get a new object from this book, which is also checked out. So I have achieved exactly what the method was supposed to do.

KaiWesterschwiensterdt: I update the book class and the CheckoutUsingIndexOf() method.
These changes allow me to find a book in the list based on its properties rather than the specific object instance (reference).

collections-csharp/UpdateItemInListCSharp/UpdateItemInListCSharp.Test/LibraryTest.cs
```diff
@@ -0,0 +1,100 @@
...
+        [Fact]
+        public void WhenCheckoutBookFind_ThenBookIsCheckedOut()
+        {
+            // Arrange
+            var book = new Book("Dolor sit amet Sadipscing", "Lorem Ipsum", "123-0123456789", false);
+            _library.Books.Add(book);
+
+            // Act
+            _library.CheckoutBookUsingFind("123-0123456789");
+            var checkoutBook = _library.Books.Find(b => b.ISBN == "123-0123456789");
```
sahebagarwal-codemaze, line 18: We don't need to find an object again, the same book object will be checkout. We can directly check if book.IsCheckOut is true or not

collections-csharp/UpdateItemInListCSharp/UpdateItemInListCSharp/Library.cs
```diff
@@ -0,0 +1,86 @@
...
+                if (book.ISBN == isbn)
+                {
+                    book.IsCheckedOut = true;
+                }
+            }
+        }
+
+        public void CheckoutBookUsingIndexOf(Book book)
+        {
+            var newBook = new Book(book.Title, book.Author, book.ISBN, book.IsCheckedOut);
```
sahebagarwal-codemaze, line 68: We should not created a newBook object here, he we should just use the book parameter passed in the method. The caller should create a new book object(different one then it used to Add in the list) for searching.

collections-csharp/UpdateItemInListCSharp/UpdateItemInListCSharp/Models/Book.cs
```diff
@@ -0,0 +1,30 @@
...
+            {
+                return this.Title == b.Title && this.Author == b.Author && this.ISBN == b.ISBN;
+            }
+            else
+            {
+                return false;
+            }
+        }
+
+        public override int GetHashCode()
```
sahebagarwal-codemaze, line 25: What is the use of GetHashCode function for us?

collections-csharp/UpdateItemInListCSharp/UpdateItemInListCSharp/Program.cs
```diff
@@ -0,0 +1,23 @@
+﻿using BenchmarkDotNet.Running;
+using UpdateItemInListCSharp.Models;
```
sahebagarwal-codemaze, line 2: unused using:
using UpdateItemInListCSharp.Models;

collections-csharp/UpdateItemInListCSharp/UpdateItemInListCSharp/Library.cs
KaiWesterschwiensterdt: Would this be enough? 
library.CheckoutBookUsingIndexOf(library.Books[0]);

collections-csharp/UpdateItemInListCSharp/UpdateItemInListCSharp.Test/LibraryTest.cs
KaiWesterschwiensterdt: Okay, that's smarter. Should I remove all "var checkoutBook = _library.Books.Find(b => b.ISBN == "123-0123456789");" lines in every unit test?

sahebagarwal-codemaze: Yes, this comment is applicable for all unit test

collections-csharp/UpdateItemInListCSharp/UpdateItemInListCSharp/Library.cs
sahebagarwal-codemaze: That is again not the correct way, in that way you will again end up using the object reference present in the Books list in library class. You should simply create the new Book object in the caller with the Title, Author and ISBN property set to the value of one of the existing object present in Books list and pass it to CheckoutBookUsingIndexOf() method.

Instead of the below line in CheckoutBookUsingIndexOf() method:
var newBook = new Book(book.Title, book.Author, book.ISBN, book.IsCheckedOut);

I suggested to have this line, but in the caller which will created new object reference, and then passed that object as parameter to CheckoutBookUsingIndexOf() method.

collections-csharp/UpdateItemInListCSharp/UpdateItemInListCSharp/BenchmarkTester.cs
```diff
@@ -0,0 +1,48 @@
...
+        public void CheckoutBookUsingForeach()
+        {
+            _library.CheckoutBookUsingForeach("978-0439708180");
+        }
+
+        [Benchmark]
+        public void CheckoutBookUsingIndexOf()
+        {
+            _library.CheckoutBookUsingIndexOf(_library.Books[0]);
+        }
```
sahebagarwal-codemaze, line 40: Use the same approach to call IndexOf method as we did in Program.cs

var book = library.Books.Last();
        var newBook = new Book(book.Title, book.Author, book.ISBN, book.IsCheckedOut);
        library.CheckoutBookUsingIndexOf(newBook);

collections-csharp/UpdateItemInListCSharp/UpdateItemInListCSharp.Test/LibraryTest.cs
```diff
@@ -0,0 +1,92 @@
...
+
+            // Assert
+            Assert.True(book.IsCheckedOut);
+        }
+
+        [Fact]
+        public void WhenCheckoutBookIndexOf_ThenBookIsCheckedOut()
+        {
+            // Arrange
+            var book = _library.Books[0];
```
sahebagarwal-codemaze, line 69: Use same approach as Program.cs

var book = library.Books.Last();
        var newBook = new Book(book.Title, book.Author, book.ISBN, book.IsCheckedOut);
        library.CheckoutBookUsingIndexOf(newBook);

PR #1919, opened 2024-03-21
CM-1047: Using Result Pattern in .NET Web API

sswietoniowski:
Created example solution presenting the Result Pattern in .NET Web API.

It contains sample contacts API demonstrating different ways of handling errors:

- no error handling at all,
- null checking as a means to handling errors,
- exception used for flow control and error handling,
- the Result Pattern custom implementation,
- FluentResults library.


aspnetcore-webapi/UsingResultPatternInNETWebAPI/UsingResultPatternInNETWebAPI/Repositories/IContactsRepository.cs
sahebagarwal-codemaze, line 1: Keep the name singular, so instead of "IContactsRepository" rename it to "IContactRepository". This comment is applicable through out the application

aspnetcore-webapi/UsingResultPatternInNETWebAPI/UsingResultPatternInNETWebAPI/Repositories/IContactsRepository.cs
```diff
@@ -0,0 +1,9 @@
+﻿namespace UsingResultPatternInNETWebAPI.Repositories;
+
+public interface IContactsRepository
+{
+    IEnumerable<Contact> GetAll();
+    Contact? GetById(Guid id);
+    Contact Create(Contact contact);
+    bool ExistsByEmail(string email);
```
sahebagarwal-codemaze, line 8: Instead of ExistsByEmail method, I would prefer to have GetByEmail method, the way we have GetById

aspnetcore-webapi/UsingResultPatternInNETWebAPI/UsingResultPatternInNETWebAPI/v1/Services/BasicContactsService.cs
sahebagarwal-codemaze, line 1: 1. The way we have Interface and Implementation in Repositories. We can follow same for services.
2. Also for services, We can have one Services folder in the same level as Repositories and Controller's Version folder, inside that we should have Interfaces and Implementation folder.

aspnetcore-webapi/UsingResultPatternInNETWebAPI/UsingResultPatternInNETWebAPI/v3/Exceptions/ApiNotFoundException.cs
sahebagarwal-codemaze, line 1: Instead of "ApiNotFoundException", it should be something like "RecordNotFoundException"

aspnetcore-webapi/UsingResultPatternInNETWebAPI/UsingResultPatternInNETWebAPI/v4/TheResultPattern/CustomError.cs
```diff
@@ -0,0 +1,19 @@
+﻿namespace UsingResultPatternInNETWebAPI.v4.TheResultPattern;
+
+public sealed record CustomError(string Code, string Message)
+{
+    private static readonly string NotFoundCode = "NotFound";
+    private static readonly string ValidationErrorCode = "ValidationError";
+
+    public static readonly CustomError None = new(string.Empty, string.Empty);
+
+    public static CustomError NotFound(Guid id)
```
sahebagarwal-codemaze, line 10: Same comments here, it can accept two parameter, one for propertName and another for value. The error message can use those parameter to be dynamic.

aspnetcore-webapi/UsingResultPatternInNETWebAPI/UsingResultPatternInNETWebAPI/v4/TheResultPattern/CustomError.cs
```diff
@@ -0,0 +1,19 @@
...
+    {
+        return new CustomError(NotFoundCode, $"contact with id {id} was not found");
+    }
+
+    public static CustomError ValidationError(string email)
```
sahebagarwal-codemaze, line 15: Since this is used for Validation related error. Instead of hardcoding message, the complete message must be received as a parameter.

aspnetcore-webapi/UsingResultPatternInNETWebAPI/UsingResultPatternInNETWebAPI/v4/TheResultPattern/CustomResult.cs
```diff
@@ -0,0 +1,56 @@
...
+
+    private CustomResult(T value)
+    {
+        Value = value;
+        IsSuccess = true;
+        Error = CustomError.None;
+    }
+
+    private CustomResult(bool isSuccess, CustomError error)
+    {
```
sahebagarwal-codemaze, line 15: Since it is a private method and called only for isSuccess=false, the isSuccess=true condition check is not required.

aspnetcore-webapi/UsingResultPatternInNETWebAPI/UsingResultPatternInNETWebAPI/v5/Errors/ApiNotFoundError.cs
sahebagarwal-codemaze, line 1: Same comment as above, it should be "RecordNotFoundError"

aspnetcore-webapi/UsingResultPatternInNETWebAPI/UsingResultPatternInNETWebAPI/v5/Controllers/FluentResultsContactsController.cs
```diff
@@ -0,0 +1,61 @@
...
+        _contactsService = contactsService;
+    }
+
+    [HttpGet]
+    [ProducesResponseType(200)]
+    public ActionResult<IEnumerable<ContactDto>> GetAll()
+    {
+        var result = _contactsService.GetAll();
+
+        if (result.IsFailed)
```
sahebagarwal-codemaze, line 22: For GetAll() method I don't see any case when this block of code will be executed.

aspnetcore-webapi/UsingResultPatternInNETWebAPI/Tests/ApiApplicationFactory.cs
sahebagarwal-codemaze, line 1: Overall I have two comments in Test Project.
1. Since we have test cases for Controllers and Services, we can have two folders first inside Tests for Controllers and Services, inside that we can have test case files. Also all the test class name must end only with "Tests" not "IntegrationTest"
2. The way we organized the code in "OtherContactsControllerIntegrationTest" using Theory and Inline is more cleaner and readable then "BasicContactsControllerIntegrationTest". And for consistency let's maintain one way of writing.

aspnetcore-webapi/UsingResultPatternInNETWebAPI/UsingResultPatternInNETWebAPI/Repositories/IContactsRepository.cs
sswietoniowski: > Instead of ExistsByEmail method, I would prefer to have GetByEmail method, the way we have GetById

GetById would suggest that we are returning the contact, while this method is only to verify that this contact exists, are you sure that we should that?

aspnetcore-webapi/UsingResultPatternInNETWebAPI/UsingResultPatternInNETWebAPI/v1/Services/BasicContactsService.cs
sswietoniowski: There is a reason for this organization: the repositories are not changing; I created subdirectories to present what is changing; if we do so, there should be v1, v2, ... subdirectory inside every Services and Controllers folder. This way (at least to me), it would be more difficult to follow what changes. Sure, we can place all services and controllers inside respective folder, but at the same time as is, it is easier to see what was created for every version without the need to look into many subfolders. What do you think?

aspnetcore-webapi/UsingResultPatternInNETWebAPI/UsingResultPatternInNETWebAPI/v3/Exceptions/ApiNotFoundException.cs
sswietoniowski: As for the "message must be specified", sure, we can change it that way, but what exactly can we achieve by this? This error is related to the fact that the given ID was not found. What if I add a public property that would allow me to retrieve the ID if anyone needs it, as for the message, we can pass it too, but truly I don't see any advantage of that.

aspnetcore-webapi/UsingResultPatternInNETWebAPI/UsingResultPatternInNETWebAPI/v4/TheResultPattern/CustomResult.cs
sswietoniowski: Yes, and no, sure, I can't call it from the outside to create a situation when we've got a contradiction (isSuccess == true && some error). Still, as it is a piece of code that someone else might use and reuse, the fact that we are preventing from creating an inconsistent state seems to be helpful. What do you think?

aspnetcore-webapi/UsingResultPatternInNETWebAPI/UsingResultPatternInNETWebAPI/v5/Controllers/FluentResultsContactsController.cs
sswietoniowski: It can be used for testing (there is a test case in Api.http for that), given a user clones this repo and wants to verify that the API is working as expected. I didn't use it in the article as it is pretty long already. Same question - should we remove it?

aspnetcore-webapi/UsingResultPatternInNETWebAPI/Tests/ApiApplicationFactory.cs
sswietoniowski: I can change that but this naming is in accordance with this CodeMaze document: https://docs.google.com/document/d/1PRj3IPPPNTF-NU-XGQqthflVd3m8tkX7c2JGn4gfX74/edit These tests are not unit tests, but integration tests and in this document it is explicitly stated that they should be named:
"
- all tests should follow our standard naming convention: 
FooUnitTest.cs // unit tests
FooIntegrationTest.cs or FooIntTest.cs // integration tests
FooLiveTest.cs // tests that require a running component
"
I'm using [Fact] when there is only one case and [Theory] otherwise, but I'm sure I can use one everywhere.

sswietoniowski: I'm using [Fact] instead of [Theory] for the "BasicContactsControllerIntegrationTest" as it is slightly, but still different than "OtherContactsControllerIntegrationTest" and tests inside the basic controller are parameterless.

aspnetcore-webapi/UsingResultPatternInNETWebAPI/UsingResultPatternInNETWebAPI/Repositories/IContactsRepository.cs
sswietoniowski: As for the "applicable throughout the application", I am not sure whether I understand your intent as we are talking about personal preferences. Do you want me to change "contacts" to "contact" for services and controllers as well?

sahebagarwal-codemaze: You understood it correctly, I am expecting it for all services and controller.
"SomethingContactsService" to be SomethingContactService

aspnetcore-webapi/UsingResultPatternInNETWebAPI/UsingResultPatternInNETWebAPI/Repositories/IContactsRepository.cs
sahebagarwal-codemaze: This is just a sample project and it is not the core part which we are focusing in the article, so it hardly matters. But why I suggested this approach is that, if in the real application we need to fetch the record using EmailId or want to check if the record exists with EmailId, with a single method we can cover both the scenario.

aspnetcore-webapi/UsingResultPatternInNETWebAPI/UsingResultPatternInNETWebAPI/v1/Services/BasicContactsService.cs
sahebagarwal-codemaze: My suggestion is we should have Controllers folder, inside that v1,v2, etc. For Services, we should not have version folder.

aspnetcore-webapi/UsingResultPatternInNETWebAPI/UsingResultPatternInNETWebAPI/v3/Exceptions/ApiNotFoundException.cs
sahebagarwal-codemaze: This error was about the contact(record) was not found with the given ID. This is one scenario. What if somebody wants to fetch the record based on any other field(emailId, etc) and the record was not found? For those scenarios, it make sense to have "RecordNotFoundException" type with messages explicitly passed from caller.

aspnetcore-webapi/UsingResultPatternInNETWebAPI/UsingResultPatternInNETWebAPI/v4/TheResultPattern/CustomResult.cs
sahebagarwal-codemaze: If you want to keep it as is, I don't mind.

aspnetcore-webapi/UsingResultPatternInNETWebAPI/Tests/ApiApplicationFactory.cs
sahebagarwal-codemaze: I believe in maintaining consistency in the code. For reader also, I will be simple if we follow one way unless absolutely required.

aspnetcore-webapi/UsingResultPatternInNETWebAPI/UsingResultPatternInNETWebAPI/v5/Controllers/FluentResultsContactsController.cs
sahebagarwal-codemaze: We are returning Result.Ok() from GetAll() method? In which scenario, result.IsFailed will be true?

sswietoniowski: You are right. It was added artificially to keep it consistent.

aspnetcore-webapi/UsingResultPatternInNETWebAPI/UsingResultPatternInNETWebAPI/TheResultPattern/CustomError.cs
```diff
@@ -0,0 +1,19 @@
...
+
+    public static readonly CustomError None = new(string.Empty, string.Empty);
+
+    public static CustomError RecordNotFound(string message)
+    {
+        return new CustomError(RecordNotFoundCode, message);
+    }
+
+    public static CustomError ValidationError(string email)
+    {
```
sahebagarwal-codemaze, line 16: The ValidationError implementation should be same as RecordNotFound which will accept the message from caller. The current implementation cannot be used for other validation errors.

aspnetcore-webapi/UsingResultPatternInNETWebAPI/Tests/Controllers/ContactControllerIntegrationTest.cs
```diff
@@ -0,0 +1,173 @@
...
+    {
+        // Arrange
+        var baseUri = string.Format(_baseApiTemplate, version);
+        _client.BaseAddress = new Uri(baseUri);
+
+        // Act
+        var response = await _client.GetAsync("");
+
+        // Assert
+        var message = await response
```
sahebagarwal-codemaze, line 32: We don't need to print the message in test case

PR #1921, opened 2024-03-22
CM-1042: Build a URL Shortener in .NET

cjaynjoku:


aspnetcore-webapi/DotNetURLShortener/DotNetURLShortener/Services/UrlShortenerService.cs
```diff
@@ -0,0 +1,55 @@
...
+    {
+        static string GenerateShortCode()
+        {
+            var shortCode = new char[ShortCodeLength];
+            for (var i = 0; i < ShortCodeLength; i++)
+            {
+                shortCode[i] = AlphaNumericChars[Random.Shared.Next(AlphaNumericChars.Length)];
+            };
+
+            return new string(shortCode);
```
jshergal, line 22: You can use [Random.Shared.GetItems()](https://learn.microsoft.com/en-us/dotnet/api/system.random.getitems?view=net-8.0#system-random-getitems-1(system-readonlyspan((-0))-system-span((-0)))) here in combination with [string.Create](https://learn.microsoft.com/en-us/dotnet/api/system.string.create?view=net-8.0#system-string-create-1(system-int32-0-system-buffers-spanaction((system-char-0))))

aspnetcore-webapi/DotNetURLShortener/DotNetURLShortener/Services/UrlShortenerService.cs
```diff
@@ -0,0 +1,55 @@
...
+        }
+
+        foreach (var pair in UrlDict)
+        {
+            if (pair.Value == longUrl)
+            {
+                return pair.Key;
+            }
+        }
```
jshergal, line 31: Why not just use [Dictionary.TryGetValue](https://learn.microsoft.com/en-us/dotnet/api/system.collections.generic.dictionary-2.trygetvalue?view=net-8.0) here? We have a dictionary which as O(1) lookup, so we should take advantage of that.

cjaynjoku: Hi @jshergal. Here, we are trying to retrieve the key(the shortcode) with a known value(a long URL). We use the TryGetValue() method in the second method in this class where we have a shortcode and wish to retrieve its long URL.

jshergal: @cjaynjoku - You are correct. That is my bad, yeah since you are looking for the value the best is a linear search. From a performance perspective, it might be worthwhile to maintain two dictionaries, but I suppose in general, if we are creating a URL shortener, we wouldn't be using dictionaries but an actual database and so we would have an index on the long URL value.

cjaynjoku: That's right @jshergal, using an actual database would be better. However, the instruction in the issue's description was to avoid using databases and focus primarily on shortening the URL, probably to make the article easier to follow. I'll add this database caveat to the article.

aspnetcore-webapi/DotNetURLShortener/DotNetURLShortener/Services/UrlShortenerService.cs
elfico: @cjaynjoku this is still pending

aspnetcore-webapi/DotNetURLShortener/DotNetURLShortener/Controllers/URLShortenerController.cs
```diff
@@ -0,0 +1,41 @@
...
+            return NotFound();
+        }
+
+        return Ok(longUrl);
+    }
+
+    [HttpGet("/samplepage")]
+    public IActionResult GetSamplePage()
+    {
+        return Ok("This is a sample page to test our URL shortener.");
```
elfico, line 28: @cjaynjoku 
This route does nothing. Can we remove it?

elfico (changes requested):

aspnetcore-webapi/DotNetURLShortener/DotNetURLShortener/Controllers/URLShortenerController.cs
elfico: My mistake. Just saw where it was used in the article. @cjaynjoku

elfico (approved):

aspnetcore-webapi/DotNetURLShortener/DotNetURLShortener/Services/UrlShortenerService.cs
cjaynjoku: Hello @elfico, I have changed the implementation of that method to include the methods @jshergal suggested here.

elfico: Thanks. Somehow my updated code wasn't reflecting. I had to pull again.
Sorry for the bother

PR #1931, opened 2024-03-29
CM-72: Getting PropertyMappings with Automapper

MiaoShuYo:


grantwinney:
This entire solution should not be in the "aspnetcore-features" folder in the CodeMaze repo, since it's not specific to ASP.NET Core. Please move it to the "dotnet-client-libraries" folder instead.

aspnetcore-features/GettingPropertyMappingsWithAutomapper/Test/GettingPropertyMappingsWithAutomapperTest.cs
```diff
@@ -0,0 +1,12 @@
+namespace Test
+{
+    [TestClass]
+    public class GettingPropertyMappingsWithAutomapperTest
+    {
+        [TestMethod]
+        public void WhenProgrammeExecutionThenCorrectOutputLiveTest()
+        {
+
```
grantwinney, line 9: There should be at least one test in here. It seems like you could re-use some of the same code you're running in the `Program` class, and verify that the mapping is working correctly.

PR #1941, opened 2024-04-01
CM-930: How to Find All the Positions of a Substring in Another String in C#

lynda-nneka:


strings-csharp/SubstringInString/SubstringInString/SubstringSearchMethods.cs
```diff
@@ -0,0 +1,102 @@
...
+        List<int> indexes = new();
+        var regex = GeneratedRegex();
+
+        foreach (Match match in regex.Matches(input))
+        {
+            indexes.Add(match.Index);
+        }
+
+        return indexes;
+    }
```
jshergal, line 75: Since we are taking the search value as an argument, we can't use a source generated regex. This method signature could be changed to take a Regex though so that in the case of the benchmark especially, we can pass in the same regex for each execution and won't be generating it every time. We could then still use a source generated regex, just the source generation would now take place outside of this class.

strings-csharp/SubstringInString/SubstringInString/SubstringSearchMethods.cs
```diff
@@ -0,0 +1,102 @@
...
+            if (inputSpan.Slice(i, search.Length).SequenceEqual(searchSpan))
+            {
+                indexes.Add(i);
+            }
+        }
+
+        return indexes;
+    }
+
+    public static List<int> FindAllIndexesWithIndexOf(string input, string search)
```
jshergal, line 43: I would expect the performance to be slightly better here if we change the parameter types to `ReadOnlySpan<char>`.

strings-csharp/SubstringInString/SubstringInString/SubstringSearchMethods.cs
```diff
@@ -0,0 +1,102 @@
...
+            indexes.Add(match.Index);
+        }
+
+        return indexes;
+    }
+
+    public static List<int> FindAllIndexesWithLINQ(string input, string search)
+    {
+        var indexes = input
+            .Select((c, i) => i <= input.Length - search.Length && input.Substring(i, search.Length) == search ? i : -1)
```
jshergal, line 80: We can be using a Span and SequenceEqual here just like we did in the FindIndexesWithSpan method

jshergal: For reference, here is the difference from a benchmark of the two options (the way written here, vs. changing it to use a Span and SequenceEqual):
| Method                     | Mean     | Error     | StdDev    | Ratio | RatioSD | Gen0   | Gen1   | Allocated | Alloc Ratio |
|--------------------------- |---------:|----------:|----------:|------:|--------:|-------:|-------:|----------:|------------:|
| FindAllIndexesWithLINQ     | 2.150 μs | 0.0428 μs | 0.1113 μs |  1.00 |    0.00 | 1.6365 | 0.0019 |    3424 B |        1.00 |
| FindAllIndexesWithLINQSpan | 1.199 μs | 0.0240 μs | 0.0598 μs |  0.56 |    0.05 | 0.1373 |      - |     288 B |        0.08 |

As you can see from the benchmark results, not only is there less overall allocation, the span method is nearly 2x as fast.

For reference, this was my sample data:
```
const string input = "Lorem ipsum dolor sit amet, consectetur adipiscing elip. Duis quis nisip eget sem vehipula accumsan";
const string search = "ip";
```

strings-csharp/SubstringInString/SubstringInString/Benchmarking.cs
```diff
@@ -0,0 +1,45 @@
+﻿using BenchmarkDotNet.Attributes;
+
+[MemoryDiagnoser]
+[Orderer(BenchmarkDotNet.Order.SummaryOrderPolicy.FastestToSlowest)]
+public class Benchmarking
+{
+    public string str = new("Lorem".ToCharArray()[0], 5);
```
jshergal, line 7: @lynda-nneka why did you change the string that will be searched? For the benchmark we would want a "normal" size string and preferably one that contains the sought for text multiple times so multiple indexes are returned. You changed the paragraph of text that was here into a 5 character string containing just the letter 'L'. Here is the result if you print the string that is being created here:
`LLLLL`

strings-csharp/SubstringInString/SubstringInString/SubstringSearchMethods.cs
```diff
@@ -0,0 +1,100 @@
...
+        for (int i = 0; i < input.Length - search.Length + 1; i++)
+        {
+            if (regex.IsMatch(input.AsSpan(i, search.Length)))
+            {
+                indexes.Add(i);
+            }
+        }
+
+        return indexes;
+    }
```
jshergal, line 71: My suggestion was that this method take a Regex as an input parameter. We can't do what we are doing here and then benchmark it. We are creating a new Regex on every method call. Also, we are doing something that the other methods are not, which is performing a case insensitive match. The other methods are performing case sensitive match.
We need to change the method signature here to be a caller provided Regex. This way in the benchmark we only create the Regex once, either as a source generated regex or using the Compiled option (both are about equal in performance, one has a higher cost on instantiation, but won't matter in the benchmark).

strings-csharp/SubstringInString/SubstringInString/Benchmarking.cs
lynda-nneka: Screenshot (42)
I got a feedback here for it.

jshergal: @lynda-nneka I am sorry that that advice was given. In this context we want a longer string which contains the searched for text in multiple locations. Lorem Ipsum is also a good choice in this context. For the benchmark we want to be returning at least a couple instances of matches, just so that we see the performance of the collection being built as well. Since all of this is part of the benchmark.

One way you could make that text cleaner would be to use a raw string literal, that way you don't need all the plus signs. Since the value is constant, you can also declare it as constant.

Either way you handle it, whether it is to put it in a separate file, add a library like Bogus and generate the string from the Lorem data set, or some other method of string creation, for the benchmark we would want to have at least a paragraph of text. And we definitely want text that contains the searched for text.

lynda-nneka: @jshergal Should I put it back as a paragraph in my code or read it as a text file

jshergal: @lynda-nneka I would just put it as a paragraph in the code, but as a raw string literal, which will make it a bit cleaner.

strings-csharp/SubstringInString/SubstringInString/Benchmarking.cs
```diff
@@ -1,11 +1,27 @@
...
 {
-    public string str = new("Lorem".ToCharArray()[0], 5);
+    private const string str = @"Lorem ipsum dolor sit amet, consectetur adipiscing elit.
+                               Nullam quis lorem nec lorem imperdiet tempus. Phasellus lorem justo,
+                               consequat ut lorem sit amet, tempor feugiat lorem.
+                               Proin lorem velit, aliquam vel lorem in, consequat lorem nulla.
+                               Sed lorem justo, eleifend eu lorem id, consectetur lorem neque.
+                               Vestibulum lorem mauris, placerat sit amet lorem nec, laoreet lorem nisi.
+                               Sed nec lorem libero. Nullam lorem lorem, ullamcorper quis lorem vitae.";
+
```
jshergal, line 15: I would make this a raw string literal, it will make it cleaner

strings-csharp/SubstringInString/SubstringInString/Benchmarking.cs
```diff
@@ -1,11 +1,27 @@
...
     public string toFind = "Lorem";
+    private Regex regex;
+
+    [GlobalSetup]
+    public void Setup()
+    {
+        // Compile the regex pattern into a Regex object using the Compiled option
+        regex = new Regex(toFind, RegexOptions.Compiled);
+    }
```
jshergal, line 24: 👍

strings-csharp/SubstringInString/SubstringInString/Program.cs
```diff
@@ -1,7 +1,9 @@
 ﻿using BenchmarkDotNet.Running;
+using System.Text.RegularExpressions;
 
 var str = "Lorem ipsum dolor sit amet, consectetur adipiscing elip. Duis quis nisip eget sem vehipula accumsan";
 var toFind = "ip";
+var regex = new Regex(toFind, RegexOptions.Compiled);
```
jshergal, line 6: We don't necessarily need to make it a compiled Regex here, just for the benchmark to have the highest performance we can give it. This also applies in the tests. There is a startup cost to compiling the Regex.

strings-csharp/SubstringInString/Tests/RegexTest.cs
```diff
@@ -7,10 +9,11 @@ public void Given_InputStringAndSearchString_When_FindAllIndexesWithRegexCalled_
         // Given
         var input = "Lorem ipsum dolor sit amet, consectetur adipiscing elip. Duis quis nisip eget sem vehipula accumsan.";
         var search = "ip";
+        var regex = new Regex(search, RegexOptions.Compiled);
```
jshergal, line 12: See the previous comment, the Compiled option isn't needed in the test.

strings-csharp/SubstringInString/SubstringInString/Benchmarking.cs
jshergal: Just to clarify, what you have now is a verbatim string, raw string literals begin and end with """.

strings-csharp/SubstringInString/SubstringInString/Benchmarking.cs
OTybolc: @jshergal Why can't we use something like this in a phase of global setup, or iteration setup:
private string str = "";
private const string ToFind = "Lorem";

[GlobalSetup]
public void Setup()
{
    StringBuilder sb = new StringBuilder();

    sb.Append(ToFind);
    sb.Append(new string('a', 1000));
    sb.Append(ToFind);
    sb.Append(new string('a', 1000));
    sb.Append(ToFind);

    str = sb.ToString();
}

IMO: it's better than putting a string literal that contains 10 lines. In such a way we can construct any string with any length without the necessity of changing the string literal.

jshergal: @OTybolc. 
If the string is nothing but a continuous group of the same character, let's use 'a' as you have listed here, and the string is 1000 in length, all 'a'. What indexes would you expect to be returned if we search for "aaa"? The result would be 2,3,4,...,996. There will be `1000 - searchLength` occurrences.

To exercise, in a benchmark, this code, the better example is a block of "real world" text and a search value that occurs multiple times throughout the text. Thus the _lorem ipsum_ text. It looks "real-worldy" and contains the word "lorem" multiple times. Whether we put it in a file or have it live in the code, doesn't matter so much. I would argue that having the paragraph in the code looks okay and is easily accessible to someone running the code, and easy to change.

So the main issue with the repeated character is now the string is everywhere and there is basically no index where it is not found. Whereas the _lorim ipsum_ text provides a more realistic example case, thus validating the benchmarking conditions.

strings-csharp/SubstringInString/SubstringInString/Benchmarking.cs
```diff
@@ -1,26 +1,30 @@
...
     public string toFind = "Lorem";
     private Regex regex;
+    string pattern = @"ip";
 
     [GlobalSetup]
     public void Setup()
     {
         // Compile the regex pattern into a Regex object using the Compiled option
-        regex = new Regex(toFind, RegexOptions.Compiled);
+        regex = new Regex(pattern, RegexOptions.Compiled);
```
jshergal, line 27: @lynda-nneka the Regex search should match the other searches. Since this is a benchmark where we are making a comparison amongst different methods for achieving the same goal, they each need to return the same results.

strings-csharp/SubstringInString/SubstringInString/SubstringSearchMethods.cs
```diff
@@ -36,24 +33,24 @@ public static List<int> FindAllIndexesWithSpan(string input, string search)
...
             {
-                indexes.Add(startIndex);
-                startIndex++;
+                indexes.Add(i);
             }
         }
-        while (startIndex != -1);
 
         return indexes;
     }
```
jshergal, line 52: @lynda-nneka this method has been changed to be identical to the FindAllIndexesWithSpan, it should be changed back to use the IndexOf method like it was using previously.

lynda-nneka: I got a feedback to change it, it's quite confusing taking all these feedbacks as I do not know which exact correction to follow
Screenshot (45)

OTybolc: @jshergal why do we need to use plain string here? If we're using the ReadOnlySpan, it should increase performance. Since we won't be copying string when calling the method. Or I'm missing something?

jshergal: @OTybolc and @lynda-nneka,
The issue was not the change to ReadOnlySpan, the code in the method itself was changed and now the two methods are identical and in the IndexOf method, we are not using IndexOf.

@lynda-nneka, the feedback to change to ReadOnlySpan was good feedback, and a suggestion I had made as well, what was confusing is why you also changed the internals of the method to no longer use IndexOf.

image

OTybolc: @jshergal, 
I didn't have a chance to review the changes yet - just saw that the signature changed. That's why I was a bit confused. Now I see the issue. Thank you.

strings-csharp/SubstringInString/SubstringInString/Benchmarking.cs
lynda-nneka: Okay thanks

strings-csharp/SubstringInString/SubstringInString/SubstringSearchMethods.cs
lynda-nneka: @jshergal  The IndexOf method requires a string comparison type when comparing strings and this overload isnt available for ReadOnlySpan<char>. Except I convert it to a string. What do you think?

jshergal: Hi @lynda-nneka, that is incorrect. There `IndexOf` is available for ReadOnlySpans as well as Spans.
https://learn.microsoft.com/en-us/dotnet/api/system.memoryextensions.indexof?view=net-8.0

lynda-nneka: ok thanks

PR #1955, opened 2024-04-13
CM-1110: Generate Images in C# Using ImageSharp

Emmanuel-Adom:


jshergal:
Hi @Emmanuel-Adom,
Make sure you make separate clean branches for each of these commits. This one includes the SkiaSharp code as well. Also, and this is just a suggestion, but I would encourage you when making your local branches, to name them based on the Jira. I found it makes it easier for me locally to switch between them and to keep things ordered. So for instance, I would probably name this branch "CM1110-ImageSharp"

PR #1976, opened 2024-04-22
CM-1087: How to Generate Random Color Names in C#

elliezub:


csharp-basic-topics/GenerateRandomColorName/GenerateRandomColorName/KnownColorGenerator.cs
```diff
@@ -0,0 +1,20 @@
...
+    private Random randomGen;
+
+    public KnownColorGenerator()
+    {
+        randomGen = new Random();
+    }
+
+    public Color GetRandomKnownColor()
+    {
+        KnownColor[] allColors = (KnownColor[])Enum.GetValues(typeof(KnownColor));
```
jshergal, line 15: Here we can use the generic overload [Enum.GetValues<TEnum>()](https://learn.microsoft.com/en-us/dotnet/api/system.enum.getvalues?view=net-8.0#system-enum-getvalues-1):
`Enum.GetValues<KnownColor>();`
Also, since this is something that is constant, we can make it a static readonly member of the class and just create the array once.
We also don't need to have `Random` as a member since we are not seeding it. We can just use `Random.Shared`. This means that the whole class now can be static and just have a single static `GetRandomKnownColor()` method.

csharp-basic-topics/GenerateRandomColorName/Tests/KnownColorTests.cs
```diff
@@ -0,0 +1,52 @@
...
+            if (!colorCounts.ContainsKey(knownColor))
+                colorCounts[knownColor] = 0;
+
+            colorCounts[knownColor]++;
+        }
+
+        bool hasDuplicates = colorCounts.Any(kvp => kvp.Value > 1);
+
+        Assert.True(hasDuplicates, "Expect some colors to be generated more than once to confirm randomness.");
+    }
```
jshergal, line 51: This isn't a good test as there are only 175 colors defined, generating 200 by definition will guarantee that at least one color has been generated more than once, even if we were not performing this randomly, just sequentially returning the known colors.
I think this test could just be removed. We don't need to test the random number generator since that is built into the framework.

elliezub: I see what you're saying, that definitely makes sense, I will remove it then.

PR #1979, opened 2024-04-25
CM-1043: Using Delegating Handlers to Extend HttpClient in ASP.NET Core

georgy93:
Using Delegating Handlers to Extend HttpClient in ASP.NET Core


fl0tschi:
retest this please

PR #1980, opened 2024-04-25
CM-1077: Introduction to the Semaphore Class in C#

gpanagopoulos:
…aphoreSlim and compare them with lock and mutex


threads-csharp/IntroductionToSemaphoreInCsharp/IntroductionToSemaphoreInCsharp/ExampleWithSemaphore.cs
```diff
@@ -0,0 +1,36 @@
+﻿using IntroductionToSemaphoreInCsharp.Models;
+using System.Globalization;
+
+namespace IntroductionToSemaphoreInCsharp;
+
+public class ExampleWithSemaphore
+{
+    private static List<string> _sharedResource = new List<string>();
+    private static readonly Semaphore _semaphore = new Semaphore(initialCount: 3, maximumCount: 3);
```
CodeMazeBlog, line 9: Please utilize the new() and new(initialCount: 3, maximumCount: 3)
We can add readonly modifier to the list.

CodeMazeBlog: Please check the code for other occurences of this as well.

threads-csharp/IntroductionToSemaphoreInCsharp/Tests/UnitTests.cs
```diff
@@ -0,0 +1,41 @@
...
+    public async Task GivenSomeConcurrentCode_WhenAccessWithSemaphoreSlimAsyncIsInvoked_ThenAllThreadsShouldExecute()
+    {
+        await ExampleWithSemaphoreSlim.AccessWithSemaphoreSlimAsync(SleepDelay);
+        AssertAllThreadsExecuted();
+    }
+
+    private void AssertAllThreadsExecuted()
+    {
+        var result = ConsoleOutput.ToString();
+        for (int i = 0;i < 10;i++)
```
CodeMazeBlog, line 36: Number 10 looks arbitrary here. I understand that's the number of threads run in the code, so make this number available as a constant in the examples classes and then reference it here instead of hard coding number 10.

CodeMazeBlog: Also the for loop is not properly formatted, so run a bit of formatting through the file (and others if necessary).

threads-csharp/IntroductionToSemaphoreInCsharp/Tests/UnitTests.cs
```diff
@@ -0,0 +1,41 @@
...
+{
+    private StringBuilder ConsoleOutput { get; } = new StringBuilder();
+    const int SleepDelay = 50;
+
+    [TestInitialize]
+    public void Init()
+    {
+        Console.SetOut(new StringWriter(this.ConsoleOutput));
+        this.ConsoleOutput.Clear();
+    }
```
CodeMazeBlog, line 17: I don't know if you've done this before, but we shouldn't rely on Console.SetOut to do the tests. This is an easy way to do any test out there in a console application, but we should avoid it because it's unpredictable and it's not really a test. If we're testing threads or concurrency we should really test that, and not the console output. Do some research on this and try to rewrite these to not include Console.SetOut please.

threads-csharp/IntroductionToSemaphoreInCsharp/IntroductionToSemaphoreInCsharp/ExampleWithSemaphore.cs
gpanagopoulos: Done. changed as below:
```
private static readonly List<string> _sharedResource = [];
private static readonly Semaphore _semaphore = new(initialCount: 3, maximumCount: 3);
```
```
private static readonly List<string> _sharedResource = [];
private static readonly SemaphoreSlim _semaphoreSlim = new(3, 3);
```

threads-csharp/IntroductionToSemaphoreInCsharp/Tests/UnitTests.cs
gpanagopoulos: Done, created a new class to include constants and declared a new const variable to hold the number of threads

threads-csharp/IntroductionToSemaphoreInCsharp/Tests/UnitTests.cs
gpanagopoulos: I 've not done this again. I used it as a convenient trade-off wanting to satisfy 2 requirements at the same time and without adding complexity in code:
1. Have a Console output to show in the article what will the code print when we run program.cs
2. Run all code and check that all threads have completed in the Unit Test.

I refactored to how I would do it in normal software:
1. Introduced a ConcurrentQueue to hold all strings generated by the threads.
2. When code completes, I loop/check the output of the ConcurrentQueue.

Can you please review this one and let me know what you think/do you have any improvements to suggest?

One we are happy with the code, I will then update the Article code snippets and code description accordingly.

PR #1981, opened 2024-04-25
CM-1053: How to Compare Two Files in C#

bazen-teklehaymanot:


files-csharp/ComparingTwoFiles/ComparingTwoFiles/FileComparer.cs
```diff
@@ -0,0 +1,56 @@
...
+        return Enumerable.SequenceEqual(bytes1, bytes2);
+    }
+
+    public static bool CompareByChecksum(string file1, string file2)
+    {
+        var hash1 = MD5.HashData(File.ReadAllBytes(file1));
+        var hash2 = MD5.HashData(File.ReadAllBytes(file2));
+
+        return Enumerable.SequenceEqual(hash1, hash2);
+    }
```
jshergal, line 41: Rather then reading the entire contents of the file into an array and then processing, both of these methods could employ FileStream.

In the checksum method, we can just pass the stream directly into the MD5 hash method. Since we are looking at performance in this code (benchmarking), we could stackalloc the resultant array, thus we would remove all heap allocation in the checksum method.

In the CompareByBytes method, again, we could use FileStreams, checking length first, then allocate two buffers for reading, and then read the files in chunks, comparing each chunk. This way if the files differ early on, we bail before reading the entire file. This also has the advantage of minimizing allocation. For instance if the files being compared where 100 MB in length, instead of reading 200 MB of data into arrays and then comparing the arrays, we could open the streams and read 1K chunks or something like that into Span buffers, and then compare those chunks using SequenceEquals.

files-csharp/ComparingTwoFiles/ComparingTwoFiles/FileComparer.cs
```diff
@@ -21,23 +21,54 @@ public static bool CompareByNameAndSize(string file1, string file2)
...
+        {
+            var bytesRead1 = stream1.Read(buffer1);
+            var bytesRead2 = stream2.Read(buffer2);
+
+            if (bytesRead1 == bytesRead2 && bytesRead1 == 0)
+            {
+                return true;
+            }
+
+            if (!MemoryExtensions.SequenceEqual(buffer1, buffer2))
```
jshergal, line 46: Good switch to the stream and spans. Just one cosmetic change here and in the next method. Since `SequenceEqual()` is an extension method, this can be simplified to:
`buffer1.SequenceEqual(buffer2)`

files-csharp/ComparingTwoFiles/ComparingTwoFiles/FileComparer.cs
```diff
@@ -21,23 +21,54 @@ public static bool CompareByNameAndSize(string file1, string file2)
...
+        }
+
+        Span<byte> hash1 = stackalloc byte[16];
+        Span<byte> hash2 = stackalloc byte[16];
+
+        MD5.HashData(stream1, hash1);
+        MD5.HashData(stream2, hash2);
 
-        return Enumerable.SequenceEqual(hash1, hash2);
+        return MemoryExtensions.SequenceEqual(hash1, hash2);
```
jshergal, line 71: Same simplification here:
`hash1.SequenceEqual(hash2)`

fl0tschi:
retest

fl0tschi:
retest this please

PR #1983, opened 2024-04-27
CM-1085: How to Compare Two Dictionaries in C#

cjaynjoku:


collections-dictionary/CompareTwoDictionaries/CompareTwoDictionaries/DictionaryEqualityComparer.cs
```diff
@@ -0,0 +1,47 @@
...
+    public static bool AreKeysAndCountEqual<TKey, TValue>(Dictionary<TKey, TValue> dict1, Dictionary<TKey, TValue> dict2)
+        where TKey : notnull where TValue : IEquatable<TValue>
+    {
+        if (dict1.Count != dict2.Count) return false;
+
+        var dict1HashSet = new HashSet<TKey>(dict1.Keys);
+        var dict2HashSet = new HashSet<TKey>(dict2.Keys);
+
+        return dict1HashSet.SetEquals(dict2HashSet);
+    }
```
jshergal, line 14: Hi @cjaynjoku and @fl0tschi,
If we want to keep a method for checking the count, then we could make this a one-liner expression statement:
`IsCountEqual<TKey, TValue>(Dictionary<TKey, TValue> dict1, Dictionary<TKey, TValue> dict2) => dict1.Count == dict2.Count;`

The other part is pretty heavy, building up a couple of HashSets, then performing a `SetEquals` on them. While that could potentially be a good way to check if the keys are equal, we don't want to be calling it in each of the other methods, otherwise we are massively duplicating work. I will explain:

UseSequenceEqual method: We are comparing each of the `KeyValuePair<Tk, Tv>` objects in order. This means every key will be compared to every key in the other collection, but based on our `AreKeysAndCountEqual` we have already done that, so for this method we perform that comparison twice, but with the added cost of building two HashSets.

UseEnumerableAll method:
Here we are performing a `ContainsKey()` check between the two dictionaries for each element, then and only then do we compare the value. So again, we are comparing each key a second time.

UseForeachLoop method:
Once again we are iterating all of the `KeyValuePair<Tk, Tv>` objects and checking for the key being contained in both dictionaries, followed by a compare of the values. So we are performing a second check to verify the keys exist in both dictionaries.

I don't think we should change the comparison methods, I think the better approach is to change the `AreKeysAndCountEqual` method to `IsCountEqual` or do just drop it altogether and just check the counts directly in the compare methods (for me this would be preferable as it makes it absolutely clear that we are checking the count and it potentially saves a function call)

collections-dictionary/CompareTwoDictionaries/CompareTwoDictionaries/DictionaryEqualityComparerBenchmark.cs
```diff
@@ -0,0 +1,35 @@
...
+            {1, "Rosary Ogechi"},
+            {2, "Clare Chiamaka"},
+        };
+
+    private static readonly Dictionary<int, string> _dict2
+        = new()
+        {
+            {1, "Rosary Ogechi"},
+            {2, "Clare Chiamaka"},
+        };
```
jshergal, line 22: This is very small for a benchmark. We should probably be using a dictionary with around 1000 elements, especially since we are benchmarking an OrderBy LINQ method, which does a bunch of stuff behind the scenes to create a map of index values to iterate the collection in the specified order, which means that it is allocating arrays internally, which of course will add pressure to the GC, so to get a more "realistic" benchmark, we should be looking at a minimum of 1000 items in the dictionary. We can easily generate those in code though:
`Enumerable.Range(1, 1000).ToDictionary(k => k, v => v.ToString())`

This has the added advantage of making the string comparison in the value equality check a little more intensive. Since the above strings are compile time constants, they are reference equal and the string comparison method first checks for reference equality, so each of the values above pass the equality check very quickly.

collections-dictionary/CompareTwoDictionaries/CompareTwoDictionariesBenchmark/Program.cs
```diff
@@ -0,0 +1,4 @@
+﻿using BenchmarkDotNet.Running;
+using CompareTwoDictionariesBenchmark;
+
+BenchmarkRunner.Run<CompareDictionariesBenchmark>();
```
jshergal, line 4: I think this is dead code since we are running the benchmark in the main console application.

PR #2007, opened 2024-05-27
CM-1129: Weak Events in C#

lasaetarubia:


csharp-advanced-topics/WeakEventsInCSharp/Tests/Tests.cs
```diff
@@ -0,0 +1,48 @@
...
+    public void GivenWeakEventWhenGCCollectedThenEventNotRaised()
+    {
+        var weakEventPublisher = new WeakReferencePublisher();
+        var weakEventSubscriber = new WeakReferenceSubscriber();
+        weakEventSubscriber.Subscribe(weakEventPublisher);
+        weakEventPublisher.RaiseEvent();
+
+        Assert.True(weakEventPublisher.Event.HandlerInvoked);
+
+        weakEventSubscriber = null;
```
grantwinney, line 41: When I comment out this line, and the `subscriber = null` line in the other test, both tests pass anyway.

It seems like calling `GC.Collect()` on the weak reference is enough for it to dispose of it, even if it hasn't been set to `null`. Do you think that's expected behavior?

lasaetarubia: The explicit null set is more to highlight the intent that the subscriber is being removed. Without a strong reference present to the weakEventSubscriber, GC will be able to collect it without the explicit null set.

I think this is expected.

PR #2008, opened 2024-06-01
CM-1132: How to Find All the Positions of a Substring in Another String in C#

MPrtenjak:
Initial code sample for the article: How to Find All the Positions of a Substring in Another String in C#


csharp-algorithms/FindAllPositionsOfAString/FindAllPositionsOfAString/Algorithms/FirstAttempt.cs
```diff
@@ -0,0 +1,18 @@
+﻿namespace FindAllPositionsOfAString.Algorithms;
+
+public class FirstAttempt
```
jshergal, line 3: Never instantiated, can be made static

csharp-algorithms/FindAllPositionsOfAString/FindAllPositionsOfAString/Algorithms/FirstAttempt.cs
```diff
@@ -0,0 +1,18 @@
...
+{
+    public static List<int> FindAll(string text, string searchText)
+    {
+        List<int> positions = [];
+
+        var index = text.IndexOf(searchText);
```
jshergal, line 9: Just a note here, but this is performing a search using CurrentCulture for matching. If that is expected it is fine, but we may want to still explicitly list that for clarity sake. This also applies in the while loop.

csharp-algorithms/FindAllPositionsOfAString/FindAllPositionsOfAString/Algorithms/SearchUsingBruteForceAlgorithm.cs
```diff
@@ -0,0 +1,46 @@
...
+            }
+        }
+
+        return positions;
+    }
+
+    protected bool AreEqualCharactersSensitive(char a, char b) 
+        => a == b;
+
+    protected bool AreEqualCharactersInsensitive(char a, char b)
```
jshergal, line 44: These can be private and the class could be sealed

csharp-algorithms/FindAllPositionsOfAString/FindAllPositionsOfAString/Algorithms/SearchUsingIndexOf.cs
```diff
@@ -0,0 +1,25 @@
...
+
+        var index = text.IndexOf(searchText, StringComparison);
+        while (index != -1)
+        {
+            positions.Add(index);
+            index = text.IndexOf(searchText, index + skipSize, StringComparison);
+        }
+
+        return positions;
+    }
```
jshergal, line 21: This will be more performant if we take spans on the arguments and operate over those instead of over the strings. The compiler/runtime can perform additional optimizations in the IndexOf methods when the arguments are spans

csharp-algorithms/FindAllPositionsOfAString/FindAllPositionsOfAString/Samples/SearchingSamples.cs
```diff
@@ -0,0 +1,32 @@
+﻿namespace FindAllPositionsOfAString.Samples;
+
+public record SearchPair(string Text, string SearchText);
+
+public static class SearchingSamples
+{
+    private static string JuliusCaesar = JuliusCaesarText.Read();
+
+    private readonly static SearchPair[] searchPairs =
```
jshergal, line 9: Incorrect modifier declaration order - should be `private static readonly`

csharp-algorithms/FindAllPositionsOfAString/FindAllPositionsOfAString/Algorithms/SearchUsingBruteForceAlgorithm.cs
```diff
@@ -0,0 +1,46 @@
...
+        }
+
+        return positions;
+    }
+
+    protected bool AreEqualCharactersSensitive(char a, char b) 
+        => a == b;
+
+    protected bool AreEqualCharactersInsensitive(char a, char b) 
+        => char.ToLowerInvariant(a) == char.ToLowerInvariant(b);
```
jshergal, line 45: Move to the base class since this is being repeated both here and in the KMP code.

csharp-algorithms/FindAllPositionsOfAString/FindAllPositionsOfAString/Algorithms/FirstAttempt.cs
MPrtenjak: Resolved

csharp-algorithms/FindAllPositionsOfAString/FindAllPositionsOfAString/Algorithms/FirstAttempt.cs
MPrtenjak: Stated in the article

csharp-algorithms/FindAllPositionsOfAString/FindAllPositionsOfAString/Algorithms/SearchUsingIndexOf.cs
MPrtenjak: IndexOf() method operating on ReadOnlySpan<char> does not have  an option to search from N-position

+ If I change the arguments in the interface from string to ReadOnlySpan<char> (which is ugly, for me), then I would have to again convert from ReadOnlySpan<char> to string in RegEx algorithms.

csharp-algorithms/FindAllPositionsOfAString/FindAllPositionsOfAString/Samples/SearchingSamples.cs
MPrtenjak: Corrected

csharp-algorithms/FindAllPositionsOfAString/FindAllPositionsOfAString/Algorithms/SearchUsingBruteForceAlgorithm.cs
MPrtenjak: Corrected

csharp-algorithms/FindAllPositionsOfAString/FindAllPositionsOfAString/Algorithms/SearchUsingIndexOf.cs
jshergal: 1. No need to change the method arguments, just take a span within the method.
2.  You don't need an N-position in the index of, it is a span, so you just slice it:

```
public List<int> FindAll(string text, string searchText)
{
    var spanText = text.AsSpan();
    var spanSearch = searchText.AsSpan();
    List<int> positions = [];

    var skipSize = ((SkipWholeFoundText) ? spanSearch.Length : 1);

    var offset = 0;
    var index = spanText.IndexOf(spanSearch, StringComparison);
    while (index != -1)
    {
        positions.Add(index+offset);
        offset += index + skipSize;
        spanText = spanText[(index + skipSize)..];
        index = spanText.IndexOf(spanSearch, StringComparison);
    }

    return positions;
}
```
The above code can probably be improved and made more readable around the offset computation, etc, but it should give a push in the right direction.

Benchmark using the short string from the pull request:
| Method               | Mean     | Error    | StdDev   | Gen0   | Gen1   | Allocated |
|--------------------- |---------:|---------:|---------:|-------:|-------:|----------:|
| BenchmarkIndexOf     | 68.20 ns | 0.812 ns | 0.720 ns | 0.0459 | 0.0001 |      96 B |
| BenchmarkIndexOfSpan | 62.65 ns | 0.379 ns | 0.296 ns | 0.0459 | 0.0001 |      96 B |

asif-anwar:
retest this please

MPrtenjak:
All tests have passed 👍

PR #2012, opened 2024-06-11
Changes for CM-1128 - Generate Images in an ASP.NET Core App Using OpenAI

muhammedsalimp:
Changes for CM-1128 - Generate Images in an ASP.NET Core App Using OpenAI


dotnet-client-libraries/OpenAIImageGeneration/OpenAiImageGenerator/Services/IOpenAIService.cs
```diff
@@ -0,0 +1,7 @@
+﻿namespace OpenAiImageGenerator.Services
```
grantwinney, line 1: Can you use a file-scoped namespace here, and anywhere else that isn't using it, similar to how you have it in `ImageGeneratorController.cs` class?

dotnet-client-libraries/OpenAIImageGeneration/OpenAiImageGenerator/Services/OpenAIService.cs
```diff
@@ -0,0 +1,36 @@
...
+
+        public OpenAIService(IConfiguration configuration, OpenAIClient openAIClient)
+        {
+            _configuration = configuration ?? throw new ArgumentNullException(nameof(configuration));
+            _openAIClient = openAIClient ?? throw new ArgumentNullException(nameof(openAIClient));
+        }
+
+        public async Task<Uri> GenerateImageAsync(ImageGenerationApiModel imageGenerationApiModel)
+        {
+
```
grantwinney, line 18: There's a blank line here.

dotnet-client-libraries/OpenAIImageGeneration/Tests/Tests/ImageGeneratorControllerTests.cs
```diff
@@ -0,0 +1,40 @@
+﻿using Microsoft.AspNetCore.Mvc;
+using Moq;
+using OpenAiImageGenerator;
+using OpenAiImageGenerator.Controllers;
+using OpenAiImageGenerator.Services;
+using System.Net;
```
grantwinney, line 6: I think these should be moved to the `GlobalUsings.cs` file, especially since they'd likely be present in every test class, if this project had multiple test classes...

PR #2021, opened 2024-06-22
CM-1150: StringValues in ASP.NET Core

karthikns9:


strings-csharp/StringValues/StringValues/NaiveImplementation.cs
```diff
@@ -0,0 +1,21 @@
+﻿public class NaiveImplementation
+{
+    private readonly Dictionary<string, string[]> _headers = [];
+
+    public void AddHeader(string key, string value)
+    {
+        if (_headers.ContainsKey(key))
+        {
+            var values = new List<string>(_headers[key]) { value };
+            _headers[key] = [.. values];
```
jshergal, line 10: I know this is the "naive" implementation, but I think we could do better here by just using `Array.Resize()` and then adding the new element to the end. I know that it is still inefficient, but I would expect it to be more performant than what we are doing here. Here we are allocating a new list, potentially resizing that list by adding a new element to it, and then converting that list into a new array.

PR #2032, opened 2024-07-04
CM-870: How to Convert a Stream to a File in C#

bozo-spoljaric:


csharp-intermediate-topics/HowToConvertAStreamToAFileInCSharp/Tests/ConvertAStreamToAFileInCSharpTests.cs
```diff
@@ -0,0 +1,80 @@
...
+    private string GetPath()
+    {
+        var directory = Path.GetDirectoryName(Assembly.GetExecutingAssembly().Location);
+        if (!Directory.Exists(directory))
+        {
+            return string.Empty;
+        }
+
+        var sourceFile = Path.Combine(directory, @"Files\source.png");
+        var destinationPath = Path.Combine(directory, @"Files\");
```
jshergal, line 59: @bozo-spoljaric - don't include the file separator in these methods, `Path.Combine()` will add the system-appropriate separator for us. Also, since we are not returning the source file, there is no need to compute it here, we can just compute the destination path.

bozo-spoljaric: Did so, thank you

PR #2033, opened 2024-07-04
CM-1089: How to Generate a Random Alphanumeric String in C#

lennart02:


strings-csharp/RandomAlphaNumericString/RandomAlphaNumericString/CryptographicallySecureMethods/AesMethod.cs
```diff
@@ -0,0 +1,15 @@
...
+{
+    public partial class Methods
+    {
+        //Maximal Length: 44
+        public static string? AesMethod(int length)
+        {
+            using var crypto = Aes.Create();
+            crypto.GenerateKey();
+            return Convert.ToBase64String(crypto.Key).Substring(0, length);
+        }
```
jshergal, line 13: This method is not producing only alpha-numeric strings, unless we are clear that both a '+', '/' and '=' are considered alpha-numeric.

strings-csharp/RandomAlphaNumericString/RandomAlphaNumericString/CryptographicallySecureMethods/SpanSecureMethod.cs
```diff
@@ -0,0 +1,36 @@
...
+{
+    public partial class Methods
+    {
+        public static string? SpanSecureMethod(int length)
+        {
+            Span<byte> data = stackalloc byte[4 * length];
+            using (RandomNumberGenerator rng = RandomNumberGenerator.Create())
+            {
+                rng.GetBytes(data);
+            }
```
jshergal, line 13: We can use the static methods on `RandomNumberGenerator` here without creating a new one instance

strings-csharp/RandomAlphaNumericString/RandomAlphaNumericString/CryptographicallySecureMethods/SpanSecureMethod.cs
```diff
@@ -0,0 +1,36 @@
+﻿using System.Security.Cryptography;
+
+namespace RandomAlphaNumericString
+{
+    public partial class Methods
+    {
+        public static string? SpanSecureMethod(int length)
+        {
+            Span<byte> data = stackalloc byte[4 * length];
```
jshergal, line 9: We need to validate the length otherwise this can lead to a stack overflow exception. We should be limiting the stackalloc to at most 256 characters, and since we are also doing another stackalloc based on the length later in this method, we really should limit to 128. Max stack size in a 64 bit process is 4 MB.

strings-csharp/RandomAlphaNumericString/RandomAlphaNumericString/Methods/SpanMethod.cs
```diff
@@ -0,0 +1,21 @@
+﻿using System.Security.Cryptography;
+
+namespace RandomAlphaNumericString
+{
+    public partial class Methods
+    {
+        public static string? SpanMethod(int length)
+        {
+            Span<char> result = stackalloc char[length];
```
jshergal, line 9: Same as previously mentioned, this method could overflow the stack given a large length

strings-csharp/RandomAlphaNumericString/RandomAlphaNumericString/CryptographicallySecureMethods/SpanSecureMethod.cs
```diff
@@ -0,0 +1,36 @@
...
+            {
+                var b = data[j++];
+                if (b >= acceptableByteRange)
+                {
+                    b = data[j++];
+                }
+
+                result[i] = chars[b % charSetLength];
+                i++;
+            }
```
jshergal, line 31: We could make this cleaner/faster since we are specifying our character set as `a-zA-Z0-9`. We know this is 62 characters. We can mask each byte by 0x3F (63) to ensure it stays within range. We will still need to handle the case of a value over 61, but in that case we can just move to the next byte, it would be a faster calculation than the modulus.

Also, we should benchmark whether pre-generating an array of bytes is faster than just calling `Next()` and masking it for each character. I don't know which way would be faster, but it is worth the test to see and then just show the more performant one.
It may turn out that a hybrid approach is better where we use a stackalloc buffer of say Min(length, 128) and then refresh it using GetBytes() as needed.

elfico:
@lennart02 
Here are some observations from the methods you implemented here:
In summary, 11 methods are too many

1. The AES method creates an AES key pair and then converts the key to a Base64 string. While this is "random" this is not an approach that I think anyone would actually use. It also adds characters such as '=', '+' and '/' into the string (because of Base64 encoding).

2. `Path.GetRandomFileName`, has a limited number of characters, uses only lowercase letters and does not include all digits (only 0-5). 

3. The GUID method is another one that is so restrictive. It is limited to 32 characters only and reducing the length will only increase the likelihood of collision.

4. We don't need to have a whole section with the `StringBuilder` option, since we have `string.Create()` available as of .NET Standard 2.1. Only someone who is restricted to the .NET Framework can't use it, so we can just show a quick snippet doing basically the same operation, just with a StringBuilder. The complete method should still be in the PR.

5. ForLoopMethod is really no different from other approaches and so doesn't really add any value as a separate method.

6. There unused imports in some of the classes

Let me know if you have additional questions

strings-csharp/RandomAlphaNumericString/RandomAlphaNumericString/CryptographicallySecureMethods/SpanSecureMethod.cs
lennart02: I tested something like this:
```
while (i < length)
{
    var b = data[j++] & 0x3F;
    if (b <= 61)
    {
        result[i++] = chars[b];
    }
}
```
This had a worse performance than using modulus. I committed a method with this code.

Also, I think I have to pre-generate the array, as there is nothing like a Next() operation.

elfico:
@lennart02 
In addition to my earlier feedback:

There really isn't (or at least shouldn't be) any difference between the Span() and string.Create() methods. We should create a private method that takes a span and does the work. As a whole there isn't a lot of difference between what we would do if we ourselves created and then populated a Span and if we do the work inside the string.Create() lambda. We would be operating on a span there as well, so we can make a helper method that takes a span and call it in both locations.

We can still highlight the difference between ran.GetItems() and ran.Next(length), but in that case, rather than naming the methods as "SpanMethod" and "StringCreateMethod", we can focus on the underlying algorithm that we are employing, "RandomInLoop" vs. "RandomGetItems"

lennart02:
@elfico 
I made the changes, but maybe I understand you wrong. Because now the both use a "random in loop" approach and the naming is not so obvious.

Furthermore, pls take a look at my answer of the last review from [jshergal](https://github.com/jshergal).

strings-csharp/RandomAlphaNumericString/RandomAlphaNumericString/CryptographicallySecureMethods/SpanSecureMethod.cs
jshergal: Hi @lennart02,
I am not sure about your benchmarking, when I ran it with the bit mask operation it was slightly faster (about 100 ns) using the bitmask vs. the method using modulus.

image

One thing I did change in the code was to make them match in both using RandomNumberGenerator.Fill() which is more efficient than GetBytes() when using a Span. GetBytes() will still internally create an array and then copy the data into the Span, whereas Fill() will write to the provided Span directly. That may be why you saw better performance between the modulus and the bit mask.

Regarding Next(), that was a mistake on my part, I mixed the method name between System.Random and RandomNumberGenerator, what I meant was RandomNumberGenerator.GetInt32(). Since I was running a benchmark, I ran the same using RandomNumberGenerator.GetInt32() and it was much slower than filling a stackalloc array of bytes.

Now with that being said, I looked back at the code that was using modulus and realized that we can simplify it and that leads to a method that is more performant:
image

```c#
public string OldSpanSecureMethod()
{
    if (length > 128)
    {
        throw new ArgumentOutOfRangeException(nameof(length), "Length must be less than or equal to 128");
    }
    Span<byte> data = stackalloc byte[length];
    RandomNumberGenerator.Fill(data);

    Span<char> result = stackalloc char[length];
    int charSetLength = chars.Length;

    int i = 0;
    int j = 0;
    while (i < length)
    {
        result[i++] = chars[data[j++] % 62];
    }

    return new string(result);
}
```

Notice that I removed the conditional within the while loop. Since we are going to take the modulus anyway, we can use every byte that was generated. That also means that our data array can be smaller. Removing that conditional speeds things up.

Lastly, in the initial code we supported returning an arbitrary length string, now the code has been changed to only support a max of 128 characters. We could still support an arbitrary length string (and should add benchmarks for small strings (~32), medium (~128) and large (~1024)). We just need to update the code to switch from stackalloc to ArrayPool in the case of a larger string. A simple ternary check with a finally to clear and return a rented array when we have one would suffice.
You can see an example of this in the .NET source code for System.Text.Json:
https://github.com/dotnet/runtime/blob/5b5d7919a6099c9ea0056eaa800d59f53b3d9fe1/src/libraries/System.Text.Json/src/System/Text/Json/JsonEncodedText.cs#L87

PR #2048, opened 2024-08-25
CM-1154: How to Use the REPR (Request-Endpoint-Response) Pattern in .NET

cjaynjoku:


csharp-design-patterns/REPRPattern/REPRPattern/BookEndpoint.cs
```diff
@@ -0,0 +1,15 @@
...
+
+public class BookEndpoint : Endpoint<BookEndpointRequest, BookEndpointResponse>
+{
+    public override void Configure()
+    {
+        Post("/api/book/create");
+        AllowAnonymous();
+    }
+
+    public override async Task HandleAsync(BookEndpointRequest req, CancellationToken ct)
```
elfico, line 13: @cjaynjoku we can improve the naming of the arguments in this method. `ct` is not contextual enough.

csharp-design-patterns/REPRPattern/REPRPattern/Program.cs
```diff
@@ -0,0 +1,8 @@
+using FastEndpoints;
+
+var bld = WebApplication.CreateBuilder();
```
elfico, line 3: @cjaynjoku here too. `bld` is not contextual. Can we rename to something else?

elfico (changes requested):

elfico (approved):

PR #2190, opened 2026-09-13
Hangfire vs Quartz.NET: remove the em dashes from three sample comments

vladimir-pecanac-main:
HangfireMonitoring.cs, LoggingJobListener.cs and QuartzStorageConfig.cs each carried one em dash in a comment.
Comment text only, so no executable line changes and the sample builds and tests as it stood.


vladimir-pecanac-main:
Superseded by #2220, merged as 29afa516. It rewrote these three comments for Quartz.NET 4.1.0, and main now has no em dashes in them.

